PR #3, opened 2012-06-04
SWC-2 Fix

RadicalZephyr:


RadicalZephyr:
Cleaned up the commit.  Should be good now?

PR #5, opened 2012-06-12
Develop

RadicalZephyr:
Fixes for SWC-16 and SWC-53.

I'm not sure that the way that I implemented the displayGlobalAlert function in DisplayUtils is really the best way to do it, but it's the only way that I could get it to work without giving me an error.


RadicalZephyr:
Also fixed SWC-82 (Download message being incorrect).  The button will now redirect you to the login page if you click it when not logged in.

PR #11, opened 2012-06-25
SWC-60: change from code review: overload method to hide implementation detail

jay-hodgson:


dburdick:
Rad, thanks for making the change Jay

PR #16, opened 2012-07-10
SWC-107 (and SWC-108): add a user profile page to Synapse (including profile picture support), and pull out private settings to a new page.

jay-hodgson:


jay-hodgson:
I don't want these changes as part of the next web client release, since I'd like to include an additional significant refactor to SWC-107 (replacing UserData).  Will reopen later.

PR #30, opened 2012-07-26
Block creation of Study entities (see PLFM-1404)

bkng:


bkng:
Will re-submit when Study -> Folder migration tool is ready.

PR #38, opened 2012-08-01
SWC-161: to fix null element set style error

jay-hodgson:


dburdick:
Looks good. Use of LayoutContainer over named DOM elements makes the element lifestyle transparent(ish)

PR #61, opened 2012-08-17
SWC-144: Add a check for empty or null project names

RadicalZephyr:
Also add tests for the Presenter


jay-hodgson:
Please move the nice error message to the platform services (as this is probably a common problem across clients).

PR #65, opened 2012-08-21
SWC-127: Add Download as a permission level to the Web UI

RadicalZephyr:
Large refactor, small feature changes.


jay-hodgson:
Need to talk with Dave about a new download permission (he said repo side governance logic currently determines the can-download permission)

PR #66, opened 2012-08-21
Refactor AclUtils to not have so much repetition

RadicalZephyr:
Now there's one canonical mapping of ACCESS_TYPE to PermissionLevel
and the other representations are built off that.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/shared/users/AclUtils.java
+++ b/src/main/java/org/sagebionetworks/web/shared/users/AclUtils.java
@@ -1,9 +1,10 @@
 package org.sagebionetworks.web.shared.users;
 
-import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/web/shared/users/AclUtils.java
```diff
@@ -38,15 +76,30 @@ public void onSuccess(Boolean canAdmin) {
...
 						@Override
 						public void onFailure(Throwable caught) {
 							callback.onFailure(caught);
 						}
-					});				
+					});
 				} else {
-					// user can only view
-					callback.onSuccess(PermissionLevel.CAN_VIEW);
+					// user can only view/download
```
dburdick: remove this due to governance confusion

PR #71, opened 2012-08-24
SWC-213: 'edit' permissions enables Move Entity; SWC-215: block child Folders

bkng:


bkng:
Jay, this is ready to go. We reviewed this Friday; the only changes is the updated pom reference. Thanks!

PR #72, opened 2012-08-28
SWC-85: 'Apply permissions to children' button on Sharing dialog

bkng:


src/main/java/org/sagebionetworks/web/client/widget/sharing/AccessControlListEditorViewImpl.java
```diff
@@ -437,6 +451,7 @@ public void onClick(ClickEvent event) {
 	/*
 	 * Private Classes
 	 */
+	@SuppressWarnings("serial")
```
bkng: remove this

PR #90, opened 2012-09-24
Develop

RadicalZephyr:


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -251,4 +183,53 @@ private void sendDetailsToView(boolean isAdmin, boolean canEdit) {
...
 
+	private void sendVersionInfoToView() {
+		Entity entity = bundle.getEntity();
+		if (entity instanceof Versionable) {
+			synapseClient.getEntityVersions(entity.getId(),
+					new AsyncCallback<String>() {
+
+						@Override
+						public void onFailure(Throwable caught) {
+							// TODO ????
```
jay-hodgson: please update view to tell user that no versions were found (for some reason or another)

PR #95, opened 2012-09-27
Develop

brucehoff:
This is the SWC work for Jira issues SWC-52, SWC-121, SWC-157, SWC-159, SWC-160.


pom.xml
```diff
@@ -15,7 +15,10 @@
 		<maven.compiler.source>1.6</maven.compiler.source>
 		<maven.compiler.target>1.6</maven.compiler.target>
 		<webappDirectory>${project.build.directory}/${project.build.finalName}</webappDirectory>
-		<synapse.version>2012-09-21-a6d6ad7-125</synapse.version>
+
+		<!--synapse.version>2012-09-21-a6d6ad7-125</synapse.version -->
+		<synapse.version>develop-SNAPSHOT</synapse.version>
```
jay-hodgson: (synapse.version should not point to develop-SNAPSHOT)

brucehoff: Jay,
I'll update the pom.xml before the review; yes we can 'hang out'.
Thanks,
Bruce

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -998,7 +1000,17 @@ public static String getVersionDisplay(Versionable versionable) {
 		}
 		return version;
 	}
+
 	
+	// from http://stackoverflow.com/questions/3907531/gwt-open-page-in-a-new-tab
+	public static native JavaScriptObject newWindow(String url, String name, String features)/*-{
```
jay-hodgson, line 1006: I'd like to get Dave's response to this change.  How important is it that it opens in a new tab in the callback? Does it warrant introducing this native code for the workaround.  This applies to Chrome and IE, so if this were replaced with the standard Window.open, we would need to be ok with it popping up in a new window for the majority of users.

src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
```diff
@@ -151,7 +151,7 @@ public void accepted() {
 						});			
 				}
 				public void onFailure(Throwable throwable) {
-					view.showErrorMessage("An error occured. Please try logging in again.");
+					view.showErrorMessage("An error occurred. Please try logging in again.");
```
jay-hodgson, line 154: good catch, please pull out as constant

src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -139,7 +139,7 @@ public void onSuccess(String result) {
 									@Override
 									public void onFailure(Throwable caught) {
 										view.userUpdateFailed();
-										view.showErrorMessage("An error occured. Please try reloading the page.");
+										view.showErrorMessage("An error occurred. Please try reloading the page.");
```
jay-hodgson, line 142: same, good catch, please pull out as constant

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -180,7 +195,174 @@ public void onFailure(Throwable caught) {
 	private void sendDetailsToView(boolean isAdmin, boolean canEdit) {
 		ObjectSchema schema = schemaCache.getSchemaEntity(bundle.getEntity());
 		entityTypeDisplay = DisplayUtils.getEntityTypeDisplay(schema);
-		view.setEntityBundle(bundle, entityTypeDisplay, isAdmin, canEdit, readOnly);
+		UserSessionData sessionData = authenticationController.getLoggedInUser();
+		UserProfile userProfile = (sessionData==null ? null : sessionData.getProfile());
```
jay-hodgson: may as well use your private method getUserProfile() here

src/main/resources/org/sagebionetworks/web/client/widget/entity/JiraGovernanceConstants.properties
```diff
@@ -0,0 +1,15 @@
+# properties used in the interface to Jira for Governance
```
jay-hodgson, line 1: this is cool, where else should we be using this?

src/test/java/org/sagebionetworks/web/unitclient/utils/GovernanceServiceHelperTest.java
```diff
@@ -0,0 +1,83 @@
...
+				fail("Exception not expected");
+			}};
+		final SynapseClientAsync synapseClient = (SynapseClientAsync) Proxy.newProxyInstance(
+				GovernanceServiceHelperTest.class.getClassLoader(), 
+				new Class<?>[]{SynapseClientAsync.class}, 
+				 new InvocationHandler() {
+					@Override
+					public Object invoke(Object synapseClient, Method method, Object[] args)
+							throws Throwable {
+						if (method.equals(SynapseClientAsync.class.getMethod("createAccessApproval", EntityWrapper.class, AsyncCallback.class))) {
```
jay-hodgson, line 54: Could use AsyncMockStubber to mock success and failure

src/main/java/org/sagebionetworks/web/client/widget/entity/download/LocationableUploaderViewImpl.java
```diff
@@ -209,12 +262,147 @@ public void componentSelected(ButtonEvent ce) {
...
+		lf.setAutoHeight(true);
+		tabItem.add(lf);
+	}
+	
+	private void addRadioButtonsToLayoutContainer(
+			LayoutContainer layoutContainer,
+			Radio openRadio,
+			Radio restrictedRadio) {
+		FormPanel radioButtonPanel = new FormPanel();
+		radioButtonPanel.setHeaderVisible(false);
```
jay-hodgson, line 285: like to see reworking the radio UI portion of this dialog (considering borders, and radio buttons could live outside of the form), can wait until later.

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
brucehoff, line 1006: Response:  Opening a window rather than a tab is blocked by Chrome's popup blocker.  This is the only solution I found that allows opening up a tab in the callback.

src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
brucehoff, line 154: I agree that English text should be moved to DisplayConstants as statically defined strings, but that's beyond the scope of this commit.  (I was just sick of seeing "occurred" misspelled!)  I created https://sagebionetworks.jira.com/browse/SWC-271 to capture this.

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
brucehoff: Yes, done!

src/main/java/org/sagebionetworks/web/client/widget/licenseddownloader/LicensedDownloaderViewImpl.java
```diff
@@ -80,47 +70,30 @@ protected void onRender(Element parent, int pos) {
...
-		case NONE:
-			// show download window
-			if (presenter.isDownloadAllowed()) {
-				createDownloadWindow();
-				downloadWindow.show();
-			}
-			break;
-		default:
-			throw new IllegalStateException(approvalRequired.toString());
+		if (!presenter.isDownloadAllowed()) return;
```
brucehoff, line 73: Per Jay, the logic around what window to show should be in the presenter.  Opened https://sagebionetworks.jira.com/browse/SWC-272 for this.

PR #99, opened 2012-10-02
More versions visual tweaks

RadicalZephyr:


src/main/java/org/sagebionetworks/web/client/widget/entity/TitleWidget.java
```diff
@@ -90,8 +89,10 @@ public void setVersions(Versionable entity, TreeMap<Long, String> latestVersions
 			if (!entity.getVersionNumber().equals(entry.getKey())) {
 				StringBuilder target = new StringBuilder("Synapse:");
 				target.append(entity.getId());
-				target.append("/version/");
-				target.append(entry.getKey());
+				if (!first) {
```
dburdick: You should be using DisplayUtils.getSynapseHistoryTokenNoHash (...) or one of its variants to create link urls to Synapse entities

PR #109, opened 2012-10-16
markdown changes (found in code review)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1050,4 +1050,26 @@ public static Anchor createIconLink(AbstractImagePrototype icon, ClickHandler cl
 		anchor.addClickHandler(clickHandler);
 		return anchor;
 	}
+
```
dburdick: I would pull the Markdown methods out into their own utility class as I'm guessing there will be more of them, but they are (mostly) only useful to the markdown code.

PR #111, opened 2012-10-17
SWC-292: fix error message when dupe email

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/servlet/UserAccountServiceImpl.java
```diff
@@ -347,6 +348,12 @@ public void createUser(UserRegistration userInfo) throws RestServiceException {
 		ResponseEntity<String> response;
 		try {
 			response = templateProvider.getTemplate().exchange(url, method, entity, String.class);
+		} catch(HttpClientErrorException ex) {
+			if (ex.getStatusCode() == HttpStatus.BAD_REQUEST)
```
dburdick, line 352: add test to createUser method to evoke this response

PR #125, opened 2012-11-01
Refactor of TitleWidget and EntityMetadata

RadicalZephyr:
Going from two simple (nested) widgets, to one widget which has a Presenter etc.

Essentially splits the EntityPageTopView in half, pulling a lot of the work done there down into EntityMetadata.

This is a step to make it easier to implement the new controls going into the previous versions table.


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTopViewImpl.java
```diff
@@ -268,10 +239,10 @@ public void clear() {
...
 		if (fullWidthContainer != null)
 			fullWidthContainer.removeAll();
-		if (titleWidget != null) {
-			titleWidget.clear();
-			titleWidget = null;
-		}
+//		if (titleWidget != null) {
+//			titleWidget.clear();
+//			titleWidget = null;
+//		}
```
RadicalZephyr: Remove commented out code.

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadataViewImpl.java
```diff
@@ -0,0 +1,526 @@
...
+						previousVersions.el().slideIn(Direction.DOWN, config);
+					}
+				}
+			}
+		});
+		previousVersions.setLayout(new FlowLayout(10));
+	}
+
+	@Override
+	public void setEntityBundle(EntityBundle bundle, boolean readOnly) {
```
RadicalZephyr, line 224: Reuse the grid instead of making a new one each time.   Also, make sure that we set the state of the previousVersions table to closed.

dburdick:
Looks good

PR #126, opened 2012-11-07
Visual changes to versions

RadicalZephyr:
Disable user selection, use selection for highlighting the currently being viewed object.


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadataViewImpl.java
```diff
@@ -438,8 +462,6 @@ public Object render(BaseModelData model, String property,
 					InlineLabel label = new InlineLabel(comment);
 					label.setTitle(comment);
 					return label;
-				} else if (model.get(property) != null) {
-					return model.get(property).toString();
```
jay-hodgson, line 442: need this?

PR #148, opened 2012-12-07
Wire up promote version service

RadicalZephyr:
The icon sucks, but that's what the code review is for!


src/main/java/org/sagebionetworks/web/client/IconsImageBundle.java
```diff
@@ -302,4 +302,9 @@
 	@Source("images/icons/visualize-16.png")
 	ImageResource visualize16();	
 
+	@Source("images/icons/arrow-turn-090-left.png")
+	ImageResource arrowTurnLeft16();
```
jay-hodgson: remove

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadata.java
```diff
@@ -321,9 +321,30 @@ public void onSuccess(EntityWrapper result) {
 	}
 
 	@Override
-	public void promoteVersion(String entityId, Long versionNumber) {
-		// TODO Auto-generated method stub
-
+	public void promoteVersion(final String entityId, final Long versionNumber) {
+		synapseClient.promoteEntityVersion(entityId, versionNumber, new AsyncCallback<String>() {
```
jay-hodgson, line 325: please test to make sure synapseClient.promoteEntityVersion is called when promoteVersion is

PR #155, opened 2012-12-10
Fix SWC-335: Show all versions only works after page refresh

RadicalZephyr:


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadataViewImpl.java
```diff
@@ -246,6 +248,9 @@ private void setPreviousVersionsVisible(boolean setVisible) {
 				previousVersions.setVisible(true);
 				allVersions.setText(DisplayConstants.HIDE_VERSIONS);
 				previousVersions.el().slideIn(Direction.DOWN, config);
+			} else {
+				// If we don't initiate an animation, unset the animating property
+				allVersions.getElement().setPropertyBoolean("animating", false);
```
RadicalZephyr, line 253: This pattern is too useful to be stuck in widget specific code.  Can we make it a display util?  Or otherwise generify it?

PR #170, opened 2012-12-23
Swap the all versions button labels

RadicalZephyr:


RadicalZephyr:
Thanks Jay!  I kinda forgot about this...

PR #175, opened 2013-01-09
Created EntityFinder widget for SWC-313 and other entity/entity version ...

dburdick:
...selection needs


src/main/java/org/sagebionetworks/web/server/servlet/SynapseProviderImpl.java
```diff
@@ -12,7 +12,9 @@
 
 	@Override
 	public Synapse createNewClient() {
-		return new Synapse();
+		// ONE LINE CHANGE TO USE STUB SYNAPSE CLIENT:		
+		//return new Synapse();
+		return SynapseClientStubUtil.createSynapseClient();
```
jay-hodgson: Is this a mistaken check-in (using the mock stubbed synapse client here)?

dburdick: You're absolutely right. good catch!

PR #185, opened 2013-01-17
SWC-369: for the login page to render correctly in Internet Explorer wit...

jay-hodgson:
...h default security settings, get the google image via https.


brucehoff:
Good thing Jay is not paid by the character.

PR #192, opened 2013-01-29
API SuperTable

jay-hodgson:
Support paging, column filtering, column rendering, and table styling and setting the width. Supports the ability for a column renderer to add its own set of columns. First set of column renderers are user id, synapse id, datetime, and a special case for entity annotation gathering. Add an alpha mode in the website to show alpha features (like the SuperTable widget editor).


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableColumnRenderer.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author jayhodgson
+ *
+ */
+public interface APITableColumnRenderer {
+	/**
+	 * renderer is initialized with the model data
+	 * @param columnData
+	 * @param callback
+	 */
+	void init(List<String> columnData, AsyncCallback<Void> callback);
```
dburdick: Maybe wrap columnCount, renderedColumnName and render into an object that is returned in the initialized callback (instead of void)

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableColumnRendererEntityIdAnnotations.java
```diff
@@ -0,0 +1,171 @@
...
+			public void onSuccess(EntityBundleTransport result) {
+				
+				try {
+					EntityBundle bundle = nodeModelCreator.createEntityBundle(result);
+					List<EntityRow<?>> entityRowList =  PropertyWidget.getRows(bundle.getEntity(), bundle.getAnnotations(), factory, cache);
+
+					//let's add a single row scalar to display the entity name
+					ObjectSchema schema = cache.getSchemaEntity(bundle.getEntity());
+					
+					EntityRowAnnotation<String> nameRowAnnotation = new EntityRowAnnotation<String>(new HashMap(), "Name", String.class);
```
jay-hodgson: don't include entity name or id in "annotations" renderer output

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableColumnRendererUserId.java
```diff
@@ -0,0 +1,102 @@
...
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.user.client.rpc.AsyncCallback;
+import com.google.inject.Inject;
+
+public class APITableColumnRendererUserId implements APITableColumnRenderer {
+
+	private SynapseClientAsync synapseClient;
+	private Map<String, String> userId2html;
+	private NodeModelCreator nodeModelCreator;
+	private String baseProfileAttachmentUrl = GWT.getModuleBaseURL()+"profileAttachment";
```
jay-hodgson: "profileAttachment "should be a constant

PR #205, opened 2013-02-25
SWC-208: make the links that are on the home page have standard hrefs in...

jay-hodgson:
...stead of the click handled by javascript (for SEO)

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
@@ -3,8 +3,6 @@ package org.sagebionetworks.web.client.view;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
 
 import org.gwttime.time.DateTime;
 import org.sagebionetworks.repo.model.EntityHeader;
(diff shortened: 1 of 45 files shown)


src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
```diff
@@ -593,16 +599,10 @@ private String formatFacetName(String name) {
...
-		a.setHTML(anchorName);	
-		a.addClickHandler(new ClickHandler() {					
-			@Override
-			public void onClick(ClickEvent event) {
-				Window.scrollTo(0, 0);
-				presenter.setStart(newStart);				
-			}
-		});
+		a.setHTML(anchorName);
+		a.setHref(DisplayUtils.getSearchHistoryToken(currentSearchJSON, newStart));
```
jay-hodgson, line 605: scroll to top on page change in search

jay-hodgson:
Using a regular expression to find all potential issues.  Note, this is the regular expression I came up with:
._[#]{1}[^!]{1}._

src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
jay-hodgson, line 605: since changing to href, window is automatically scrolled to the top when clicked.

PR #212, opened 2013-03-08
set content type based on stream using the new single FileHandle creation method

jay-hodgson:


jay-hodgson:
this is groundwork for SWC-248 (Generate appropriate previews for all leaf entities)

PR #221, opened 2013-03-18
Project default search SWC-420

dburdick:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -249,6 +246,10 @@ public static SearchQuery getDefaultSearchQuery() {
 		kv.setValue("link");
```
jay-hodgson: remove link

PR #227, opened 2013-03-25
Revert "PLFM-1792: patch production to roll back FileEntity support, unt...

jay-hodgson:
...il we have a way to reduce hops (during upload to S3)."

This reverts commit 63d64b450f60a817522cc2d6b8f436d1308edaa3.


jay-hodgson:
Note that I still need to update the pom.xml to point to a new bui

PR #239, opened 2013-03-28
SWC-437: ability to scale image, and wrap text

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/ImageWidgetViewImpl.java
```diff
@@ -23,31 +25,77 @@ public ImageWidgetViewImpl(SynapseJSNIUtils synapseJsniUtils) {
...
+			}
+		});
 		
-		add(new HTMLPanel(sb.toString()));
+		add(image);
 		this.layout(true);
 	}
 	
+	public void showError(String error) {
+		add(new HTML(error));
```
jay-hodgson, line 95: Idea: extract out error UI for reuse (using class=\"alert alert-block\")

PR #260, opened 2013-04-09
code splitting

john-hill:


src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -38,32 +38,48 @@ public void onModuleLoad() {
...
+				@Override
+				public void onFailure(Throwable reason) {
+					// Not sure what to do here.
+					loading.hide();
+					Window.alert(reason.getMessage());	
+				}
+
+				@Override
+				public void onSuccess() {
+					EventBus eventBus = ginjector.getEventBus();
```
jay-hodgson: Could you wrap this in a try/catch and call your onFailure() in the catch.

src/main/java/org/sagebionetworks/web/client/AppLoadingView.java
```diff
@@ -0,0 +1,50 @@
...
+ * 
+ * @author John
+ *
+ */
+public class AppLoadingView extends PopupPanel {
+	private final FlowPanel container = new FlowPanel();
+
+	public AppLoadingView() {
+		/**
+		 * DO NOT ADD THIS TO THE MAIN IMAGE BUNGLE!
```
john-hill, line 20: sp Bundle

src/main/java/org/sagebionetworks/web/client/presenter/Presenter.java
```diff
@@ -0,0 +1,15 @@
...
+import com.google.gwt.place.shared.Place;
+
+/**
+ * A presenter extends an Active and has a setPlace
+ * @author John
+ *
+ * @param <T> - the presenter's place
+ */
+public interface Presenter<T extends Place> extends Activity {
+
```
john-hill, line 13: doc

PR #317, opened 2013-04-25
Fix rendering for an entity be used multiple times SWC-580

dburdick:


src/main/java/org/sagebionetworks/web/client/widget/provenance/nchart/NChartUtil.java
```diff
@@ -49,8 +46,10 @@ public static NChartLayersArray createLayers(JsoProvider jsoProvider, ProvGraph
 		List<NChartLayer> layers = new ArrayList<NChartLayer>();
 		for(int i=0; i<=maxLayer; i++) {
 			List<NChartLayerNode> layerNodes = new ArrayList<NChartLayerNode>();
+			if(!layerToNode.containsKey(i)) continue; // possible with strange shifting in layer creation
 			// process ProvGraphNodes into a list of NChartLayerNodes
 			for(ProvGraphNode node : layerToNode.get(i)) {
+				System.out.println("Layer " + i + " " + node);
```
dburdick: remove

PR #322, opened 2013-04-30
Plfm-267: ability to change your email address via the profile page

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/users/PasswordResetPresenter.java
```diff
@@ -131,6 +144,21 @@ public void onFailure(Throwable caught) {
 							view.showErrorMessage(DisplayConstants.PASSWORD_SET_FAILED_TEXT);
 						}
 					});
+		} else if (changeEmailToken != null) {
+				userService.changeEmailAddress(changeEmailToken, newPassword, new AsyncCallback<Void>() {
+							@Override
+							public void onSuccess(Void result) {
+								view.showInfo(DisplayConstants.PASSWORD_AND_EMAIL_SET_TEXT);
```
jay-hodgson, line 151: show this in an area that persists on the page, or a popup

src/main/java/org/sagebionetworks/web/server/servlet/UserAccountServiceImpl.java
```diff
@@ -223,6 +223,54 @@ public void setRegistrationUserPassword(String registrationToken, String newPass
 			
 			throw new RestClientException("An error occurred. Please try again.");
 	}
+	
+	
+	@Override
+	public void changeEmailAddress(String changeEmailToken, String newPassword) {
+		// First make sure the service is ready to go.
```
jay-hodgson, line 230: this class needs to be refactored to use Synapse Java client

PR #337, opened 2013-05-15
Swc 610 track

xschildw:
Add logging for SWC-610 post stack-u


src/main/resources/log4j.properties
```diff
@@ -10,6 +10,13 @@ log4j.appender.A1=org.apache.log4j.ConsoleAppender
 log4j.appender.A1.layout=org.apache.log4j.PatternLayout
 log4j.appender.A1.layout.ConversionPattern=%d{HH:mm:ss,SSS} %-5p [%c] - %m%n
 
+#DebugFileAppender
```
jay-hodgson, line 13: need to remember to remove before branching

PR #344, opened 2013-05-17
Swc 619: UI to support Evaluation terms of use

jay-hodgson:
Note, must still update the pom to point to a compatible plfm artifact


brucehoff:
Awesome job, Jay!!

PR #345, opened 2013-05-20
Swc-613: support multi-part direct to s3 upload from the client (via javascript file slicing)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -177,25 +177,38 @@ private final static native boolean _isDirectUploadSupported() /*-{
...
+	
+	private final static native void _directUploadFile(String contentType, String fileFieldId, int startByte, int endByte, String url, XMLHttpRequest xhr) /*-{
 		var fileToUploadElement = $doc.getElementById(fileFieldId);
 		var fileToUpload = fileToUploadElement.files[0];
+		var start = parseInt(startByte) || 0;
+		var end = parseInt(endByte) || fileToUpload.size - 1;
+		var fileSliceToUpload;
+		//in versions later than Firefox 13 and Chrome 21, Blob.slice() is not prefixed (and the vendor prefixed methods are deprecated)
+		if (fileToUpload.slice) {
+        	fileSliceToUpload = fileToUpload.slice(start, end+1, contentType);
```
jay-hodgson, line 193: check to see if there is a performance improvement that can be made here, and test with bigger file (>1GB)

PR #360, opened 2013-05-30
SWC-643: widgetize the join button.  needed to retire Evaluation wiki

jay-hodgson:
refactoring logic from Evaluation page presenter to a new markdown widget


src/main/java/org/sagebionetworks/web/client/presenter/EvaluationPresenter.java
```diff
@@ -1,31 +1,15 @@
 package org.sagebionetworks.web.client.presenter;
```
jay-hodgson: Remove Evaluation Place

PR #375, opened 2013-06-06
SWC-658: collect profile info on registration, add team name to submitter list, and refactor profile edit form

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/ProfileFormPresenter.java
```diff
@@ -0,0 +1,152 @@
+package org.sagebionetworks.web.client.presenter;
```
jay-hodgson: rename to ProfileFormWidget

PR #384, opened 2013-06-11
Initial performance improvements

dburdick:


pom.xml
```diff
@@ -179,6 +179,12 @@
 					<!-- Run the GWT test with the headless html unit. -->
 					<mode>htmlunit</mode>
 					<module>org.sagebionetworks.web.Portal</module>
+					<!-- GWT compiler options -->
+			        <compileReport>true</compileReport>
+			        <encoding>${project.build.sourceEncoding}</encoding>
+			        <logLevel>INFO</logLevel>
+			        <extraJvmArgs>-Xmx896M -Xms896m -XX:PermSize=64m -XX:MaxPermSize=128m</extraJvmArgs>
```
jay-hodgson, line 186: Funny, I've always seen powers of 2 for initial/max heap size.  Now that I think about it, it probably makes no difference!

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1020,7 +1005,7 @@ public static boolean hasChildrenOrPreview(EntityBundle bundle){
 	 * @param pos where to position the tooltip relative to the widget
 	 */
 	public static void addTooltip(final SynapseJSNIUtils util, Widget widget, String tooltipText, TOOLTIP_POSITION pos){
-		Map<String, String> optionsMap = new TreeMap<String, String>();
+		Map<String, String> optionsMap = new HashMap<String, String>();
 		optionsMap.put("title", tooltipText);
```
jay-hodgson, line 1009: Strange that a TreeMap was used if key order does not matter. Nice catch.

PR #394, opened 2013-06-18
Removing Model Objects from AuthenticationController

dburdick:
...Factories (unused still)


jay-hodgson:
Reminder: need to auto-break the build if the bulk fragment shows up in the leftover!

Login page is probably next.

PR #395, opened 2013-06-18
Swc 670

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/download/UploaderViewImpl.java
```diff
@@ -471,10 +472,15 @@ private Widget createExternalPanel() {
...
 		externalLinkFormPanel.setButtonAlign(HorizontalAlignment.LEFT);
 		externalLinkFormPanel.setLabelWidth(110);
-		externalLinkFormPanel.setFieldWidth(230);
+		externalLinkFormPanel.setFieldWidth(PANEL_WIDTH-150);
 		pathField.setFieldLabel("External Path or URL");
 		
 		externalLinkFormPanel.add(pathField);
+		
+		nameField = new TextField<String>();
+		nameField.setFieldLabel("Name (Optional)");
```
dburdick, line 482: validate that name is valid entity name (use same regex as in ProjectsHomePresenter)

PR #416, opened 2013-06-25
Add global error handling code

dburdick:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -394,18 +395,24 @@ public static String getFileNameFromExternalUrl(String path){
 	 * @param placeChanger
 	 * @return true if the user has been prompted
 	 */
-	public static boolean handleServiceException(Throwable ex, PlaceChanger placeChanger, boolean isLoggedIn) {
-		if(ex instanceof UnauthorizedException) {
+	public static boolean handleServiceException(Throwable ex, PlaceChanger placeChanger, boolean isLoggedIn, SynapseView view) {
+		if(ex instanceof ReadOnlyModeException) {
+			MessageBox.info(DisplayConstants.READ_ONLY_MODE, DisplayConstants.SYNAPSE_IN_READ_ONLY_MODE, null);
```
jay-hodgson: If this method is being called from presenters, this MessageBox should also be replace.  
Probably with:
view.showErrorMessage(DisplayConstants.SYNAPSE_IN_READ_ONLY_MODE)

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -414,23 +421,29 @@ public static boolean handleServiceException(Throwable ex, PlaceChanger placeCha
...
-		} else if(ex instanceof ReadOnlyModeException) {
+		}
+		
+		// For other exceptions, allow the consumer to send a good message to the user
+		return false;
+	}
+	
+	public static boolean checkForRepoDown(Throwable caught, PlaceChanger placeChanger) {
+		if(caught instanceof ReadOnlyModeException) {
 			MessageBox.info(DisplayConstants.READ_ONLY_MODE, DisplayConstants.SYNAPSE_IN_READ_ONLY_MODE, null);
```
jay-hodgson: (here too)

PR #435, opened 2013-07-11
Patch Synapse version to revert ACL ID change

eric-wu:


eric-wu:
This points to a release build version and will NOT be merged back to develop.

PR #441, opened 2013-07-15
SWC-453 Reference/Footnote Widget

heidiso:


src/main/java/org/sagebionetworks/web/client/widget/entity/SharedMarkdownUtils.java
```diff
@@ -19,7 +19,13 @@ public static String getWidgetHTML(int widgetIndex, String suffix, String widget
 		sb.append(WebConstants.DIV_ID_WIDGET_PREFIX);
 		sb.append(widgetIndex);
 		sb.append(suffix);
-		sb.append("\" class=\"widgetContainer\" widgetParams=\"");
+		
+		//Reference widgets will be inline
+		if(widgetProperties.contains(WidgetConstants.REFERENCE_FOOTNOTE_KEY)) {
```
jay-hodgson: please change to a generic param that any widget could use to indicate that it should be placed inline.  The string that you're looking for might be something like "inlineWidget=true"

src/main/java/org/sagebionetworks/web/client/widget/entity/registration/WidgetConstants.java
```diff
@@ -95,7 +98,8 @@
 	
 	public static final String BUTTON_LINK_URL_KEY = "url";
 	public static final String BUTTON_LINK_TEXT_KEY = "text";
-	
+	public static final String REFERENCE_TEXT_KEY = BUTTON_LINK_TEXT_KEY;
```
jay-hodgson: Just use BUTTON_LINK_TEXT_KEY (you can rename to something like BUTTON_TEXT_KEY)

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/ReferenceWidgetViewImpl.java
```diff
@@ -0,0 +1,62 @@
...
+	public ReferenceWidgetViewImpl() {
+	}
+	
+	@Override
+	protected void onLoad() {
+		super.onLoad();
+		if(!hasLoaded) {
+			hasLoaded = true;
+			HTMLPanel parentPanel = (HTMLPanel)this.getParent();
+			Element heading = parentPanel.getElementById("footnote" + id);
```
jay-hodgson: please extract "footnote", and use in the parser

src/main/java/org/sagebionetworks/web/server/markdownparser/ReferenceParser.java
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	@Override
+	public void processLine(MarkdownElements line) {
+		String input = line.getMarkdown();
+		Matcher m = p1.matcher(input);
+		StringBuffer sb = new StringBuffer();
+		while(m.find()) {
+			//Store reference text and insert extra parameter for the renderer to link to
+			footnotes.add(input.substring(m.start() + 17, m.end() - 1));
```
jay-hodgson: Please make where 17 came from obvious.  like "${reference?text=".length() (at least I think that's where this comes from)

src/main/java/org/sagebionetworks/web/server/markdownparser/ReferenceParser.java
```diff
@@ -0,0 +1,43 @@
...
+			footnoteNumber++;
+		}
+		m.appendTail(sb);
+		line.updateMarkdown(sb.toString());
+	}
+	
+	@Override
+	public void completeParse(StringBuilder html) {
+		for(int i = 0; i < footnotes.size(); i++) {
+			String text = footnotes.get(i).replaceAll("%2E", ".");
```
jay-hodgson: This might be a bigger issue with over-escaping

src/main/java/org/sagebionetworks/web/server/markdownparser/ReferenceParser.java
```diff
@@ -0,0 +1,49 @@
...
+	public void processLine(MarkdownElements line) {
+		String input = line.getMarkdown();
+		Matcher m = p1.matcher(input);
+		StringBuffer sb = new StringBuffer();
+		while(m.find()) {
+			//Store reference text and insert extra parameter for the renderer to link to
+			int lenOfSyntax = "${reference?text=".length();
+			footnotes.add(input.substring(m.start() + lenOfSyntax, m.end() - 1));
+			String updated = input.substring(m.start(), m.end() - 1) + 
+				"&" + WidgetConstants.INLINE_WIDGET_KEY + "=true&" + WidgetConstants.REFERENCE_FOOTNOTE_KEY + "=" + footnoteNumber + "}";
```
jay-hodgson: Please put(WidgetConstants.INLINE_WIDGET_KEY, true) in the widget params from the editor instead of injecting in the parser.  Other widgets should do the same thing without having a parser side.

PR #449, opened 2013-07-16
Fixed the bug showing parameters in the footnotes

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/MarkdownRegExConstants.java
```diff
@@ -88,7 +88,7 @@
 	 * Recognized example input:
 	 * ${reference?text=So H et al. Journal of Awesomeness. 2013 March 15}
 	 */
-	public static final String REFERENCE_REGEX = "\\$\\{reference\\?text=(.+?)\\}";
+	public static final String REFERENCE_REGEX = "\\$\\{reference\\?text=(.+?)&inlineWidget=(.+?)\\}";
```
jay-hodgson: no guarantee inlineWidget param is going to be output in this order.   Possible that it is output as ${reference?inlineWidget=true&text=my text}

PR #455, opened 2013-07-18
Adding tooltips for GWT widgets

heidiso:
Suggestions on other ways to support tooltips for these widgets? I'm just returning an attached panel for callers to customize however needed...

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
@@ -108,7 +108,6 @@ import com.google.gwt.safehtml.shared.SafeHtmlBuilder;
 import com.google.gwt.safehtml.shared.SafeHtmlUtils;
 import com.google.gwt.user.client.DOM;
 import com.google.gwt.user.client.Element;
-import com.google.gwt.user.client.Event;
 import com.google.gwt.user.client.ui.AbstractImagePrototype;
 import com.google.gwt.user.client.ui.Anchor;
 import com.google.gwt.user.client.ui.FlexTable;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -956,6 +960,89 @@ public static boolean hasChildrenOrPreview(EntityBundle bundle){
...
+	/**
+	 * Creates and attaches a simple tooltip to a GWT widget
+	 * 
+	 * Customization of content and style are made by the caller method
+	 * 
+	 * @param widget: the widget to attach the popup/tooltip to
+	 * @return the new popup/tooltip that can be customized
+	 */
+	public static PopupPanel addToolTip(final Widget widget) {
+		final PopupPanel popup = createToolTip(false);
```
jay-hodgson: Please change this code to instead simply call the more complicated signature of the same method name (essentially making default values for some parameters).

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -956,6 +960,89 @@ public static boolean hasChildrenOrPreview(EntityBundle bundle){
...
+	}
+	
+	private static PopupPanel createToolTip(boolean autoHide) {
+		final PopupPanel popup = new PopupPanel(autoHide);
+		popup.setGlassEnabled(false);
+		popup.addStyleName("topLevelZIndex");
+		return popup;
+	}
+	
+	private static void setMouseOverBehavior(final Widget widget, final PopupPanel popup) {
```
jay-hodgson: Based on the current method names, I would expect setMouseOverBehavior to take a MouseOverHandler (and setMouseOutBehavior to take a MouseOutHandler) as parameters.

PR #467, opened 2013-07-25
SWC-701 Added bookmark/in-page linking feature to the markdown processor

heidiso:
Changed the container to a span for widgets that need to be inline because reference links/bookmark links would not properly inline when adjacent to other (inline) elements like links, bolded terms etc.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/editor/BookmarkConfigEditor.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/editor/BookmarkConfigEditor.java
@@ -11,6 +11,7 @@ import com.google.inject.Inject;
 
 public class BookmarkConfigEditor implements BookmarkConfigView.Presenter, WidgetEditorPresenter{
 	private BookmarkConfigView view;
+	private Map<String, String> descriptor;
 	
 	@Inject
 	public BookmarkConfigEditor(BookmarkConfigView view) {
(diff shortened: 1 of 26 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/editor/BookmarkConfigEditor.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+	
+	@Override
+	public Widget asWidget() {
+		return view.asWidget();
+	}
+
+	@Override
+	public void configure(WikiPageKey wikiKey,
+			Map<String, String> widgetDescriptor) {
```
jay-hodgson, line 30: please update view with values from widgetDescriptor.  In the future, we will enhance the markdown framework so that you can edit existing synapse widget strings (will detect and bring up the proper editor).

src/main/java/org/sagebionetworks/web/client/widget/entity/editor/BookmarkConfigViewImpl.java
```diff
@@ -0,0 +1,126 @@
...
+		linkTextLabel.setWidth(70);
+		linkTextField.setWidth(270);
+		hp.add(linkTextLabel);
+		hp.add(linkTextField);
+		hp.addStyleName("margin-top-left-10");
+		vp.add(hp);
+		
+		hp = new HorizontalPanel();
+		hp.setVerticalAlign(VerticalAlignment.MIDDLE);
+		bookmarkIdField = new TextField<String>();
```
jay-hodgson: What do you think about the idea to simplify user interaction by eliminating this bookmark target id field?  I foresee some confusion around why they need to specify this parameter. 
Can we try to automatically generate a unique id (unique within the scope of the document) on behalf of the user (perhaps the current time in ms would be sufficient?)?

PR #474, opened 2013-07-29
Swc-728: evaluation ACL UI

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/sharing/EvaluationAccessControlListEditor.java
```diff
@@ -0,0 +1,413 @@
...
+	@Override
+	public void setAccess(Long principalId, PermissionLevel permissionLevel) {
+		validateEditorState();
+		String currentUserId = getCurrentUserId();
+		if (currentUserId != null && principalId.toString().equals(currentUserId)) {
+			showErrorMessage(ERROR_CANNOT_MODIFY_ACTIVE_USER_PERMISSIONS);
+			return;
+		}
+		if (principalId.equals(publicAclPrincipalId))
+			uep.setCanPublicRead(true);
```
jay-hodgson, line 284: might be participate (should look at permission level)?

src/main/java/org/sagebionetworks/web/client/widget/sharing/EvaluationAccessControlListEditor.java
```diff
@@ -0,0 +1,413 @@
...
+		
+		ResourceAccess toSet = null;
+		for (ResourceAccess ra : acl.getResourceAccess()) {
```
jay-hodgson, line 287: Change Make Public to Open For Participation (Close Participation), that adds PARTICIPATE permission level to authenticated users group (if that is how it ends up working on the back-end).

src/main/java/org/sagebionetworks/web/client/widget/sharing/EvaluationAccessControlListEditor.java
```diff
@@ -0,0 +1,413 @@
...
+	@Override
+	public void removeAccess(Long principalIdToRemove) {
+		validateEditorState();
+		String currentUserId = getCurrentUserId();
+		if (currentUserId != null && principalIdToRemove.toString().equals(currentUserId)) {	
+			showErrorMessage(ERROR_CANNOT_MODIFY_ACTIVE_USER_PERMISSIONS);
+			return;
+		}
+		if (principalIdToRemove.equals(publicAclPrincipalId))
+			uep.setCanPublicRead(false);
```
jay-hodgson, line 316: also needs to change

PR #475, opened 2013-07-29
SWC-715 Checking for unmet requirements before submission

heidiso:


src/main/java/org/sagebionetworks/web/client/widget/entity/EvaluationSubmitter.java
```diff
@@ -150,6 +156,13 @@ public void submitToEvaluations(Reference selectedReference, final List<Evaluati
 		final String id = entityId;
 		final Long ver = version;
 		
+		//Check access requirements with the entity id before moving on with submission
+		try {
+			submitToEvaluations(entityId);
```
jay-hodgson: One way to catch that issue I mentioned (around the asynchronous call) is to add a test.  In this case it would be something like:  mock > 0 unmet requirements, and verify it does not call synapseClient.createSubmission().  Then mock 0 unmet requirements, and verify that it does call synapseClient.createSubmission().

PR #478, opened 2013-08-01
SWC-735 Reference widget improvements

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/DoiAutoLinkParser.java
```diff
@@ -0,0 +1,14 @@
...
+
+import org.jsoup.Jsoup;
+import org.sagebionetworks.web.server.DoiAutoLinkDetector;
+
+public class DoiAutoLinkParser extends BasicMarkdownElementParser {
+	DoiAutoLinkDetector detector = DoiAutoLinkDetector.getInstance();
+	
+	@Override
+	public void processLine(MarkdownElements line) {
+		detector.createLinks(Jsoup.parse(line.getHtml()));
```
jay-hodgson: Looks like it will work in most cases, but would rather see removal of the overhead from Jsoup.parse() that is currently executed on every line by this code (removing AutoLinkDetectors from Synapse Markdown Processor).  I think these AutoLinkDetectors can all be translated directly to standard parsers (extract the associated regular expressions, and extract the regular expression groups to replace params in the output html).

PR #486, opened 2013-08-06
SWC-746 Markdown table style

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/MarkdownRegExConstants.java
```diff
@@ -73,6 +73,18 @@
 
 	/**
 	 * Recognized example input:
+	 * {| class="border"
+	 */
+	public static final String TABLE_START_REGEX = "(\\{\\|\\sclass=\"(.*?)\")|(\\{\\|)";
```
jay-hodgson: be more lenient with regard to whitespace (I think the start tag plus one or more whitespace, for example).  Same comment for TABLE_END_REGEX (if the line only contains the end tag and whitespace, it should match)

src/main/java/org/sagebionetworks/web/server/markdownparser/TableParser.java
```diff
@@ -1,32 +1,62 @@
...
+		isTableStart = false;
+		isTableEnd = false;
 		tableCount = 0;
 	}
 
 	@Override
 	public void processLine(MarkdownElements line) {
 		String markdown = line.getMarkdown();
+		
 		boolean isTableMatch = p.matcher(markdown).matches();
```
jay-hodgson: I would rework slightly so that if we are inside of one of the fenced tables, then we should assume it's a table (don't even ask TABLE_REGEX, as it could be a single column table).

heidiso: But there could be single column tables in non-fenced tables, so should I not be calling TABLE_REGEX at all?

jay-hodgson: The currently supported syntax does not support single column tables in non-fenced tables, since we're looking for column separators rather than table borders.
In the fenced table, do not call TABLE_REGEX at all (if the markdown indicates that we are in a table, then be in a table until we see the end tag).

src/main/java/org/sagebionetworks/web/server/markdownparser/TableParser.java
```diff
@@ -1,83 +1,132 @@
...
 	Pattern p = Pattern.compile(MarkdownRegExConstants.TABLE_REGEX, Pattern.DOTALL);;
-	boolean isInTable;
-	int tableCount;
+	Pattern end = Pattern.compile(MarkdownRegExConstants.TABLE_END_REGEX);
+	boolean hasTags; 		//Are we creating a fenced table versus a table with just column separated by pipes
+	boolean isInTable;		//Have we started/are we in the middle of creating a table
+	boolean isTableStart;	//Is this the opening fence of the table
+	boolean isTableEnd;		//Is this the closing end of the table
+	boolean headColMade;	//To determine what type of row to create
+	boolean isTableMatch;	//Is the line a part of a table
```
jay-hodgson: Let's walk though this tomorrow, but it seems like we have at least one extra flag in here. Can you remove isTableMatch (just use isInTable)?

heidiso: isTableMatch is being used to determine if we're still reading rows or if we're done, and isInTable is keeping track of whether or not we're in the middle of creating a table so that if we're not in a table, we don't update the markdown. But I can take isTableMatch out, use isInTable and follow the same logic by using headColMade to serve as the flag that tells us if we're making a table.

src/main/java/org/sagebionetworks/web/server/markdownparser/MarkdownRegExConstants.java
```diff
@@ -73,6 +73,18 @@
...
 	 * Recognized example input:
+	 * {| class="border"
+	 */
+	public static final String TABLE_START_REGEX = "(\\{\\|\\s*class=\"(.*?)\"\\s*)|(\\{\\|\\s*)";
+	
+	/**
+	 * Recognized example input:
+	 * |}
+	 */
+	public static final String TABLE_END_REGEX = "\\|\\}\\s*";
```
jay-hodgson: sorry, could you allow space before the end as well?  (in an attempt to minimize when we accidentally don't close the table)

PR #501, opened 2013-08-12
SWC-766 [Markdown] Handles urls starting with www in linkParser

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/LinkParser.java
```diff
@@ -27,6 +28,11 @@ public void processLine(MarkdownElements line) {
 				url.substring(bookmarkTarget.length()) + WidgetConstants.WIDGET_END_MARKDOWN;			
 
 			} else {
+				//Check for incomplete url
+				if(url.startsWith(WebConstants.URL_WWW_PREFIX)) {
```
jay-hodgson: By just looking for "www", many other urls will still be broken (like github.com for example, or "WWW" for that matter).  I suggest looking for url.toLowercase().startsWith("http").  If false, then prepend the URL_PROTOCOL to the url.

src/main/java/org/sagebionetworks/web/server/markdownparser/LinkParser.java
```diff
@@ -27,6 +28,11 @@ public void processLine(MarkdownElements line) {
...
+					url = WebConstants.URL_PROTOCOL + url;
+				}
+				//Create link
```
jay-hodgson, line 36: And please add unit tests around this.

src/main/java/org/sagebionetworks/web/server/markdownparser/LinkParser.java
jay-hodgson: (with the exception of our places notation, since those are valid links in the same Synapse domain.  so if it starts with "#" do not prepend)

PR #506, opened 2013-08-13
SWC-776: autoselect entity id and md5 on click

jay-hodgson:


jay-hodgson:
does this interfere with behavior around the auto-select label (like selecting the entity name+id, or double clicking)

PR #508, opened 2013-08-13
SWC-673: add editor, and fix encoding since pipes are in TeX commands

jay-hodgson:


src/main/webapp/Portal.html
```diff
@@ -91,7 +92,18 @@
 		<script type="text/javascript" language="javascript" src="js/nchart.js"></script>
 		<script type="text/javascript" language="javascript" src="Portal/Portal.nocache.js"></script>
 		<script language='javascript' src='resources/flash/swfobject.js'></script>
-			
+		
+		<script>
+			window.layoutMath = function() {
```
dburdick: remove

src/main/java/org/sagebionetworks/web/client/widget/entity/editor/MathJaxConfigViewImpl.java
```diff
@@ -0,0 +1,109 @@
...
+		add(hp);
+	}
+	
+	@Override
+	public void configure(WikiPageKey wikiKey, Map<String, String> descriptor) {
+		String text = descriptor.get(WidgetConstants.MATHJAX_WIDGET_EQUATION_KEY);
+		if (text != null)
+			field.setValue(text);
+		else
+			field.setValue(WebConstants.DEFAULT_MATHJAX_VALUE);
```
dburdick: how about wrapping it with the prefix/suffix if they forget to add it

PR #511, opened 2013-08-14
SWC-775 [Markdown] Added syntax for explicit table header

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/TableParser.java
```diff
@@ -38,7 +44,7 @@ public void processLine(MarkdownElements line) {
 		if(isTableStart) {
```
jay-hodgson, line 44: Please break processLine into smaller logical methods (and reuse as much as possible, like the "reset to false for future tables" block)

src/main/java/org/sagebionetworks/web/server/markdownparser/TableParser.java
```diff
@@ -47,57 +53,86 @@ public void processLine(MarkdownElements line) {
...
-					}
-					builder.append("<thead>");
-					builder.append("<tr>");
-					String[] cells = markdown.split("\\|");
-					for (int j = 0; j < cells.length; j++) {
-						builder.append("<th>");
-						builder.append(cells[j]);
-						builder.append("</th>");
+				if(!readFirstRow) {
+					if(rowData.isEmpty()) {
```
jay-hodgson: is it possible for rowData to have data if we have not readFirstRow?

src/main/java/org/sagebionetworks/web/server/markdownparser/TableParser.java
```diff
@@ -47,57 +53,86 @@ public void processLine(MarkdownElements line) {
...
+								builder.append(rowData.get(j));
+								builder.append("</th>");
+							}
+							builder.append("</tr>");
+							builder.append("</thead>");
+							builder.append("<tbody>");
+						} else {
+							//Create a normal row with the stored data
+							builder.append("<tbody>");
+							createTableRow(builder);
```
jay-hodgson: I would make the interaction with rowData more explicit (instead of calling one method that you know modifies rowData, then calling another that implicitly uses rowData).  Change createTableRow to take a row of data as input, and rename storeRowData to getRowData (and only store across rows when you absolutely need to, in the case of the header row).

PR #513, opened 2013-08-15
SWC-784: edit existing widget markdown

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1613,5 +1614,38 @@ public static Long getVersion(Entity entity) {
...
+			startWord++;
+			String possibleWidget = text.substring(startWord);
+			if (possibleWidget.startsWith(WidgetConstants.WIDGET_START_MARKDOWN)) {
+				//find the end
+				int endWord = cursorPos;
+				while(endWord < text.length() && !WidgetConstants.WIDGET_END_MARKDOWN.equals(String.valueOf(text.charAt(endWord)))) {
+					endWord++;
+				}
+				//invalid widget specification if we went all the way to the end of the markdown
+				if (endWord != text.length()) {
```
dburdick: use comparator that indicates where endWord should be in relation to text.length()

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1613,5 +1614,38 @@ public static Long getVersion(Entity entity) {
 			version = ((Versionable) entity).getVersionNumber();
 		return version;
 	}
+	
+	public static WidgetSelectionState getWidgetSelectionState(String text, int cursorPos) {
```
dburdick: maybe add so that it updates an existing WidgetSelectionState for memory reasons.

src/main/java/org/sagebionetworks/web/client/widget/entity/WidgetSelectionState.java
```diff
@@ -0,0 +1,54 @@
+package org.sagebionetworks.web.client.widget.entity;
+
+public class WidgetSelectionState {
+	private boolean isWidgetSelected = false;
+	private String innerWidgetText = null;
+	private String fullMarkdownText = null;
```
dburdick: remove if not needed

PR #520, opened 2013-08-16
SWC-330 - small header on non home pages

dburdick:


src/main/java/org/sagebionetworks/web/client/view/HomeViewImpl.java
```diff
@@ -128,6 +128,7 @@ public HomeViewImpl(HomeViewImplUiBinder binder,
 		this.myEvaluationsList = myEvaluationsList;
 		this.iconsImageBundle = icons;
 		
+		headerWidget.configure(false);
```
jay-hodgson: this shouldn't be true?

src/main/java/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.java
```diff
@@ -299,4 +312,23 @@ public void onClick(ClickEvent event) {
 			}
 		}
 	}
+
+	@Override
+	public void showLargeLogo(boolean isLarge) {
```
jay-hodgson: rename method to showLogo

PR #525, opened 2013-08-19
SWC-790 Handled special characters in urls with new/enhanced renderers

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/ImageParser.java
```diff
@@ -2,13 +2,54 @@
...
+			html.append(alt + "\" />");
+			extractor.putContainerIdToContent(getCurrentDivID(), html.toString());
+		}
+		m.appendTail(sb);
+		line.updateMarkdown(sb.toString());
+	}
+
+	@Override
+	public void completeParse(Document doc) {
+		for(String key: extractor.getContainerIds()) {
```
jay-hodgson: Please refactor to server markdown util (takes a markdown extractor and a doc)

src/main/java/org/sagebionetworks/web/server/markdownparser/MathParser.java
```diff
@@ -81,7 +81,7 @@ private void processMathSpan(MarkdownElements line) {
 		StringBuffer sb = new StringBuffer();
 		while(m.find()) {
 			//leave containers to filled in on completeParse()
-			String containerElement = getNewMathElementStart() + ServerMarkdownUtils.END_MATH;
```
jay-hodgson, line 84: Rework my class so that it uses your new nifty MarkdownExtractor

src/main/java/org/sagebionetworks/web/server/SynapseMarkdownProcessor.java
```diff
@@ -66,6 +66,11 @@ private void init() {
 		//first initialize parsers that handle escaping
 		allElementParsers.add(new UnderscoreParser());
 		allElementParsers.add(new BacktickParser());
+		//parsers parsing urls need to handle them first
+		allElementParsers.add(new ImageParser());
+		allElementParsers.add(new LinkParser());
+		allElementParsers.add(new UrlAutoLinkParser());
```
jay-hodgson, line 74: I think CodeSpanParser needs to use the MarkdownExtractor too. (and should be the first simple parser to be run, since no other parsers should act on text inside the code span)

PR #526, opened 2013-08-19
Swc 673 alpha: Synapse markdown processor support for math blocks and single line span

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/markdownparser/MathParser.java
```diff
@@ -0,0 +1,115 @@
...
+		else {
+			if (isInMathBlock) {
+				if (!isFirstMathLine) {
+					line.prependElement("\n");	
+				}
+			}
+			else {
+				//not currently in a math block, and this is not the start/end of a math block
+				//check for math span
+				m = p2.matcher(line.getMarkdown());
```
jay-hodgson: pull out into it's own method

src/main/java/org/sagebionetworks/web/server/markdownparser/MathParser.java
```diff
@@ -0,0 +1,115 @@
...
+
+	
+	/**
+	 * Fill in the stored equations into the containers that we made during parse
+	 */
+	@Override
+	public void completeParse(Document doc) {
+		for (String key : div2equation.keySet()) {
+			Element el = doc.getElementById(key);
+			el.appendText(div2equation.get(key));
```
jay-hodgson: check for null

PR #530, opened 2013-08-20
SWC-798: table auto width (wrap in a div that has a fixed width, and auto overflow).

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/ServerMarkdownUtils.java
```diff
@@ -324,7 +324,7 @@ public static String resolveCodeWithLanguage(String markdown) {
 
 	
 	public static int appendNewTableHtml(StringBuilder builder, String regEx, String[] lines, int tableCount, int i) {
-		builder.append("<table id=\""+WidgetConstants.MARKDOWN_TABLE_ID_PREFIX+tableCount+"\" class=\"tablesorter\">");
+		builder.append("<div class=\"span-24 overflow-auto\"><table id=\""+WidgetConstants.MARKDOWN_TABLE_ID_PREFIX+tableCount+"\" class=\"tablesorter\">");
```
jay-hodgson: add notopmargin and last to div classes

PR #532, opened 2013-08-20
update to secure cdn to fix page load

jay-hodgson:


jay-hodgson:
(This is to fix SWC-800)

PR #534, opened 2013-08-21
SWC-795 [Markdown] Added an option to limit size of large tables

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/TableParser.java
```diff
@@ -101,6 +111,9 @@ public void processLine(MarkdownElements line) {
 					}
 					//End table and reset state for future tables
```
jay-hodgson: Please pull out into a new method (something like writeEndTable()) to remove duplicate code.

PR #536, opened 2013-08-22
SWC-773 Lists in blockquotes. Handled case where listParser parses before blockquoteParser

heidiso:


src/main/java/org/sagebionetworks/web/server/markdownparser/ListParser.java
```diff
@@ -98,8 +121,14 @@ public void getListItem(MarkdownElements line, Matcher m, boolean isOrderedList,
         	//no list in the stack
         	//create a new list
         	MarkdownList newList = getNewList(depth, isOrderedList);
-    		line.prependElement(newList.getStartListHtml());
-    		newList.addListItemHtml(line, value);
+        	if(preserveForBlockQuoteParser) {
+        		//Preserve the ">" character for the blockquote parser to prepend the blockquote element
+        		line.updateMarkdown(prefix + newList.getStartListHtml() + "<li><p>" + value + "</p>");
```
jay-hodgson: Possible to put all the html is protected area of MarkdownElements?  Say, updateMarkdown(prefix) and then append the rest?

src/main/java/org/sagebionetworks/web/server/markdownparser/ListParser.java
```diff
@@ -70,15 +76,32 @@ else if (isInMarkdownElement()) {
...
+        	} else {
+        		//The blockquote tag has not been made/this list item is starting the blockquote
+        		//We need to encompass the list with the blockquote element 
+        		preserveForBlockQuoteParser = true;
+        		hasSeenBlockQuote = true;
+        	}
+        	
+        	//Account for leading ">" blockquote character
+        	depth--;
+        } else if(line.getHtml().contains("<blockquote>")) {
```
jay-hodgson: Please extract <blockquote> constant and use here and in blockquote parser.  Maybe add hasElement to Markdownelements (that checks all prepended and appended elements for the given html) instead of asking for full line html?

PR #538, opened 2013-08-23
SWC-805 Fix table parser so single-row tables are parsed by simple parsers

heidiso:


jay-hodgson:
After talking with Heidi, the conclusion was to give access to the list of simple parsers to each parser (so that it can process before output to a protected area of MarkdownElements).  In this case, the table parser could process using the simple parsers before prepending the first table row.

src/main/java/org/sagebionetworks/web/server/markdownparser/ReferenceParser.java
```diff
@@ -87,7 +87,7 @@ public void completeParse(StringBuilder html) {
 		//No new lines; process over constructed footnote markdown
 		MarkdownElements elements = new MarkdownElements(footnoteMarkdown.toString());
 		for (MarkdownElementParser parser : parsersOnCompletion) {
-			parser.processLine(elements);
+			parser.processLine(elements, null);
```
jay-hodgson: Suggest removal of parsersOnCompletion and to use simple parsers

src/test/java/org/sagebionetworks/web/unitserver/markdownparser/TableParserTest.java
```diff
@@ -85,17 +87,17 @@ public void testTableWithOneLine() {
 		String exampleLine1 = "Row 1 Content Cell 1 | Row 1 Content Cell 2  | Row 1 Content Cell 3";
 		StringBuilder tableOutput = new StringBuilder();
 		MarkdownElements elements = new MarkdownElements(exampleLine1);
-		parser.processLine(elements);
-		underscoreParser.processLine(elements);
+		List<MarkdownElementParser> simple = new ArrayList<MarkdownElementParser>();
```
jay-hodgson: As a test, please add some simple markdown to a table cell and verify that it was processed (because that parser is in the simpleParser list you hand it)

PR #542, opened 2013-08-26
SWC-807: fix query table service pagination and editing existing widget

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableWidget.java
```diff
@@ -187,8 +188,14 @@ private String getColumnValue(JSONObjectAdapter row, String key) throws JSONObje
 	}
 	
 	private String getPagedURI() {
-		String firstCharacter = tableConfig.getUri().contains("?") ? "&" : "?";
-		return tableConfig.getUri() + firstCharacter + "limit="+tableConfig.getPageSize()+"&offset="+tableConfig.getOffset();
+		String uri = tableConfig.getUri();
```
dburdick, line 191: should write a test for proper query string creation

src/main/java/org/sagebionetworks/web/shared/WebConstants.java
```diff
@@ -162,5 +162,7 @@
 	
 
 	public static final String TEXT_TAB_SEPARATED_VALUES = "text/tab-separated-values";
+
+	public static final String QUERY_SERVICE_PREFIX = "/query?query=";
```
dburdick: move to ClientProperties

PR #543, opened 2013-08-27
SWC-804/743 Created new parser to detect/extract/protect synapse widget syntax

heidiso:


src/main/java/org/sagebionetworks/web/client/widget/entity/SharedMarkdownUtils.java
```diff
@@ -4,6 +4,7 @@
 import org.sagebionetworks.web.shared.WebConstants;
 
 public class SharedMarkdownUtils {
+	public static final int SYNAPSE_MARKDOWN_WIDGET_PARSER_INDEX = 3;
```
jay-hodgson: Please remove SYNAPSE_MARKDOWN_WIDGET_PARSER_INDEX. 
Suggest moving set simpleParsers away from processLine (since it does not change per line) to reset(List<MarkdownElementParser> simpleParsers), and figure out the synapse markdown widget index there (not per line) in the LinkParser.  Otherwise, if someone adds a new parser before the SynapseMarkdownWidgetParser in SynapseMarkdownProcessor.allElementParsers (or changes the position of the SynapseMarkdownWidgetParser) then they need to somehow remember to change this index too.

src/main/java/org/sagebionetworks/web/server/markdownparser/SynapseMarkdownWidgetParser.java
```diff
@@ -0,0 +1,102 @@
...
+	@Override
+	public void reset() {
+		extractor = new MarkdownExtractor();
+	}
+	
+	private String getCurrentDivID() {
+		return WebConstants.DIV_ID_WIDGET_SYNTAX_PREFIX + extractor.getCurrentContainerId() + suffix;
+	}
+	
+	private String getNewElementStart() {
```
jay-hodgson: I know I've seen this exact getNewElementStart code before in other parsers, could you extract and reuse?

src/main/java/org/sagebionetworks/web/server/markdownparser/SynapseMarkdownWidgetParser.java
```diff
@@ -0,0 +1,102 @@
...
+			html.append(m.group(2));
+			extractor.putContainerIdToContent(getCurrentDivID(), html.toString());
+		}
+		m.appendTail(sb);
+		line.updateMarkdown(sb.toString());
+	}
+	
+	@Override
+	public void completeParse(StringBuilder html) {
+		String subpagesWidgetMarkdown = SharedMarkdownUtils.getWikiSubpagesMarkdown();
```
jay-hodgson: Instead of depending on the WikiSubpageParser completeParse to run before this (and knowing what that parser will do), I'd rather you just incorporate WikiSubpageParser logic into this class (look for the subpage markdown in this class, and delete WikiSubpageParser).

src/test/java/org/sagebionetworks/web/unitserver/markdownparser/ReferenceParserTest.java
```diff
@@ -25,12 +27,16 @@
...
 		parser.reset();
-		
+		SynapseMarkdownWidgetParser widgetParser = new SynapseMarkdownWidgetParser();
+		widgetParser.reset();
 		LinkParser linkParser = new LinkParser();
 		linkParser.reset();
-		processor = SynapseMarkdownProcessor.getInstance();
 		simpleParsers = new ArrayList<MarkdownElementParser>();
 		simpleParsers.add(linkParser);
+		simpleParsers.add(linkParser);
```
jay-hodgson: Why do you add so many references (to linkParser) to the simpleParsers list in this test?

PR #544, opened 2013-08-27
Swc-347: report public if authenticated users group or all users group are in the acl (which would mean they have at least read).

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/sharing/AccessControlListEditor.java
```diff
@@ -130,8 +130,10 @@ public void onFailure(Throwable caught) {
 		return view.asWidget();
 	}
 	private void initViewPrincipalIds(){
-		view.setPublicPrincipalId(publicAclPrincipalId);
-		view.setAuthenticatedPrincipalId(authenticatedAclPrincipalId);
+		if (publicPrincipalIds != null) {
+			view.setPublicPrincipalId(publicPrincipalIds.getPublicAclPrincipalId());
+			view.setAuthenticatedPrincipalId(publicPrincipalIds.getAuthenticatedAclPrincipalId());
```
dburdick: set anonymous user into view and render it similarly to public group

PR #547, opened 2013-08-28
Swc-430: do not redirect search to the entity page unless it points to an entity that exists in the system (and that I have access to).

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/search/SearchBox.java
```diff
@@ -68,7 +72,7 @@ public void search(String value) {
 				// if fail, fall back on regular search
 			}
 		}
-		globalApplicationState.getPlaceChanger().goTo(new Search(value));
+		HomeSearchBox.searchForTerm(value, globalApplicationState, synapseClient);
```
dburdick: This should probably be refactored out into a static util class

PR #563, opened 2013-09-05
SWC-387;SWC-549;PLFM-1793;SWC-830

brucehoff:


src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
```diff
@@ -180,6 +180,10 @@ public void onFailure(Throwable t) {
 					}
 				});
 			} 
+		} else if (WebConstants.OPEN_ID_ERROR_TOKEN.equals(token)) {
+			globalApplicationState.getPlaceChanger().goTo(new LoginPlace(ClientProperties.DEFAULT_PLACE_TOKEN));
+			view.showErrorMessage("An error occurred. Please try logging in again.");
+			view.showLogin(openIdActionUrl, openIdReturnUrl);
```
brucehoff, line 186: move error message into display constants

PR #564, opened 2013-09-06
SWC-772 Added a check for uninitialized column data before rendering tables

heidiso:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableWidgetViewImpl.java
```diff
@@ -109,44 +115,37 @@ public void configure(Map<String,List<String>> columnData, String[] columnNames,
...
-		hp.addStyleName("margin-bottom-40");
-		hp.addStyleName("margin-top-10");
-		Image pageBack = new Image(iconsImageBundle.NavigateLeft16());
-		pageBack.addStyleName("imageButton");
-		pageBack.addStyleName("left");
-		pageBack.addClickHandler(new ClickHandler() {
+		String label = start + "-" + end + " of " + total;
+		StringBuilder builder = new StringBuilder();
+		builder.append("<ul class=\"pager\">");
+		builder.append("<li><a id=\"prevButton\">Previous</a></li>");
```
jay-hodgson: This technique is not going to work when there are more than one paginated supertable on the page.  I think you can use our UnorderedListPanel for this purpose (but I think you need to add a setClassName on it's UListElement to add the bootstrap pager class).

PR #572, opened 2013-09-10
SWC-841 Added check for null when updating view from descriptor

heidiso:


jay-hodgson:
Nice catch.  Please open a pull request into release-12 too.

PR #576, opened 2013-09-11
SWC-449 Gave widget editors access to submit button to disable/enable when appropriate

heidiso:


src/main/java/org/sagebionetworks/web/client/widget/entity/dialog/BaseEditWidgetDescriptorViewImpl.java
```diff
@@ -92,11 +78,31 @@ private void initializeBaseContentPanel() {
...
+	    String width = Integer.toString(STARTING_WIDTH + widgetDescriptorPresenter.getAdditionalWidth()) + "px";
+		String height = Integer.toString(STARTING_HEIGHT + widgetDescriptorPresenter.getDisplayHeight())+"px";
+		container.setWidth(width);
+		container.setHeight(height);
+		window.setWidth(width);
+		window.setHeight(height);
+		window.setHeading(windowTitle);
+		
+		//The ok/submitting button will be enabled when attachments/images are uploaded
+		if(windowTitle.equals(WidgetConstants.ATTACHMENT_PREVIEW_FRIENDLY_NAME) || windowTitle.equals(WidgetConstants.IMAGE_FRIENDLY_NAME)) {
```
jay-hodgson: Try editing an existing markdown widget definition, I don't think you would be able to save until you upload a new file.
In general, I'm not happy about disabling the button based on the windowTitle here.  Please move this to the attachment and image widgets (in the presenters, determine whether to enable, and pass to the view to make it so)?

heidiso: You're right, when you edit existing widgets, you have to load a new file. If you're not uploading a new file, you have to cancel out or browse, choose and reupload the same file. Reuploading the same file immediately gives an error.

Making edits to fix that.

PR #622, opened 2013-10-17
Swc 908

dburdick:


src/main/resources/org/sagebionetworks/web/client/view/HomeViewImpl.ui.xml
```diff
@@ -85,8 +85,16 @@
 				<g:SimplePanel ui:field="projectPanel" />
 			</div> <!-- end row -->				
 
+			<!-- SEARCH BOX -->
```
jay-hodgson, line 88: Add spacing above Search Synapse

PR #658, opened 2013-11-01
SWC-981: new join team and submit to evaluation

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/view/ProfileFormViewImpl.java
```diff
@@ -134,7 +134,7 @@ private void createProfileForm() {
 	     teamName = new TextField<String>();  
 	     teamName.setFieldLabel("DREAM 8 Team Name");  
 	     teamName.setAllowBlank(true);
-	     left.add(teamName, formData);
+//	     left.add(teamName, formData);
```
dburdick, line 137: remove

src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -1801,11 +1808,40 @@ public String getAvailableEvaluations() throws RestServiceException {
 		}
 	}
 	
+	@Override
```
dburdick, line 1811: evaluations to filter by should probably be pushed into the service

PR #660, opened 2013-11-01
swc-897 - fix annotations label issue

dburdick:


jay-hodgson:
(Note that DialogUtils is only used by AnnotationsWidgetViewImpl and AddAnnotationDialog)

PR #663, opened 2013-11-04
OpenID fix (develop)

kaysoky:
OpenID library has changed slightly (must specify OpenID provider name)


kaysoky:
Looks like a few methods have changed in the Java client...

PR #673, opened 2013-11-06
Accommodate auth schema changes

kaysoky:
_Note_: Requires the pom from: https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/808


kaysoky:
Looks like there will be some breakage as a result of https://github.com/Sage-Bionetworks/Synapse-Repository-Services/pull/806 as well.  I'll incorporate those fixes.

kaysoky:
Or not, looks like switching between branches with a different number of projects confuses Eclipse's compiler sometimes.

PR #688, opened 2013-11-11
Swc-1043: pass client host info to server, and set href target based on it.

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1939,4 +1939,8 @@ public void run() {
 	    return label;
 	}
 	
+	public static String getHostPrefix() {
+		return com.google.gwt.user.client.Window.Location.getProtocol()+"//"+com.google.gwt.user.client.Window.Location.getHost();
```
dburdick: this call to the gwt client should probably be moved behind the interface in our GWTWrapper

PR #704, opened 2013-11-19
[SWC-1054] Wiki Markdown Helper

heidiso:


kaysoky:
You may want to consider renaming the class to be more generic.  These classes will be helpful for saving messages (or any other text block) to S3, in future.

PR #723, opened 2013-11-26
V2 Wiki Switch

heidiso:
1) Added methods to Synapse Client which use V2 services. Also added methods to zip up and retrieve/read markdown using file handles etc.

2) Widgets now call on these new methods (and zips/unzips markdown). Through the UI, I've tried creating (and creating children), updating, adding attachments, and deleting.


src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -171,8 +190,14 @@ public void validateService() {
 		if (tokenProvider == null) {
 			throw new IllegalStateException("The token provider was not set");
 		}
+		if (s3Client == null) {
+			throw new IllegalStateException("The S3Client was not set");
+		}
 	}
 
+	private AmazonS3Client s3Client = new AmazonS3Client(new BasicAWSCredentials(StackConfiguration.getIAMUserId(), StackConfiguration.getIAMUserKey()));
```
john-hill: This will not work.  The Portal IAM user does not have permission to read from this bucket.  That is private data of the Repository Services and all clients must access it only through the public API (see PLFM-2368).  Therefore we cannot use the AmazonS3Client in the Portal.

Clients access data in S3 via per-signed URLs issued by the Repository Service.  See how the Portal both uploads and downloads wiki attachments.  Please follow the same pattern for makrdown upload and download.

src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -106,6 +114,10 @@
 import org.sagebionetworks.web.shared.exceptions.RestServiceException;
 import org.sagebionetworks.web.shared.exceptions.UnknownErrorException;
 
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.model.GetObjectRequest;
+import com.google.gwt.user.client.rpc.AsyncCallback;
```
john-hill: Why are the amazon imports here?

src/test/java/org/sagebionetworks/web/unitserver/SynapseClientImplTest.java
```diff
@@ -115,6 +122,10 @@
 import org.sagebionetworks.web.shared.users.AclUtils;
 import org.sagebionetworks.web.shared.users.PermissionLevel;
 
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.amazonaws.services.s3.model.GetObjectRequest;
+import com.amazonaws.services.s3.model.ObjectMetadata;
+
```
john-hill: again aws imports

src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -1280,6 +1288,334 @@ public String getWikiAttachmentHandles(org.sagebionetworks.web.shared.WikiPageKe
...
+			throw new UnknownErrorException(e.getMessage());
+		}
+	}
+	
+	private File zipUp(String content, String fileName) throws IOException {
+		// Create a temporary file to write content to
+		File tempFile = File.createTempFile(fileName, ".tmp");
+		if(content != null) {
+			FileUtils.writeByteArrayToFile(tempFile, content.getBytes());
+		} else {
```
john-hill, line 1521: All string.getBytes("UTF-8").

PR #734, opened 2013-12-05
SWC-1075: By popular demand, bringing back the My Challenges section to the front page (but it's no longer based on evaluations, it's a loose team2challenge mapping)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1566,7 +1567,18 @@ public static String createWikiAttachmentUrl(String baseFileHandleUrl, WikiPageK
 	public static String createFileEntityUrl(String baseFileHandleUrl, String entityId, Long versionNumber, boolean preview){
 		return createFileEntityUrl(baseFileHandleUrl, entityId, versionNumber, preview, false);
 	}
-
+	
+	/**
+	 * Create the url to a FileEntity filehandle.
```
dburdick: change comment and method signature to indicate that it is proxying and shouldn't be used much (if at all)

src/main/java/org/sagebionetworks/web/client/presenter/HomePresenter.java
```diff
@@ -208,6 +237,79 @@ public void onFailure(Throwable caught) {
 		});		
 	}
 	
+	public void getChallengeProjectIds(final List<Team> myTeams) {
+		getTeamId2ChallengeIdWhitelist(new CallbackP<JSONObject>() {
+			@Override
+			public void invoke(JSONObject mapping) {
+				List<String> challengeEntities = new ArrayList<String>();
```
dburdick: could use a set here and then just sort challenges by name on display

src/main/java/org/sagebionetworks/web/client/presenter/HomePresenter.java
```diff
@@ -208,6 +237,79 @@ public void onFailure(Throwable caught) {
...
+	
+	public void getTeamId2ChallengeIdWhitelist(final CallbackP<JSONObject> callback) {
+		requestBuilder.configure(RequestBuilder.GET, DisplayUtils.createRedirectUrl(synapseJSNIUtils.getBaseFileHandleUrl(), gwt.encodeQueryString(ClientProperties.TEAM2CHALLENGE_WHITELIST_URL)));
+	     try
+	     {
+	    	 requestBuilder.sendRequest(null, new RequestCallback() {
+	            @Override
+	            public void onError(Request request, Throwable exception) 
+	            {
+	            	view.setMyChallengesError("Could not load My Challenges: " + exception.getMessage());
```
dburdick: probably do nothing as you don't know if they have challenges or not

src/main/java/org/sagebionetworks/web/client/presenter/HomePresenter.java
```diff
@@ -208,6 +237,79 @@ public void onFailure(Throwable caught) {
...
+	            }
+
+	            @Override
+	            public void onResponseReceived(Request request,Response response) 
+	            {
+	            	String responseText = response.getText();
+	                JSONValue parsed = JSONParser.parseStrict(responseText);
```
dburdick: wrap with exception handler & cache results in HomePresenter memory

src/test/java/org/sagebionetworks/web/unitclient/presenter/HomePresenterTest.java
```diff
@@ -157,4 +173,18 @@ public void testIsOpenInvites() {
 		homePresenter.isOpenTeamInvites(mockCallback);
 		verify(mockCallback).invoke(eq(true));
 	}
+	
+	@Test
+	public void testGetChallengeProjectHeaders() {
+		homePresenter.getChallengeProjectHeaders(new ArrayList<String>());
```
dburdick: maybe black box test the whole white-list-team-to-project mapping system

PR #739, opened 2013-12-09
[SWC-1059] Adding created/modified information on Wikis

heidiso:


src/main/java/org/sagebionetworks/web/client/DisplayConstants.java
```diff
@@ -520,6 +520,8 @@
 	public static final String CREATED_BY = "Created by";
 	public static final String MODIFIED = "Modified";
 	public static final String MODIFIED_BY = "Modified by";
+	public static final String WIKI_MODIFIED_BY = "Wiki modified by";
+	public static final String WIKI_CREATED_BY = "Wiki created by";
```
jay-hodgson: Should not say Wiki on a FileEntity page (UI is calling the project wiki the wiki)

src/main/java/org/sagebionetworks/web/client/widget/entity/WikiPageWidget.java
```diff
@@ -120,8 +125,35 @@ public void onSuccess(String result) {
 							currentPage = nodeModelCreator.createJSONEntity(result, WikiPage.class);
 							wikiKey.setWikiPageId(currentPage.getId());
 							originalMarkdown = currentPage.getMarkdown();
-							boolean isRootWiki = currentPage.getParentWikiId() == null;
-							view.configure(currentPage, wikiKey, ownerObjectName, canEdit, isRootWiki, spanWidth, isDescription);
+							final boolean isRootWiki = currentPage.getParentWikiId() == null;
+							// User ids of creator and modifier
+							List<String> userIds = new ArrayList<String>();
+							userIds.add(currentPage.getModifiedBy());
+							userIds.add(currentPage.getCreatedBy());
```
jay-hodgson: Are these principal ids? Could we just use a couple of UserBadge widgets instead (then it will perform the async call to get the user profile information).  Then it would have the user picture and a link to their profile too.

PR #764, opened 2014-01-07
SWC-1096: Notify (email) users when you share an entity with them

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/sharing/AccessControlListEditor.java
```diff
@@ -458,6 +471,39 @@ public void onFailure(Throwable caught) {
 		}
 	}
 	
+	public void notifyNewUsers() {
```
jay-hodgson, line 476: Make notification optional (with customizable message)

PR #772, opened 2014-01-10
Release 26

jay-hodgson:


jay-hodgson:
this is to pull SWC-1064-bootstrap

PR #816, opened 2014-01-30
SWC-1034: new user/team markdown widget (editor, and renderer that delegates to the UserBadge or TeamBadge) 

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/editor/UserTeamConfigEditor.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+	@Override
+	public void configure(WikiPageKey wikiKey, Map<String, String> widgetDescriptor, Dialog window) {
+		descriptor = widgetDescriptor;
+		String id = descriptor.get(WidgetConstants.USER_TEAM_BADGE_WIDGET_ID_KEY);
+		if (id != null)
+			view.setId(id);
+	}
+	
+	@SuppressWarnings("unchecked")
```
jay-hodgson, line 31: remove SuppressWarnings

src/main/webapp/Portal.css
```diff
@@ -560,7 +560,9 @@ a.shortBtn {
 .margin-right-15 {
```
jay-hodgson, line 560: idea to break up file and organize

PR #835, opened 2014-02-17
Swc-1125: rework data use restrictions UI

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayConstants.java
```diff
@@ -601,7 +608,7 @@
 	public static final String MARKDOWN_WIDGET_WARNING = "Markdown Widget Warning";
 	public static final String MARKDOWN_API_TABLE_WARNING = "API Table Warning";
 	public static final String ACCESS_WILL_BE = "Access Will Be";
-	public static final String ANY_USE = "Any Use";
+	public static final String ANY_USE = "None";
```
jay-hodgson: rename variable

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -562,6 +565,34 @@ public void handleEvent(MessageBoxEvent be) {
 
 	}
 	
+	
+	public static void showYesNoMessage(
```
jay-hodgson: delete. going to use OK dialog

src/main/java/org/sagebionetworks/web/client/IconsImageBundle.java
```diff
@@ -119,9 +119,6 @@
 	@Source("images/icons/shield-green-16.png")
 	ImageResource sheildGreen16();
```
jay-hodgson, line 120: change to simple outline?

PR #844, opened 2014-02-24
submission query service first element is now at offset 0

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableWidget.java
```diff
@@ -250,8 +250,12 @@ private String getColumnValue(JSONObjectAdapter row, String key) throws JSONObje
 	
 	public String getPagedURI(String uri) {
 		//special case for query service
-		if (isQueryService(uri)) {
-			return uri + "+limit+"+tableConfig.getPageSize()+"+offset+"+(tableConfig.getOffset()+1);
+		boolean isSubmissionQueryService = isSubmissionQueryService(uri);
+		boolean isNodeQueryService = isNodeQueryService(uri);
+		if (isSubmissionQueryService || isNodeQueryService) {
+			//the node query service's first element is at index 1! (submission query service first element is at index 0)
+			int firstIndex = isSubmissionQueryService ? 0 : 1;
```
jay-hodgson: use constant instead

PR #845, opened 2014-02-24
Swc-1209: add widget for html5 video

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/editor/VideoConfigViewImpl.java
```diff
@@ -0,0 +1,152 @@
...
+		this.entityFinder = entityFinder;
+	}
+	
+	@Override
+	public void initView() {
+		clear();
+		mp4Entity = new TextBox();
+		oggEntity = new TextBox();
+		webmEntity = new TextBox();
+		add(new HTML("<h5 class=\"margin-top-left-10\">The browser viewing the video will use the first format that it recognizes.</h5>"));
```
dburdick: maybe link to browser support here

PR #850, opened 2014-02-27
Swc-1218: promote team invitations.  show new ui between login and last place, and move notification on front page to above team list

jay-hodgson:


src/test/java/org/sagebionetworks/web/unitclient/presenter/LoginPresenterTest.java
```diff
@@ -247,6 +251,21 @@ public void testSetPlaceChangeUsernameFailure()throws JSONObjectAdapterException
 		verify(mockView).showErrorMessage(eq(exceptionMessage));
 	}
 
+	@Test
+	public void testOpenInvitations() throws JSONObjectAdapterException {
```
jay-hodgson, line 255: remove now that the team invites ui has been moved to 403 entity view

PR #863, opened 2014-03-10
SWC-1228: add additional commands to the markdown editor, with font-awes...

jay-hodgson:
...ome icon subset


src/main/java/org/sagebionetworks/web/client/widget/entity/MarkdownEditorWidget.java
```diff
@@ -274,25 +274,57 @@ public void onClick(ClickEvent event) {
...
-//			@Override
-//			public void componentSelected(ButtonEvent ce) {
-//				handleBasicCommand("--");
-//			}
-//		});
-//		mdCommands.add(strikeCommand);
 
+		startTag = "--";
+		endTag = startTag;
+		com.google.gwt.user.client.ui.Button strikeCommand = getNewCommand("Strikeout", "", getBasicCommandClickHandler(startTag, endTag, true));
```
jay-hodgson: support multiline should be false

PR #865, opened 2014-03-12
SWC-1232: auto-format columns based on name, and strip off "tablename" from node query result column names

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableColumnRendererEpochDate.java
```diff
@@ -47,7 +47,8 @@ public void init(Map<String, List<String>> columnData,
 		for (Iterator iterator2 = colValues.iterator(); iterator2
 				.hasNext();) {
 			String colValue = (String) iterator2.next();
-			Date date = new Date(Long.parseLong(colValue));
+			//parse as a double, and cast to long (truncate toward zero)
+			Date date = new Date((long)Double.parseDouble(colValue));
```
dburdick, line 51: Don't work around it here, fix in service

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableColumnRendererUserId.java
```diff
@@ -125,6 +124,13 @@ public void onFailure(Throwable caught) {
 		});
 	}
 	
+	public String getUserId(String inputUserId) {
```
dburdick, line 127: file issue

PR #887, opened 2014-03-21
side effect of PLFM-2509, catch synapse exception and wrap it in a servlet exception

jay-hodgson:
diff tool has a difficult time with this simple change.  surrounded block with try/catch, also need to update artifact.  Do not pull in until plfm build succeeds (currently pointing to this build: http://hudson-build.sagebase.org:8081/job/PLFM-develop/1066/)

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/FileHandleServlet.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/FileHandleServlet.java
@@ -184,7 +184,6 @@ public class FileHandleServlet extends HttpServlet {
 					resolvedUrl = client.getTeamIcon(teamId, false);
 				} catch (SynapseException e) {
 					return;
-					//throw new ServletException(e);
 				}
 			}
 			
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/server/servlet/FileHandleServlet.java
```diff
@@ -116,94 +116,98 @@ public void service(ServletRequest arg0, ServletResponse arg1)
...
-				HttpResponse newResponse = new HttpClientProviderImpl().execute(httpGet);
-				HttpEntity responseEntity = (null != newResponse.getEntity()) ? newResponse.getEntity() : null;
-				if (responseEntity != null) {
-					responseEntity.writeTo(response.getOutputStream());
+			} else if (teamId != null) {
+				try {
+					resolvedUrl = client.getTeamIcon(teamId, false);
+				} catch (SynapseException e) {
+					return;
+					//throw new ServletException(e);
```
dburdick: this should be removed if it is not the intended behavior.

PR #888, opened 2014-03-21
Swc 1267

dburdick:


src/main/java/org/sagebionetworks/web/client/presenter/HomePresenter.java
```diff
@@ -258,7 +258,7 @@ public void onSuccess(List<MembershipInvitationBundle> result) {
 			
 			@Override
 			public void onFailure(Throwable caught) {
-				onFailure(caught);
+				callback.onFailure(caught);
```
dburdick, line 261: unrelated to this change, but was an issue

PR #902, opened 2014-04-02
SWC-1289: Profile and profile editing cleanup

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/ProfileFormWidget.java
```diff
@@ -61,9 +66,20 @@ public Widget asWidget() {
 	}
 	
 	@Override
-	public void updateProfile(final String firstName, final String lastName, final String summary, final String position, final String location, final String industry, final String company, final String email, final AttachmentData pic, final String teamName, final String url) {				
+	public void updateProfile(final String firstName, final String lastName, final String summary, final String position, final String location, final String industry, final String company, final String email, final AttachmentData pic, final String teamName, final String url, final String userName) {				
 		final UserSessionData currentUser = authenticationController.getCurrentUserSessionData();			
 		if(currentUser != null) {
+				//check for valid url
+				if (!LoginPresenter.isValidUrl(url, true)) {
+					view.showErrorMessage(DisplayConstants.INVALID_URL_MESSAGE);
```
jay-hodgson: add error style to form instead of popup. with a feedback div below.

src/main/resources/org/sagebionetworks/web/client/view/ProfileFormViewImpl.ui.xml
```diff
@@ -0,0 +1,85 @@
...
+		div.center {   
+			margin-left: auto;   
+			margin-right: auto;   
+			width: 8em;
+			padding:15px; 
+		}
+	</ui:style>
+	
+    <g:HTMLPanel>
+		<div class="row">
```
jay-hodgson: need outer row?

PR #924, opened 2014-04-15
timed-retry-widget if api table is unavailable

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableWidgetViewImpl.java
```diff
@@ -203,6 +205,27 @@ public void showError(String message) {
 	}
 	
 	@Override
+	public void showTableUnavailable() {
+		removeAll();
+		FlowPanel unavailableContainer = new FlowPanel();
+		unavailableContainer.addStyleName("jumbotron");
+		unavailableContainer.add(new HTML("<h2>" + DisplayConstants.TABLE_UNAVAILABLE + "</h2>"));
```
jay-hodgson, line 212: Should have more information.  Either pass along message from exception, or say that the data is being updated.

PR #939, opened 2014-04-18
SWC-1339: Share button always available, only editable if you can change permissions

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1977,10 +1977,11 @@ public static void surroundWidgetWithParens(Panel container, Widget widget) {
 		container.add(paren);
```
dburdick, line 1977: Change non-admin window to have only "Close" button, not Save/Cancel
Change button on Entity page to "Sharing" instead of "Share" for non admin users

PR #955, opened 2014-04-29
SWC-1364: fix for large file upload

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
```diff
@@ -47,7 +47,7 @@
 
 	public boolean isDirectUploadSupported();
 	
-	public void uploadFileChunk(String contentType, String fileFieldId, int startByte, int endByte, String url, XMLHttpRequest xhr, ProgressCallback callback);
```
dburdick, line 50: If the intent is remedy a problem with integerspace, you should probably use Long/long instead of double. Here's how GWT handles longs:

long: JavaScript has no 64-bit integral type, so long needs special consideration. Prior to GWT 1.5, the long type was was simply mapped to the integral range of a 64-bit JavaScript floating-point value, giving long variables an actual range less than the full 64 bits. As of GWT 1.5, long primitives are emulated as a pair of 32-bit integers, and work reliably over the entire 64-bit range. Overflow is emulated to match the expected behavior. There are a couple of caveats. Heavy use of long operations will have a performance impact due to the underlying emulation. Additionally, long primitives cannot be used in JSNI code because they are not a native JavaScript numeric type.

PR #969, opened 2014-05-07
SWC-1383 and SWC-1387

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/ButtonLinkWidget.java
```diff
@@ -26,10 +30,20 @@ public void configure(final WikiPageKey wikiKey, final Map<String, String> widge
...
+			isHighlight = Boolean.parseBoolean(descriptor.get(WebConstants.HIGHLIGHT_KEY));
+		}
+		boolean openInNewWindow = isOpenInNewWindow(url);
+		
+		view.configure(wikiKey, buttonText, url, isHighlight, openInNewWindow);
 		descriptor = widgetDescriptor;
 	}
 	
+	public boolean isOpenInNewWindow(String url) {
+		return url != null && !(url.startsWith("#!") || url.startsWith(gwt.getHostPrefix()));
```
brucehoff, line 44: rather than parse the string, use java.net.*??? (or GWT equiv) to parse the URL, then write your logic to use the fields returned by the library.

PR #973, opened 2014-05-09
SWC-1389

jay-hodgson:
Make markdown formatting guide content fed by wiki (and have a Help place token).  Fix Wiki place token for root wiki.  Redirect to wiki place from Synapse place if wiki based entity.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
@@ -15,6 +15,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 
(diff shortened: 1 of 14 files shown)


src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -148,6 +149,8 @@
 	AdapterFactory adapterFactory = new AdapterFactoryImpl();
 	AutoGenFactory entityFactory = new AutoGenFactory();
 	
+	private HashMap<String, org.sagebionetworks.web.shared.WikiPageKey> pageName2WikiKeyMap;
```
eric-wu: Wrap it into AtomicReference if we want to synchronize the methods below? Or use a static initialize.

marcel-blonk: AtomicReference is overkill, volatile would be good enough for this purpose. Just make sure to build the map first in a local variable and then assign to volatile member. Memory barrier around the volatile will ensure ordering.

PR #977, opened 2014-05-14
Swc-860: Settings page rework, focussing on Change Password.

jay-hodgson:
Including dynamically showing change/create Synapse password based on SSO flag.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
@@ -118,7 +118,7 @@ public class LoginPresenter extends AbstractActivity implements LoginView.Presen
 				!LoginPlace.CHANGE_USERNAME.equals(token) && 
 				!"".equals(token) && 
 				token != null) {			
-			loginSSOUser(token);
+			loginUser(token, true);
 		} else {
 			// standard view
 			authenticationController.logoutUser();
@@ -181,7 +181,7 @@ public class LoginPresenter extends AbstractActivity implements LoginView.Presen
 	
 	@Override
 	public void setNewUser(UserSessionData newUser) {
-		loginSSOUser(newUser.getSession().getSessionToken());
+		loginUser(newUser.getSession().getSessionToken(), newUser.getIsSSO());
 	}
 	
 	/**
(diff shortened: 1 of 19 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
```diff
@@ -343,13 +343,14 @@ public void onFailure(Throwable t) {
 		});
 	}
 	
-	private void loginSSOUser(String token) {
+	private void loginUser(String token, boolean isSSO) {
```
brucehoff: would it be correct to rename "loginUser" to "revalidateSession" and "isSSO" to "isLoggedInViaSSO".  Also does this boolean parameter actually matter?

Investigate removing isSSO, since the portal's behavior shouldn't depend on how you authenticated.

src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
```diff
@@ -148,7 +146,6 @@ public void onFailure(Throwable caught) {
 				view.showErrorMessage(DisplayConstants.ERROR_GENERIC_RELOAD);
 			}
 		} else {
-			view.passwordChangeFailed();
```
brucehoff, line 151: is it correct to remove this line?

src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
```diff
@@ -198,6 +195,9 @@ public void onFailure(Throwable caught) {
 		});
 	}
 	
+	public void updateShowCreateSynapsePassword() {
```
brucehoff: this will be removed

src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
```diff
@@ -258,6 +258,7 @@ private void showView(Settings place) {
 		//String token = place.toToken();
```
brucehoff, line 230: will remove

src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
jay-hodgson, line 151: verified this is correct.  
The following 2 lines pop up an alert message and send user to the login place.  An error on the password change form is not necessary (and will not be seen)

PR #1015, opened 2014-06-09
SWC-1467: challenge signup process to include additional information

jay-hodgson:


src/main/resources/org/sagebionetworks/web/Portal.gwt.xml
```diff
@@ -35,6 +35,7 @@
 	<set-configuration-property name="devModeRedirectEnabled" value="true"/>
 	
     <!-- Uncomment to limit compilation permutations to FF <set-property name="user.agent" value="gecko1_8" /> -->
+ 	<set-property name="user.agent" value="safari" />
```
eric-wu: shouldn't be here, Jay says

PR #1018, opened 2014-06-12
Created new dialog style, and consolidated error, info, and confirm popups to use the new style

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -598,32 +599,120 @@ public void onFailure(Throwable caught) {
...
+		dialog.add(content);
+		dialog.setHeading(title);
+		FlowPanel buttonPanel = new FlowPanel();
+		buttonPanel.setHeight("50px");
+		buttonPanel.addStyleName("whiteBackground");
+		boolean isSecondaryButton = secondaryButtonText != null;
+
+		com.google.gwt.user.client.ui.Button continueButton = DisplayUtils
+				.createButton(primaryButtonText, ButtonType.PRIMARY);
+		continueButton.addStyleName("right margin-right-10 margin-bottom-10");
```
john-hill, line 680: close should call cancel

PR #1043, opened 2014-06-25
Swc-1486: use display style directly instead of setting via css

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/WikiSubpagesViewImpl.java
```diff
@@ -83,29 +83,44 @@ public void onClick(ClickEvent event) {
...
-		DisplayUtils.hide(ulContainer);
-		showHideButton.setStyleName("btn btn-default btn-xs left");
+		
+		if (ulContainer != null)
+			DisplayUtils.hide(ulContainer);
+		
+		if (wikiPageContainer != null) {
+			wikiPageContainer.setStyleName(HIDE_SUBPAGES_MD_STYLE);
+			wikiPageContainer.setVisible(false);
+			wikiPageContainer.setVisible(true);
```
jay-hodgson, line 104: if this does not work, remove!

PR #1049, opened 2014-07-08
Swc-64: finally expose the ability to change your email address in settings

jay-hodgson:


eric-wu:
UX is somewhat different from what's implemented. Maybe review the UX through Mike?

src/main/java/org/sagebionetworks/web/client/view/AccountViewImpl.java
```diff
@@ -0,0 +1,89 @@
...
+	@Override
+	public void showErrorMessage(String message) {
+		DisplayUtils.showErrorMessage(message);
+	}
+	
+	@Override
+	public void showErrorInPage(String title, String message) {
+		clear();
+		mainContainer.add(new HTML(DisplayUtils.getWarningHtml(title, message)));
+		DisplayUtils.showErrorMessage(message);
```
eric-wu: remove popup

PR #1051, opened 2014-07-10
view changes to fix SWC-1514, SWC-1517, and SWC-1519

jay-hodgson:


john-hill:
Reminder: Merge into develop as well.

PR #1057, opened 2014-07-14
Swc-1522: add new version of file to FileEntity if names match

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -2408,13 +2410,35 @@ public String getUploadDaemonStatus(String daemonId) throws RestServiceException
...
 			//create entity if we have to
 			FileEntity fileEntity = null;
-			
+			FileHandle newHandle = synapseClient.getRawFileHandle(fileHandleId);
+			if (entityId == null) {
+				//file entity not set
+				//determine if we should create a new file entity, or update an existing.
+				if (parentEntityId != null) {
+					//look for a child with the same file name
+					EntityIdList list = synapseClient.getDescendants(parentEntityId, Integer.MAX_VALUE, null);
```
xschildw: Does getDescendants() only return immediate children?

PR #1060, opened 2014-07-15
SWC-1496: World Clock Tools

RadicalZephyr:
This is my start so far.  This isn't really ready for merge-in, but I wanted to check in that the architecture I'm moving towards is appropriate and makes sense.

Basically, I'm creating a GWT RPC service to handle the needed time related requests, and I've added a simple data class for transporting the needed information between server and client (`DateTime`).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,14 +104,14 @@
 	          <exclude>**/*.properties</exclude>
 	        </excludes>
 	      </resource>
-	    </resources>	    
+	    </resources>
 
     		<!-- Generate compiled stuff in the folder used for developing mode -->
 		<outputDirectory>${webappDirectory}/WEB-INF/classes</outputDirectory>
 
 		<pluginManagement>
 			<plugins>
-				<!-- Tell the eclipse m2e plugin to execute these lifecycle mappings 
+				<!-- Tell the eclipse m2e plugin to execute these lifecycle mappings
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -561,47 +569,57 @@
...
+		</dependency>
+
+    <!-- Better time support for the World Clock widget (SWC-1496) -->
 		<dependency>
-		    <groupId>com.atlassian.jira</groupId>
-		    <artifactId>jira-rest-java-client</artifactId>
-		    <version>2.0.0-m2</version>
+			<groupId>org.threeten</groupId>
+			<artifactId>threetenbp</artifactId>
+			<version>1.0</version>
```
jay-hodgson, line 631: 3-clause revised BSD license

jay-hodgson:
Yep, makes sense.  I envision using many deprecated methods in the emulated java.util.Date on the client to make this work - but I think it's necessary given the state of gwt time support.  Good start.

pom.xml
RadicalZephyr, line 631: Is this a problem, or do we just need to comply with the license?  Because we can also use joda-time; I picked threeten because joda's API isn't as nice and this has a clear upgrade path for if we move to java 8.

RadicalZephyr:
From what I've seen, the general recommendation seems to be to avoid doing any date/time-zone work on the client.  So I was planning on having the server format the date, and then the client will just use string formatting and the simple data class to keep the clock current.

PR #1063, opened 2014-07-15
SWC-1523: cool down link color

jay-hodgson:


jay-hodgson:
Could be jarring on the first visit. Consensus (with Mike/Shannon/Brian) is that this is an improvement, so pulling this in.  Trivial to back out if there's a backlash on staging next week.

PR #1078, opened 2014-07-24
update jsplumb and load jquery from standard cdn

jay-hodgson:


src/main/webapp/Portal.html
```diff
@@ -36,14 +36,15 @@
 		<script>
 	        var start = new Date();
 		        head
-		        .js("http://ajax.googleapis.com/ajax/libs/jqueryui/1.11.0/jquery-ui.min.js")
+		        .js("//code.jquery.com/jquery-1.11.0.min.js")
+		        .js("//ajax.googleapis.com/ajax/libs/jqueryui/1.11.0/jquery-ui.min.js")
```
john-hill: we should get this direct from jquery as well:
//code.jquery.com/ui/1.11.0/jquery-ui.min.js

PR #1089, opened 2014-07-29
SWC-427

toddis:
Trash Can UI. Table to be reimplemented.

Append ?w=1 to the URL to ignore the trivial whitespace differences.


src/main/java/org/sagebionetworks/web/client/DisplayConstants.java
```diff
@@ -172,6 +172,7 @@
 	public static final String ERROR_API_TABLE_RENDERER_MISSING_INPUT_COLUMN = "Specified input column is missing from the service response: ";
 	public static final String ERROR_GETTING_PERMISSIONS_TEXT = "READ ONLY MODE. Reason: An error occurred in retrieving your level of access.";	
 	public static final String ERROR_ACL_RETRIEVAL_FAILED = "Retrieval of sharing settings failed. Please try again.";
+	public static final String ERROR_RESTORING_TRASH_PARENT_NOT_FOUND = "An error occured in restoring this entity from the trash. Perhaps the original parent is also in the trash or deleted?";
```
jay-hodgson, line 175: I would maintain the current voice of error messages.  Something like "Sorry, you cannot restore this item because the original parent (syn12345) cannot be found."

src/main/java/org/sagebionetworks/web/client/presenter/BulkPresenterProxy.java
```diff
@@ -42,22 +43,23 @@
...
+ * correct presenter given the place
  * 
  * @author Dave
- *
+ * 
  */
 public class BulkPresenterProxy extends AbstractActivity {
 
-	private static Logger log = Logger.getLogger(BulkPresenterProxy.class.getName());
+	private static Logger log = Logger.getLogger(BulkPresenterProxy.class
```
jay-hodgson, line 54: I find the longer line versions are more readable than the multiline.  
Please revert these Eclipse auto-formatter newline changes.

src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
```diff
@@ -0,0 +1,201 @@
...
+		if (offset == null)
+			this.offset = 0;
+		else
+			this.offset = offset;
+		synapseClient.viewTrashForUser(this.offset, TRASH_LIMIT, new AsyncCallback<String>() {
+			@Override
+			public void onSuccess(String result) {
+				
+				try {
+					trashList = nodeModelCreator.createPaginatedResults(result, TrashedEntity.class);
```
jay-hodgson, line 102: Note that this is using NodeModelCreator which uses JSONEntityFactoryImpl which uses AutoGenFactory.  This will cause a very large download.  Since this is the Trash can place, I think it's ok.  It would cause a significant performance problem if same code were added to the Home page (or Profile/dashboard page)

src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
```diff
@@ -0,0 +1,201 @@
...
+	 * Private Methods
+	 */
+	
+	private void showView(Trash place) {
+		Integer offset = place.getStart();
+		getTrash(offset);
+	}
+	
+	private void createFailureDisplay(Throwable caught) {
+		if (!DisplayUtils.handleServiceException(caught, globalAppState, authController.isLoggedIn(), view)) {
```
jay-hodgson, line 197: Think about reworking so that the caught.getMessage() is sent to the view, and the view shows the error in an alert in the page (along with a button to reload?). Currently you always pop up an error, and I would guess this page looks fairly empty after clicking OK on the error dialog.

src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
```diff
@@ -0,0 +1,201 @@
...
+				view.showInfo(TRASH_RESTORED_TITLE, trashedEntity.getEntityName());
+				view.refreshTable();
+			}
+			
+			@Override
+			public void onFailure(Throwable caught) {
+				if (caught instanceof NotFoundException) {
+					view.showErrorMessage(DisplayConstants.ERROR_RESTORING_TRASH_PARENT_NOT_FOUND);
+				} else {
+					createFailureDisplay(caught);
```
jay-hodgson, line 166: But I think a popup error would be ok for this error

src/main/java/org/sagebionetworks/web/client/view/TrashViewImpl.ui.xml
```diff
@@ -0,0 +1,28 @@
...
+	xmlns:s="urn:import:org.sagebionetworks.web.client.view">
+	
+	<g:HTMLPanel>
+		<div class="color-line"></div>
+		<g:SimplePanel ui:field="header" />
+		<div class="separator margin-top-10-imp "></div>
+		<div class="row margin-top-left-10 margin-right-15 margin-bottom-20">
+	        <!-- <div class="row"> -->
+				<div class="well col-md-offset-3 col-md-6 col-sm-offset-2 col-sm-8">
+				    <g:Button ui:field="deleteAllButton" styleName="btn btn-primary btn-lg margin-bottom-10 right"/>
```
jay-hodgson, line 13: Feel free to set the button text here in the template instead of in the code.  text="Delete Selected" (In general, I like the idea of setting up the UI as much as we can in the templates, both for performance and readability).

PR #1153, opened 2014-08-22
SWC-1639: set highlight box titles

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -505,14 +505,16 @@ public void updateTeamInvites(List<MembershipInvitationBundle> invites) {
 		if (invites != null && invites.size() > 0) {
 			teamNotificationCount += invites.size();
 			//update team notification count
-			view.setTeamNotificationCount(Integer.toString(teamNotificationCount));
+			if (teamNotificationCount > 0)
```
jay-hodgson, line 508: add tests for team notification count

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -314,6 +320,23 @@ public void invoke() {
 		DisplayUtils.show(navtabContainer);
 	}
 	
+	private void resetHighlightBoxes() {
+		setHighlightBoxUser(null);
+	}
+	private void setHighlightBoxUser(String displayName) {
```
jay-hodgson: move to DisplayUtils as static method

PR #1157, opened 2014-08-26
support direct upload in IE

jay-hodgson:


jay-hodgson:
Internet Explorer 11 on Windows 8.1 already sends a User-Agent string that does not contain msie "Mozilla/5.0 (Windows NT 6.3; Trident/7.0; rv:11.0) like Gecko"

PR #1166, opened 2014-08-29
SWC-1671

toddis:
Couldn't find existing code that sorted entity headers, so just passed a simple comparator to Collections.sort.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityTreeBrowserViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityTreeBrowserViewImpl.java
@@ -1,6 +1,8 @@
 package org.sagebionetworks.web.client.widget.entity.browse;
 
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityTreeBrowserViewImpl.java
```diff
@@ -98,9 +100,21 @@ public void clear() {
...
+		if (sort) {
+			// TODO: Find already existing sorting code. Reuse that.
+			Collections.sort(rootEntities, new Comparator<EntityHeader>() {
+
+				@Override
+				public int compare(EntityHeader arg0, EntityHeader arg1) {
+					return arg0.getName().toLowerCase().compareTo(arg1.getName().toLowerCase());
+				}
+				
+			});
```
jay-hodgson: Please pull out the same code in HomePresenter.getChallengeProjectHeaders() into a new method.  
You could run that new method in HomePresenter before handing the entity headers to the entitytreebrowser, and remove sort from entitytreebrowser.  Alternatively, you can deal with the sort in the EntityTreeBrowser presenter (and test it there).  Sorting the collection should not take place in the view.

PR #1181, opened 2014-09-08
SWC-1382

toddis:
Added Drag & Drop functionality to file uploader.


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -267,6 +267,54 @@ private static native String _getFilesSelected(String fileFieldId) /*-{
 	    return out;
 	}-*/;
 	
+	@Override
+	public void addDropZoneStyleEventHandling(String fileFieldId, String dropStyleName) {
+		_addDropZoneStyleEventHandling(fileFieldId, dropStyleName);
+	}
+	
+	private static native void _addDropZoneStyleEventHandling(String fileFieldId, String dropStyleName) /*-{
+		$doc.addEventListener("dragover", function( event ) {
```
jay-hodgson, line 282: It appears that every time the uploader is created, a new eventlistener for each event is added to the document (for each of these event types). Would this work if you targeted the actual element?  Like fileFieldElement.addEventListener("dragover", anonymousFunction, false)?  If not, another option is to init once, and have it reference a global (static) fileFieldId (I think we will only want to style a single drag and drop zone in the page)?

toddis, line 282: One thing stopping me from doing something like "fileFieldElement.addEventListener("dragover", anonymousFunction, false)" is this first event listener. It targets the document and prevents the default dragover behavior, which changes the cursor to have the green plus. I need that listener on the document to prevent the green plus cursor from showing up when not in the drop box.

PR #1190, opened 2014-09-11
Swc 1685 Table pagination

john-hill:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/QueryResultsListner.java
```diff
@@ -0,0 +1,12 @@
+package org.sagebionetworks.web.client.widget.table.v2.results;
+
+import org.sagebionetworks.web.client.widget.pagination.PageChangeListener;
+
+/**
+ * 
+ * @author John
+ *
+ */
+public interface QueryResultsListner extends QueryExecutionListener, PageChangeListener{
```
jay-hodgson, line 10: sp: expected QueryResultsListener

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWidget.java
```diff
@@ -25,10 +25,10 @@
  * @author jmhill
  *
  */
-public class TableQueryResultWidget implements TableQueryResultView.Presenter, IsWidget {
+public class TableQueryResultWidget implements TableQueryResultView.Presenter, IsWidget, PageChangeListener {
 	public static final String QUERY_CANCELED = "Query canceled";
 	// Mask to get all parts of a query.
-	private static final Long ALL_PARTS_MASK = new Long(0x15);
+	private static final Long ALL_PARTS_MASK = new Long(255);
```
jay-hodgson, line 31: How is this mask derived?

src/main/resources/org/sagebionetworks/web/client/widget/table/v2/results/TablePageViewImpl.ui.xml
```diff
@@ -3,14 +3,20 @@
...
+			<t:Table styleName="table table-striped table-condensed">
+				<t:THead>
+					<t:TableRow ui:field="header">
+					</t:TableRow>
+				</t:THead>
+				<t:TBody ui:field="body"></t:TBody>
+			</t:Table>
+		</g:ScrollPanel>
+		<b:Row>
+			<b:Column size="XS_2" offset="XS_5">
```
jay-hodgson, line 17: A shame the center-block helper css class didn't do this for you.  We may want to play with it a bit more in the future to see if we can get it working (seems like a cleaner solution to me).

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWidget.java
john-hill, line 31: The portal always needs all parts.  The API documentation is incorrect as 0x15 does not yield all parts (hex 0x15 = 10101 bin while 15 dec = 1111 bin) .  Since  255 dec  =  11111111 bin that is what used.
We could break this out as each part mask binary or'ed together.

src/main/resources/org/sagebionetworks/web/client/widget/table/v2/results/TablePageViewImpl.ui.xml
john-hill, line 17: I tried every combination of "ceneter-block" I could think of.  I even exhausted all suggestions from stack-overflow.  Nothing seemed to have any effect (the widget would always be left-aligned).
If we can figure out how to get the center-block to work I would gladly switch.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/QueryResultsListner.java
john-hill, line 10: This will be removed. The page widget needs the QueryExecutionListener and PageChangeListener to be separate because it disables selection when the former is null and hides pagination when the latter is null.  Currently the view has pagination visible and hides selection while the editor has selection visible and hides pagination.  In the future we will add pagination to the editor as well.

PR #1225, opened 2014-09-25
Swc 1702

john-hill:


src/main/java/org/sagebionetworks/web/client/widget/table/KeyboardNavigationHandlerImpl.java
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	private void bindEditor(final IsWidget editor) {
+		// Can only listen to widget that implement HasKeyDownHandlers
+		if(editor instanceof HasKeyDownHandlers){
+			HasKeyDownHandlers keyDownCell = (HasKeyDownHandlers) editor;
+			keyDownCell.addKeyDownHandler(new KeyDownHandler() {
+				@Override
+				public void onKeyDown(KeyDownEvent event) {
+					editorKeyPressed(editor, event);
```
jay-hodgson, line 58: Concerned about editor listener explosion.  Could add a unique editor set and check before adding another key down handler.  Or keep reference to dereference objects, and call on removeAll

src/main/java/org/sagebionetworks/web/client/widget/table/KeyboardNavigationHandlerImpl.java
```diff
@@ -0,0 +1,204 @@
...
+			break;
+		}
+	}
+
+	/**
+	 * Give focus to the cell that is right of the passed cell.
+	 * @param editor
+	 */
+	private void onRight(IsWidget editor) {
+		Address current = cellAddressMap.get(editor);
```
jay-hodgson, line 140: check for null Address

PR #1234, opened 2014-10-06
Swc-1750: add support for post_message access restriction type when joining a team.

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/utils/GovernanceServiceHelper.java
```diff
@@ -14,25 +16,55 @@
 import com.google.gwt.user.client.rpc.AsyncCallback;
 
 public class GovernanceServiceHelper {
+	
+	public static EntityWrapper getTermsOfUseAccessApproval(
```
jay-hodgson: change to SelfSignAccessRequirement, collapse into one.

src/main/java/org/sagebionetworks/web/client/widget/team/JoinTeamWidget.java
```diff
@@ -236,23 +231,33 @@ public void sendJoinRequestStep2() {
...
+					setLicenseAccepted(accessRequirement);
 				}
 			};
 			//pop up the requirement
 			view.updateWizardProgress(currentPage, getTotalPageCount());
-			view.showAccessRequirement(text, termsOfUseCallback);
+			APPROVAL_TYPE type = GovernanceServiceHelper.accessRequirementApprovalType(accessRequirement);
+			if (APPROVAL_TYPE.USER_AGREEMENT.equals(type) || APPROVAL_TYPE.ACT_APPROVAL.equals(type)) {
+				String text = GovernanceServiceHelper.getAccessRequirementText(accessRequirement);
+				view.showTermsOfUseAccessRequirement(text, termsOfUseCallback);
```
brucehoff, line 248: Seems like you shouldn't be calling 'showTermsOfUseAccessRequirement' for an ACT requirement.  Instead you should call something like view.showACTAccessRequirement(text); (i.e. there is no 'accept' button).

src/main/java/org/sagebionetworks/web/client/widget/team/JoinTeamWidgetViewImpl.java
```diff
@@ -233,7 +241,54 @@ public void showAccessRequirement(
...
+	}
+
+	@Override
+	protected void onAttach() {
+		//register to listen for the "message" events
+		if (messageHandler == null) {
+			messageHandler = EventHandlerUtils.addEventListener("message", EventHandlerUtils.getWnd(), new JavaScriptCallback() {
+				
+				@Override
+				public void invoke(JavaScriptObject event) {
```
jay-hodgson, line 274: log message being returned to the console.

PR #1255, opened 2014-10-20
Swc-1662: show where files will be uploaded to (Synapse storage, or sftp domain)

jay-hodgson:
Include upload destination banner text (if filled in)


src/main/java/org/sagebionetworks/web/client/DisplayConstants.java
```diff
@@ -778,6 +778,8 @@
 	public static final String UPDATE_FILE = "Update File";
 	public static final String SESSION_TIMEOUT = "Session Timeout";
 	public static final String SESSION_HAS_TIMED_OUT = "Your session has timed out. Please login again.";
+	
+	public static final String UPLOAD_DESTINATION = "All uploaded files will be stored on ";
```
brucehoff, line 782: Shouldn't it be "stored in" rather than "stored on"?

PR #1267, opened 2014-10-23
Swc 1212

john-hill:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/QueryInputViewImpl.java
```diff
@@ -41,6 +41,10 @@
 	Alert queryResultsMessage;
 	@UiField
 	Button resetButton;
+	@UiField
+	Button editResultsButton;
```
jay-hodgson, line 45: cog button needs a tooltip (at least).

PR #1271, opened 2014-10-24
SWC-1836: fix upload server logging

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -567,7 +567,7 @@ public void logErrorToRepositoryServices(String message) throws RestServiceExcep
 		try {
 			org.sagebionetworks.client.SynapseClient synapseClient = createSynapseClient();
 			LogEntry entry = new LogEntry();
-			entry.setLabel("Synapse web client error");
+			entry.setLabel("SWC: " + message);
```
brucehoff: try to ensure that 'message' has a small number of (informative) variations, not a large number of varations. E.g. it should not include an object ID.

PR #1289, opened 2014-11-03
SWC-1863, SWC-1866

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
```diff
@@ -219,16 +219,24 @@ public void onFailure(Throwable caught) {
...
 		if (url == null)
 			return null;
 		if (!url.toLowerCase().startsWith(WebConstants.SFTP_PREFIX)) {
 			throw new IllegalArgumentException("Not a sftp url: " + url);
 		}
 		String domain = url.substring(WebConstants.SFTP_PREFIX.length());
-		int slashIndex = domain.indexOf("/");
-		if (slashIndex != -1) {
-			domain = domain.substring(0, slashIndex);
+		//if a port is specified, then find the colon
```
john-hill: let a "proper" URL parser do this work (server-side)

PR #1307, opened 2014-11-10
Swc 1870

john-hill:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/RowSetUtils.java
```diff
@@ -127,6 +127,42 @@ private static PartialRow buildPartialRow(List<String> headers,
...
+			}
+			return true;
+		}else{
+			return !tOriginal.equals(tChanged);
+		}
+	}
+	/**
+	 * @return Returns null if the passed value is null.  Returns null if the trimmed string is empty else the trimmed string.  
+	 */
+	public static String trimWithEmptyAsNull(String toUpdate){
```
jay-hodgson, line 153: Generally useful outside of tables.  New String Util class?

PR #1311, opened 2014-11-13
SWC-1903: Trash Error Message

toddis:
Made trash error message (restore when parent not found) display the message of the returned exception. Fixed tests.


src/test/java/org/sagebionetworks/web/unitclient/presenter/TrashPresenterTest.java
```diff
@@ -132,12 +132,16 @@ public void testRestoreFailure() {
...
 	
 	@Test
 	public void testRestoreParentNotFoundFailureRestoreCall() {
+		ForbiddenException ex = mock(ForbiddenException.class);
+		when(ex.getMessage()).thenReturn("Message");
 		AsyncMockStubber.callFailureWith(new ForbiddenException()).when(mockSynapse).restoreFromTrash(
 				anyString(), anyString(), any(AsyncCallback.class));
+		when(mockAuthenticationController.isLoggedIn()).thenReturn(true);
 		presenter.restoreEntity(trashList.getResults().get(0));
 		verify(mockView).showErrorMessage(anyString());
```
jay-hodgson: capture string being sent to the view, and assert that it contains your exception message (this should fail before your fix, but succeed after)

PR #1323, opened 2014-11-20
Swc 1840

john-hill:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/cell/EnumCellEditorViewImpl.java
```diff
@@ -0,0 +1,72 @@
...
+ * 
+ * @author jhill
+ *
+ */
+public class EnumCellEditorViewImpl implements EnumCellEditorView {
+
+	public interface Binder extends UiBinder<Widget, EnumCellEditorViewImpl> {}
+	
+	@UiField
+	Select select;
```
jay-hodgson, line 22: I'm not a fan of the "Nothing selected" text for no selection, and lack of ability to clear the value.  If users agree, then we may need to add an empty cell value.

PR #1324, opened 2014-11-20
SWC-1848: add project filters to the dashboard!

jay-hodgson:
also changed to default to My Projects (projects that I created), and added loading UI to show as the profile is downloaded.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
@@ -1,8 +1,10 @@
 package org.sagebionetworks.web.client.view;
 
+import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -231,6 +255,55 @@ public void onClick(ClickEvent event) {
...
+	}
+	
+	@Override
+	public void setMyProjectFilterSelected() {
+		myProjectsFilter.setValue(true);
+	}
+	
+	@Override
+	public void setTeamProjectFilterSelected(Team team) {
+		for (Radio teamFilter : teamFilters) {
```
jay-hodgson, line 291: could move search to the presenter

PR #1354, opened 2014-12-05
SWC-1903: Parent in trash restore message

toddis:
Changed message when restoring to parent that is in trash. I no longer call handleServiceException, which will prepend a message about insufficient privileges to the error message.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
@@ -172,7 +172,7 @@ public class TrashPresenter extends AbstractActivity implements TrashView.Presen
 				if (caught instanceof ForbiddenException) {
 					// Show to view, as handleServiceException call in createFailureDisplay
 					// will show message about insufficient privileges.
-					view.displayFailureMessage(DisplayConstants.ERROR_RESTORING_TRASH_PARENT_NOT_FOUND, caught.getMessage());
+					view.showErrorMessage(caught.getMessage());
 				} else {
 					createFailureDisplay(ERROR_RESTORING_ENTITY_TITLE, caught);
 				}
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
```diff
@@ -170,7 +170,9 @@ public void onSuccess(Void result) {
 			@Override
 			public void onFailure(Throwable caught) {
 				if (caught instanceof ForbiddenException) {
-					createFailureDisplay(DisplayConstants.ERROR_RESTORING_TRASH_PARENT_NOT_FOUND, caught);
+					// Show to view, as handleServiceException call in createFailureDisplay
+					// will show message about insufficient privileges.
+					view.showErrorMessage(caught.getMessage());
```
jay-hodgson: This means ERROR_RESTORING_TRASH_PARENT_NOT_FOUND is not used.  Do you want to prefix this exception message with ERROR_RESTORING_TRASH_PARENT_NOT_FOUND?

src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
```diff
@@ -170,7 +170,9 @@ public void onSuccess(Void result) {
 			@Override
 			public void onFailure(Throwable caught) {
 				if (caught instanceof ForbiddenException) {
-					createFailureDisplay(DisplayConstants.ERROR_RESTORING_TRASH_PARENT_NOT_FOUND, caught);
+					// Show to view, as handleServiceException call in createFailureDisplay
+					// will show message about insufficient privileges.
+					view.showErrorMessage(DisplayConstants.ERROR_RESTORING_TRASH_PARENT_NOT_FOUND + " " + caught.getMessage());
```
toddis, line 175: Ahh, I had originally had that as a title to an error panel. Prepended!

PR #1355, opened 2014-12-05
include gwt-codeserver jar dependency for SuperDevMode

jay-hodgson:


pom.xml
```diff
@@ -441,6 +441,11 @@
 			<scope>provided</scope>
 		</dependency>
 		<dependency>
+			<groupId>com.google.gwt</groupId>
+			<artifactId>gwt-codeserver</artifactId>
+			<version>${gwtVersion}</version>
+		</dependency>
```
john-hill, line 448: shouldn't this have a scope of runtime?

jay-hodgson, line 448: Yes, yes it should.

PR #1367, opened 2014-12-16
SWC-1973

brucehoff:


pom.xml
```diff
@@ -16,7 +16,7 @@
 		<maven.compiler.target>1.6</maven.compiler.target>
 		<webappDirectory>${project.build.directory}/${project.build.finalName}</webappDirectory>
 
- 		<synapse.version>2014-12-12-1527-d50496c</synapse.version>
+ 		<synapse.version>71.0</synapse.version>
```
john-hill: Version 71.0 of PLFM will not work.  It must be 71.0-9-g00f3761 or newer.

PR #1395, opened 2015-01-09
SWC-1947: remove favorites tab, add favorites filter.  convert filters i...

jay-hodgson:
...nto a bar (essentially radio buttons, and a drop down button for teams)


src/test/java/org/sagebionetworks/web/unitclient/presenter/ProfilePresenterTest.java
```diff
@@ -460,24 +467,6 @@ public void testGetHeadersFromQueryResults() {
 	}
 	
 	@Test
-	public void testApplyFilterClickedTeam() {
```
jay-hodgson, line 463: add Team filter test if not already there.

PR #1403, opened 2015-01-16
Swc-2019: add alpha set project background command, and modify wiki editors...

jay-hodgson:
Image and Attachment wiki editors reworked so that they no longer use filehandleservlet call for upload, they do a direct upload (by using a FileInputWidget).  Also rewrite Image editor to use bootstrap components (no gxt components), and ui xml based.


src/main/java/org/sagebionetworks/web/client/widget/entity/editor/ImageConfigViewImpl.java
```diff
@@ -190,113 +142,50 @@ public void setAlignment(String alignment) {
...
+	
 	@Override
 	public void checkParams() throws IllegalArgumentException {
 		if (isExternal()) {
-			if (!urlField.isValid())
-				throw new IllegalArgumentException(urlField.getErrorMessage());
-			if (!nameField.isValid())
-				throw new IllegalArgumentException(nameField.getErrorMessage());
-
+			if (!LoginPresenter.isValidUrl(urlField.getValue(), false))
```
jay-hodgson, line 181: this should be moved to the presenter

PR #1411, opened 2015-01-27
Swc 2038

john-hill:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/QueryResultEditorViewImpl.java
```diff
@@ -127,4 +161,50 @@ public void setErrorMessageVisible(boolean visible) {
...
+	}
+
+	@Override
+	public void setProgressPanelVisible(boolean visible) {
+		this.progressPanel.setVisible(visible);
+	}
+	
+	@Override
+	public void showConfirmDialog(String message, final Callback callback) {
+		Bootbox.confirm(message, new ConfirmCallback() {
```
jay-hodgson, line 181: This looks like a snippet of code that has the potential to be copy-pasted all over the app, consider refactoring to a utility.

PR #1418, opened 2015-02-02
SWC-2061: create a file browser for both entity and project (do not shar...

jay-hodgson:
...e), and directly inform listener of entity selected event


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityTreeBrowser.java
```diff
@@ -246,7 +227,8 @@ public void clearRecordsFetchedChildren() {
 	 * Private Methods
 	 */
 	private void fireEntitySelectedEvent() {
-		handlerManager.fireEvent(new EntitySelectedEvent());
```
jay-hodgson, line 249: This should be a place change, rather than communicating this back up to the entitypagetop

PR #1433, opened 2015-02-10
SWC-2086: new banner for front page

jay-hodgson:


jay-hodgson:
Not to be merged until graphic is updated, and should only be included in code deployed on 2/16.

PR #1456, opened 2015-02-25
SWC-2115: part 1: remove erratic bootstrap extras Select component (adds additional components to the DOM that it does not clean up)

jay-hodgson:


john-hill:
Can we remove the dependency?

jay-hodgson:
Almost.  Still one reference left in
org.sagebionetworks.web.client.widget.table.v2.schema.ColumnModelTableRowEditorViewImpl

PR #1465, opened 2015-02-26
SWC-2115: remove participants link.  add icon to challenge badge, and fi...

jay-hodgson:
...x app favicon so that it i svisible


john-hill:
Okay.

PR #1466, opened 2015-02-26
SWC-2124: include host in username and password text fields

jay-hodgson:


john-hill:
I still against 3rd party credentials flowing through Synapse.

PR #1480, opened 2015-03-03
show individual and team selection UI when there are no teams to make it...

jay-hodgson:
... clear an individual submission will be made


brucehoff:
The plan is to refine the UI for submission further and that this is the first pass.

PR #1483, opened 2015-03-04
Swc-2129: add hide/show profile button to your dashboard

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -192,6 +200,51 @@ private void proceed() {
...
+	
+	private void setIsProfileVisible(boolean isVisible, boolean isSetCookie) {
+		if (isVisible){
+			view.showProfile();
+		} else {
+			view.hideProfile();
+		}
+		view.setShowProfileButtonVisible(!isVisible);
+		view.setHideProfileButtonVisible(isVisible);
+		if (isSetCookie) {
```
john-hill: Setting the cookie could be its own method called only when the show/hide buttons are pressed.

PR #1492, opened 2015-03-06
fix SWC-2155, SWC-2156, SWC-2109

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
```diff
@@ -79,6 +79,7 @@ public void showView(final LoginPlace place) {
 		String token = place.toToken();
 		if(LoginPlace.LOGOUT_TOKEN.equals(token)) {
 			authenticationController.logoutUser();
+			globalApplicationState.clearLastPlace();
```
john-hill: There are cases where this is not desired.

PR #1539, opened 2015-03-25
ready after plfm build

john-hill:


pom.xml
```diff
@@ -16,7 +16,7 @@
 		<maven.compiler.target>1.6</maven.compiler.target>
 		<webappDirectory>${project.build.directory}/${project.build.finalName}</webappDirectory>
 
- 		<synapse.version>2015-03-21-1670-98c2de9</synapse.version>
+ 		<synapse.version>develop-SNAPSHOT</synapse.version>
```
jay-hodgson: Needs updated Synapse version.

PR #1540, opened 2015-03-25
SWC-2210: rewrite wiki attachments dialog

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/WikiAttachmentsViewImpl.java
```diff
@@ -1,233 +1,115 @@
...
+		modal.hide();
 	}
 	
 	@Override
 	public void configure(WikiPageKey wikiKey, List<FileHandle> list) {		
-		gridStore.removeAll();
+		reset();
 		if(list == null || list.size() == 0){
-			addNoAttachmentRow();
+			showNoAttachmentRow();
```
jay-hodgson: move to presenter

src/main/java/org/sagebionetworks/web/client/widget/entity/WikiPageWidgetViewImpl.java
```diff
@@ -462,6 +462,8 @@ public void invoke() {
 				});
 				
 				add(markdownEditorWidget.asWidget());
+				//also add the wiki attachments dialog
+				add(wikiAttachments.asWidget());
```
jay-hodgson: should be manager, where you can add attachments, delete, and choose to save or discard changes.

jay-hodgson: Get rid of wiki page modal, and instead incorporate into Insert Attachments and Insert Image wiki editors.

PR #1551, opened 2015-03-30
Completion of GXT removal (snapshot/summary widget) and update to GWT 2.7.0

jay-hodgson:
Refactor resources so that build does not rely on ClassLoader to locate.  
No need to manually copy the code server RPC policies after gwt update.  
Code clean up and fix web.xml warnings.

NOTE: NOT TO BE PULLED UNTIL AFTER SUMMARY ENTITIES ARE MIGRATED.


src/main/java/org/sagebionetworks/web/client/transform/JSONEntityFactoryImpl.java
```diff
@@ -18,13 +19,14 @@
 public class JSONEntityFactoryImpl implements JSONEntityFactory {
 	// This factory allows the creation of new class instances using only the class name.
 	// Since reflection is not allowed in GWT, it uses an auto-generated swtich.
-	AutoGenFactory internalFactory;
+	AutoGenFactory internalFactory = null;
```
jay-hodgson: this doesn't help because it's still in imports.

jay-hodgson: need to remove all uses of PaginatedResults, or remove autogen factory from PaginatedResults

PR #1562, opened 2015-04-06
SWC-2247: certification check, and rework annotation editor

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/annotation/AnnotationsRendererWidget.java
```diff
@@ -70,6 +73,11 @@ public void setEntityUpdatedHandler(EntityUpdatedHandler handler) {
 
 	@Override
 	public void onEdit() {
-		editorDialog.configure(bundle, entityUpdatedHandler);
+		preflightController.checkUploadToEntity(bundle, new Callback() {
```
john-hill, line 76: check update

PR #1563, opened 2015-04-06
SWC-2203

kimyen:


src/main/resources/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.ui.xml
```diff
@@ -1,22 +1,29 @@
...
-				    	<g:Image ui:field="logoLarge" resource="{sageImages.synaspseHeader}" />					
+				    	<g:Image ui:field="logoLarge" resource="{sageImages.synaspseHeader}" />
 						<g:Image ui:field="logoSmall" resource="{sageImages.synapseLogoSmall}" />
 					</a>
 			    </div>
 				<b:Tooltip placement="BOTTOM" title="Log out">
 					<b:Button size="EXTRA_SMALL" ui:field="logoutLink" icon="SIGN_OUT" iconSize="LARGE" addStyleNames="right margin-5" height="26px" />	
 				</b:Tooltip>
+				<b:Tooltip placement="BOTTOM" title="Favorite">
+					<b:ButtonGroup ui:field="headerFavButtonGroup" addStyleNames="right margin-5">
```
jay-hodgson: instead of "right" add pull="RIGHT"

src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
```diff
@@ -95,4 +93,20 @@ public void onLoginClick() {
...
+		if (headers == null || headers.size() == 0) {
+			view.setEmptyFavorite();
+		} else {
+			view.addFavorite(headers);
+		}
+	}
+
+	@Override
+	public void applyFavClick(EntityHeader header) {
+		globalApplicationState.getPlaceChanger().goTo(new Synapse(header.getId()));
```
jay-hodgson: try setting the anchor list item href to DisplayUtils.getSynapseHistoryToken() for each entity, rather than js handling.

PR #1564, opened 2015-04-07
refresh favorites

kimyen:
In Progress


src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
```diff
@@ -50,32 +57,45 @@ public Widget asWidget() {
...
+				@Override
+				public void invoke() {
+					// only get called on failure
+					view.clearFavorite();
+				}
+			});
+		}
 		view.refresh();
 		view.setSearchVisible(true);
 	}
```
kimyen, line 82: Please review refresh()
case 1, both cachedUserSessionData & userSessionData are null => view.setUser()
case 2, cachedUserSessionData is null and userSessionData is not => view.setUser() then initUserFavorites()
case 3, cachedUserSessionData is not null and userSessionData is null => view.setUser() then do not initUserFavorites()
case 4, both are not null & different => view.setUser() then initUserFavorites()
case 5, both are not null & same => initUserFavorites()

src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
```diff
@@ -103,4 +125,19 @@ public void onFavoriteClick() {
...
+			@Override
+			public void onSuccess(List<EntityHeader> favorites) {
+				globalApplicationState.setFavorites(favorites);
+			}
+			@Override
+			public void onFailure(Throwable caught) {
+				callback.invoke();
+			}
+		});
+	}
```
kimyen, line 141: should initUserFavorites() be private?

src/main/java/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.java
```diff
@@ -241,21 +240,12 @@ public void setMenuItemActive(MenuItems menuItem) {
 	public void removeMenuItemActive(MenuItems menuItem) {
 	}
 
-	private void refreshTestSiteHeader() {
-		testSitePanel.setVisible(DisplayUtils.isInTestWebsite(cookies));
-	}
-
```
kimyen, line 247: moved this down together with other private methods

src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
```diff
@@ -50,32 +57,45 @@ public Widget asWidget() {
...
-	
+
 	public void setSearchVisible(boolean searchVisible) {
 		view.setSearchVisible(searchVisible);
 	}
-	
+
 	public void refresh() {
+		UserSessionData userSessionData = authenticationController.getCurrentUserSessionData();
+		if (cachedUserSessionData == null || cachedUserSessionData.equals(userSessionData)) {
```
jay-hodgson: Should only get into if null or user change.  Should contain init favorites call (not get favorites on every refresh).

src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
jay-hodgson, line 141: Leave public for test cases

src/test/java/org/sagebionetworks/web/unitclient/widget/header/HeaderTest.java
```diff
@@ -134,4 +139,34 @@ public void testOnFavoriteClickNonEmptyCase() {
...
+		// After User Logged in
+		UserSessionData userSessionData = new UserSessionData();
+		when(mockAuthenticationController.getCurrentUserSessionData()).thenReturn(userSessionData);
+		header.refresh();
+		verify(mockSynapseClient).getFavorites(any(AsyncCallback.class));
+
+		// User refresh
+		when(mockAuthenticationController.getCurrentUserSessionData()).thenReturn(userSessionData);
+		header.refresh();
+		verify(mockSynapseClient, Mockito.times(2)).getFavorites(any(AsyncCallback.class));
```
jay-hodgson: 1 time

PR #1566, opened 2015-04-07
Changed the Welcome message to one about getting certified and changed t...

Njanderson:
...he cookie signature so that those who dismissed the welcome message will still get the message to get certified.


src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -202,6 +202,9 @@ public void onSuccess(String passingRecordJson) {
 					PassingRecord passingRecord = new PassingRecord(adapterFactory.createNew(passingRecordJson));
 					view.updateView(profile, isOwner, passingRecord);
 					tabClicked(area);
+					if (passingRecord == null) {
```
jay-hodgson: remove

PR #1571, opened 2015-04-09
SWC-2113

kimyen:
Task: 
- In Wiki tab, when user clicks on a node in the navigation tree, reloading the wiki content instead of reloading the whole page.

Approach: 
- Passing a CallbackP from WikiPageWidget to WikiSubpageNavigationTree (through WikiSubpageWidget), invoking the callback when a node is clicked.

Summary:
- add GlobalApplicationState.pushCurrentPlace() to change the url and save the change
- add WikiPageWidget.reloadWikiPage() -- called by CallbackP.invoke() above.
- in WikiPageWidgetViewImpl, extract the code that reload the WikiPagePanel to resetWikiMarkdown() - called by presenter.reloadWikiPage() above.


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -444,5 +444,13 @@ public void replaceHistoryState(String token) {
...
+
+	@Override
+	public void pushHistoryState(String token) {
+		_pushHistoryState(token);
+	}
+
+	private static native void _pushHistoryState(String token)/*-{
+		$wnd.history.pushState( {} , '', '#'+token );
+	}-*/;
 }
```
kimyen: In console, there is an error in this file (line 374), the 2nd line of this method:
public final static native void _consoleError(String message) /*-{
		console.error(message);
	}-*/;
Maybe this is why the pushState() did not work.

src/main/java/org/sagebionetworks/web/client/widget/entity/WikiPageWidgetViewImpl.java
```diff
@@ -145,7 +222,7 @@ private void showDefaultViewWithWiki() {
 			FlowPanel noticePanel = createDifferentVersionNotice();
 			add(wrapWidget(noticePanel, "alert alert-"+BootstrapAlertType.WARNING.toString().toLowerCase()+" wikiVersionNotice"));
 		}
```
kimyen: couldn't hide / remove noticePanel. I have tried a to set noticePanel and noticeWidget as global variables but it causes some bugs. Maybe I'll try again later. Do you have any suggestion on how to handle this?

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/WikiSubpagesWidget.java
```diff
@@ -52,10 +54,11 @@ public WikiSubpagesWidget(WikiSubpagesView view, SynapseClientAsync synapseClien
 	}	
 	@Override
 	public void configure(final WikiPageKey wikiKey, Map<String, String> widgetDescriptor, Callback widgetRefreshRequired, Long wikiVersionInView) {
```
jay-hodgson: remove this configure (and do not implement WidgetRendererPresenter), because this is not a wiki widget

PR #1578, opened 2015-04-14
SWC-2279: FF logic removed

jay-hodgson:


jay-hodgson:
merging up from release-86 to release-85

PR #1599, opened 2015-04-17
Added .userProfileId to the end of the visibility cookie, so the cookie ...

Njanderson:
...will not be held between users using the same cookies.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
@@ -246,7 +246,7 @@ public class ProfilePresenter extends AbstractActivity implements ProfileView.Pr
 		if (isOwner) {
 			boolean isCertificationMessageVisible = false;
 			try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -780,7 +780,7 @@ public void setGetCertifiedDismissed() {
 		//set certification message visible=false for a year
 		Date yearFromNow = new Date();
 		CalendarUtil.addMonthsToDate(yearFromNow, 12);
-		cookies.setCookie(USER_PROFILE_CERTIFICATION_VISIBLE_STATE_KEY, Boolean.toString(false), yearFromNow);
+		cookies.setCookie(USER_PROFILE_CERTIFICATION_VISIBLE_STATE_KEY + "." + currentUserId, Boolean.toString(false), yearFromNow);
```
jay-hodgson, line 783: update ProfilePresenterTest

Njanderson, line 783: I repaired the tests.

PR #1607, opened 2015-04-21
SWC-2294 and SWC-2285

jay-hodgson:


src/test/java/org/sagebionetworks/web/unitclient/widget/entity/WikiPageWidgetTest.java
```diff
@@ -102,7 +102,7 @@ public void testConfigure() throws JSONObjectAdapterException{
 	public void testConfigureNoWikiPageCannotEditIsEmbedded(){
 		AsyncMockStubber.callFailureWith(new NotFoundException()).when(mockSynapseClient).getV2WikiPageAsV1(any(WikiPageKey.class), any(AsyncCallback.class));
 		presenter.configure(new WikiPageKey("ownerId", ObjectType.ENTITY.toString(), null, null), false, null, true);
-		verify(mockView).showWarningMessageInPage(anyString());
+		verify(mockView).clear();
 	}
```
xschildw, line 108: Should we have a test with no Wiki page but can edit?

jay-hodgson, line 108: Good catch, I thought that test was already there!  Added.

PR #1624, opened 2015-04-27
Swc 2307

Njanderson:


src/main/java/org/sagebionetworks/web/client/presenter/QuizPresenter.java
```diff
@@ -73,16 +96,60 @@ public void goToLastPlace() {
...
+		}
+		
+		//initialize if necessary
+		if (!isSubmitInitialized) {
+			isSubmitInitialized = true;
+			view.addSubmitHandler(new ClickHandler() {
+				@Override
+				public void onClick(ClickEvent event) {
+					//gather answers and pass them back to the presenter
+					if (questionIndexToQuestionWidget.keySet().size() < currentQuestionCount) {
```
jay-hodgson: need to go through each question widget to ask them if they have been answered.

src/main/webapp/Portal.css
```diff
@@ -1804,6 +1804,11 @@ a.projectTitle:hover {
 .padding-left-10 {
 	padding-left: 10px;
 }
+
```
jay-hodgson, line 1807: remove

PR #1636, opened 2015-05-06
Swc 2172 revised

Njanderson:
Concerned about having tested enough as this is a more nuanced feature.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
@@ -35,7 +35,6 @@ import org.sagebionetworks.web.client.widget.entity.ChallengeBadge;
 import org.sagebionetworks.web.client.widget.entity.ProjectBadge;
 import org.sagebionetworks.web.client.widget.entity.browse.EntityBrowserUtils;
 import org.sagebionetworks.web.client.widget.profile.UserProfileModalWidget;
-import org.sagebionetworks.web.client.widget.team.TeamListWidget;
 import org.sagebionetworks.web.shared.ChallengeBundle;
 import org.sagebionetworks.web.shared.ChallengePagedResults;
 import org.sagebionetworks.web.shared.LinkedInInfo;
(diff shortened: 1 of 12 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -390,20 +387,34 @@ public void refreshTeams() {
...
+				@Override
+				public void onFailure(Throwable caught) {
+					view.setTeamsError(caught.getMessage());
+				}
+				@Override
+				public void onSuccess(List<TeamRequestBundle> teams) {
+					view.setTeamsFromBundle(teams);
+				}
+			};
+			TeamListWidget.getTeamBundles(currentUserId, synapseClient, adapterFactory, teamCallback);
```
jay-hodgson: Please make getTeamBundles non-static, and move it in this class.  You can add a boolean to the servlet call to includeRequestCounts, and only use this call (remove reference to TeamListWidget.getTeams()).

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -510,13 +507,20 @@ public void onClick(ClickEvent event) {
...
+	public void setTeams(List<Team> teams) {
+		myTeamsWidget.configure(teams, false);
+	}
+	
+	@Override
+	public void setTeamsFromBundle(List<TeamRequestBundle> teamsRequestBundles) {
+		int requestCount = 0;
+		List<Team> teams = new ArrayList<Team>(teamsRequestBundles.size());
+		for (TeamRequestBundle teamAndRequest: teamsRequestBundles) {
+			requestCount += teamAndRequest.getRequestCount();
```
jay-hodgson: request count should be calculated in the presenter (and add to the total team notification count) instead of adding it up in the view and communicating it back to the presenter.

src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -390,20 +387,34 @@ public void refreshTeams() {
...
+				public void onFailure(Throwable caught) {
+					view.setTeamsError(caught.getMessage());
+				}
+				@Override
+				public void onSuccess(List<Team> teams) {
+					view.setTeams(teams);
+				}
+			};
+			
+			TeamListWidget.getTeams(currentUserId, synapseClient, adapterFactory, teamCallback);
```
jay-hodgson: TeamListWidget.getTeams is only called from ProfilePresenter.  Please remove completely.  Also remove SynapseClient.getTeamsForUser()

Njanderson:
Updates when invited to a team, or when a request is made to join another team.

PR #1642, opened 2015-05-06
do not cache redirected pre-signed url since it expires

jay-hodgson:


jay-hodgson:
If the redirect expires before the image, then we have no problem (it will ask the servlet which will respond with a fresh redirect).
If the image expires before the redirect, then the browser may attempt to use the presigned url, which of course is expired, and image will fail to load.
As it stands, Xa reports that he hasn't seen this issue for a couple of weeks, so I'm closing this PR (since performance is far better if the browser is successfully caching profile images).

PR #1674, opened 2015-05-15
Swc-2364: get entity bundle with permissions to determine if Edit Order should be shown.

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/WikiSubpagesWidget.java
```diff
@@ -67,16 +72,15 @@ public void configure(final WikiPageKey wikiKey, Map<String, String> widgetDescr
 			List<Reference> allRefs = new ArrayList<Reference>();
```
xschildw: Remove, not used

PR #1681, opened 2015-05-21
SWC 2248 + SWC 2325 + SWC 2020 + SWC 2378

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileTitleBarViewImpl.java
```diff
@@ -151,10 +152,16 @@ public void createTitlebar(
 			}
 			else if (fileHandle instanceof S3FileHandleInterface){
 				fileName.setInnerText(fileHandle.getFileName());
-				
-				S3FileHandleInterface s3FileHandle = (S3FileHandleInterface)fileHandle;
-				
-				fileSize.setInnerText("("+DisplayUtils.getFriendlySize(s3FileHandle.getContentSize().doubleValue(), true) + " - Synapse Storage)");
+
+				final S3FileHandleInterface s3FileHandle = (S3FileHandleInterface)fileHandle;
+				presenter.setS3Description(new CallbackP<String>() {
```
jay-hodgson, line 157: I'd prefer this callback not be here.  What do you think about modifying the ui xml so that there is a span for the file size, and a new span for the location description?  Then the presenter would need to call view.setLocationDescription on the async response, and the UI may look like "File:  AFILE.rpkm| 3.1415 MB | Synapse Storage | md5"

jay-hodgson, line 157: Opened SWC-2384 to track request.  Merging this important fix.

PR #1691, opened 2015-05-26
Small bug, now fixed. All tests pass.

Njanderson:


src/main/java/org/sagebionetworks/web/client/widget/entity/editor/ImageConfigEditor.java
```diff
@@ -41,6 +42,7 @@ public ImageConfigEditor(ImageConfigView view, FileHandleUploadWidget fileInputW
 		this.wikiAttachments = wikiAttachments;
 		view.setFileInputWidget(fileInputWidget.asWidget());
 		view.setWikiAttachmentsWidget(wikiAttachments.asWidget());
+		fileHandleIds = new ArrayList<String>();
```
jay-hodgson: please move initialization of file handle ids list to configure().

PR #1731, opened 2015-06-22
Fix release 96

xschildw:


xschildw:
Fixing pull error from yesterday.

PR #1748, opened 2015-06-30
Swc-1954: Expose way to change project/folder storage location

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/StorageLocationWidget.java
```diff
@@ -0,0 +1,193 @@
...
+
+	@Override
+	public void onSave() {
+		synAlert.clear();
+		StorageLocationSetting setting = getStorageLocationSettingFromView();
+		String error = validate(setting);
+		if (error != null) {
+			synAlert.showError(error);
+		} else {
+			//look for duplicate storage location in existing settings
```
marcel-blonk, line 121: This comment and the code don't seem to agree?

src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -3095,4 +3063,89 @@ public EntityBundlePlus getEntityInfo(String entityId) throws RestServiceExcepti
 		}
 	}
 	
+	@Override
+	public List<String> getMyLocationSettingBanners() throws RestServiceException{
+		try {
+			Set<String> banners = new HashSet<String>();
```
marcel-blonk: Use treeset and get sorting automatically? (have to add case insensitve comparator)

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/StorageLocationWidget.java
jay-hodgson, line 121: Good catch. I ended up doing this servlet side instead, see SynapseClientImpl.createStorageLocationSetting()

jay-hodgson:
blessed by Marcel

PR #1749, opened 2015-06-30
SWC-2470: add new Error place that takes a single LogEntry and displays it.

jay-hodgson:
Changed the FileHandleServlet to redirect to the new Error place if there are any problems with the download.


xschildw:
Please put in release-97.

PR #1759, opened 2015-07-06
SWC-2460: add synapse email to profile.  also convert View Profile pa…

jay-hodgson:
…nel to ui xml instead of building it up each time.


brucehoff:
Great job!!

PR #1781, opened 2015-07-17
Added null check that was generating an error when the activity.getUs…

Njanderson:
…ed was null

I only added a check to make sure the getUsed isn't null before I ask for an iterator for the set it may generate.


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/ProvenanceEditorWidget.java
```diff
@@ -80,41 +80,43 @@ public void onSuccess(Activity result) {
...
-					ProvenanceEntry toAdd;
-					if (provEntry instanceof UsedEntity) {
-						Reference ref = ((UsedEntity)provEntry).getReference();
-						toAdd = ginInjector.getEntityRefEntry();
-						String entityId = ref.getTargetId();
-						Long version = ref.getTargetVersionNumber();
-						String versionString = null;
-						if (version != null) {
-							versionString = version.toString();
+				if (allProvenance != null) {
```
jay-hodgson, line 83: Please add a test to ProvenanceEditorWidgetTest that returns null for the returned Activity getUsed()

PR #1783, opened 2015-07-17
Created ValidationUtils refactor, added Validation to the Prov URL Di…

Njanderson:
…alog and removed some imports/excess methods


src/main/java/org/sagebionetworks/web/client/widget/entity/editor/VimeoConfigEditor.java
```diff
@@ -21,7 +21,6 @@
 	@Inject
 	public VimeoConfigEditor(IFrameConfigView view) {
 		this.view = view;
-		view.setPresenter(this);
```
jay-hodgson, line 24: Why remove this setPresenter?  It seems dangerous for the injected view not to have a valid reference to this presenter.

PR #1789, opened 2015-07-19
Added remove text

Njanderson:


jay-hodgson:
Does marking as a favorite, or unfavorite, update the tooltip text?  I ask because setTitle does not reconfigure the underlying tooltip js object...

PR #1798, opened 2015-07-22
home-v2: remove bootstrap text area customization (and instead customize our paper theme)

jay-hodgson:


jay-hodgson:
reviewed by Xa

PR #1813, opened 2015-07-28
move css from SynapseColor.css to Portal.css to remove another blocki…

jay-hodgson:
…ng request


john-hill:
I don't see any new !important so it must be good!

PR #1833, opened 2015-08-12
Swc 2537

Njanderson:


Njanderson:
Need to add a test that demonstrates the new functionality and exposes the old weakness

PR #1865, opened 2015-08-26
Migrated changes from an old branch I never finished; should do some …

Njanderson:
…more testing


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -147,6 +147,10 @@ public void configure(ActionMenuWidget actionMenu,
 		actionMenu.addControllerWidget(uploader.asWidget());
 		if (!isUserAuthenticated) {
 			actionMenu.setToolsButtonVisible(false);
+			if (permissions.getCanPublicRead()) {
```
jay-hodgson, line 151: add a test to verify annotations and file history are configured when not logged in, and can public read (and not configured otherwise).

Njanderson, line 151: The smoke testing seemed good when I checked against a public project then a public file.

PR #1870, opened 2015-08-28
Added overflow hidden to make sure it didn't inherit overflow

Njanderson:


jay-hodgson:
I think the proposed fix is not the right solution, and would change the bug title to "Tables icon is cut off on FF/Chrome"

Njanderson:
Strangely enough it still worked (in both Chrome and FF) , but we'll go with auto because it seems the best solution for this. Agreed to make this change.

PR #1873, opened 2015-09-01
Swc 2593

Njanderson:


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -329,7 +329,7 @@ public void handleWikiReload(String wikiPageId) {
 			setArea(EntityArea.WIKI, wikiPageId);
 			view.configureProjectActionMenu(bundle, wikiPageId);
 		} else {
-			DisplayUtils.showErrorMessage("Failed to handle Wiki reload.");
+			view.configureFileActionMenu(bundle, wikiPageId);
```
jay-hodgson, line 332: I suppose since a Folder also uses the fileActionMenuContainer, this will fix for both files and folders.

jay-hodgson:
This needs to be fixed on staging (release-103) (currently this PR is aimed at develop)

Njanderson:
Gotcha. Walking to new apartment then will do that. I actually haven't even
seen it, so I'm prepared to be either pleasantly surprised or ready to move
out as move in.
Also I successfully made a new project with dependencies not generate any
errors and my plfm built with only a timeout. Small victories.

jay-hodgson:
Will merge down from release-103 instead (to get the same commits and avoid complex merge).

PR #1881, opened 2015-09-07
Added jsPlumb to the caching filter correctly

Njanderson:


jay-hodgson:
Library does not appear to have any external dependencies in the source tree, so that should be it!

PR #1884, opened 2015-09-08
Added a few setters to make sure button text is set even if not logge…

xschildw:
…d in or just invited


xschildw:
Merged 103 change into 104

PR #1892, opened 2015-09-13
Waiting for back-end service to be able to null out a profilePictureF…

Njanderson:
…ileHandleID


src/main/java/org/sagebionetworks/web/client/widget/profile/ProfileImageViewImpl.java
```diff
@@ -16,12 +21,26 @@
...
+			@Override
+			public void onClick(ClickEvent event) {
+				removePictureCallback.invoke();
+			}
+		});
+	}
+	
+	@Override
+	public void setRemovePictureCallback(Callback callback) {
+		this.removePictureCallback = callback;
```
jay-hodgson: View should not hold on to callback.  View should simply call back to presenter (presenter.onRemotePicture()).  Presenter can invoke callback if not null.  
Add test to presenter, to verify callback is invoked.

PR #1896, opened 2015-09-15
Swc-2530: ability to set the override file name

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/EditFileMetadataModalWidgetImpl.java
```diff
@@ -63,18 +66,15 @@ public void onFailure(Throwable caught) {
 	public void onPrimary() {
 		String name = StringUtils.trimWithEmptyAsNull(view.getEntityName());
 		String fileName = StringUtils.trimWithEmptyAsNull(view.getFileName());
-		String contentType = StringUtils.trimWithEmptyAsNull(view.getContentType());
 		if(name == null){
 			view.showError(RenameEntityModalWidgetImpl.NAME_MUST_INCLUDE_AT_LEAST_ONE_CHARACTER);
 		} else if(fileName == null){
```
jay-hodgson, line 71: support clearing override file name?

jay-hodgson:
The FileTitleBarViewImpl should use the bundle file name rather than show the file name of the file handle (since it may have been overwritten).

PR #1910, opened 2015-09-28
Need to run a Super Dev mode session for further testing, but current…

Njanderson:
…ly compiles and passes all tests, including new tests.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
@@ -333,4 +333,39 @@ public class SettingsPresenter implements SettingsView.Presenter {
 			}
 		});
 	}
+
+	@Override
+	public void changePassword() {
+		clearPasswordErrors();
+		String currentPassword = view.getCurrentPasswordField();
+		String password1 = view.getPassword1Field();
+		String password2 = view.getPassword2Field();
+		if (!checkPasswordDefined(currentPassword)) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
```diff
@@ -333,4 +333,39 @@ public void invoke() {
...
+		String currentPassword = view.getCurrentPasswordField();
+		String password1 = view.getPassword1Field();
+		String password2 = view.getPassword2Field();
+		if (!checkPasswordDefined(currentPassword)) {
+			view.setCurrentPasswordErrorMessage(DisplayConstants.ERROR_ALL_FIELDS_REQUIRED);
+			view.setCurrentPasswordInError(true);
+		} else if (!checkPasswordDefined(password1)){
+			view.setPassword1ErrorMessage(DisplayConstants.ERROR_ALL_FIELDS_REQUIRED);
+			view.setPassword1InError(true);
+		} else if (!checkPasswordDefined(password2) || !password1.equals(password2)) {
```
jay-hodgson: This is some great work.  But please show a different error if passwords mismatch (DisplayConstants.PASSWORDS_MISMATCH), not ERROR_ALL_FIELDS_REQUIRED in this case.

You could simply this by using a single SynapseAlert to communicate errors, which would end up removing 6 methods from the view interface.

Njanderson: This is indeed the best solution, and I was really wanting some design overview regarding this issue, as the previous solution was pretty brute force.

PR #1914, opened 2015-09-30
PLFM-3584

brucehoff:


src/main/java/org/sagebionetworks/web/client/SynapseClient.java
```diff
@@ -380,7 +380,7 @@ AccessRequirementsTransport getUnmetAccessRequirements(String entityId, ACCESS_T
 	
 	public ResponseMessage handleSignedToken(SignedTokenInterface signedToken, String hostPageBaseURL) throws RestServiceException;
 	
-	public SignedTokenInterface hexDecodeAndSerialize(String tokenTypeName, String signedTokenString) throws RestServiceException;
+	public SignedTokenInterface hexDecodeAndDeserialize(String tokenTypeName, String signedTokenString) throws RestServiceException;
```
jay-hodgson, line 383: Also need to make the name change in SynapseClientAsync (and all references, SignedTokenPresenter and SignedTokenPresenterTest).

PR #1932, opened 2015-10-09
Swc 2668

Njanderson:


src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/FilesTab.java
```diff
@@ -274,6 +290,32 @@ public void onFailure(Throwable caught) {
 		}
 	}
 	
+	 public void configureStorageLocation(String containerEntityId) {
+		 synapseClient.getUploadDestinations(containerEntityId, new AsyncCallback<List<UploadDestination>>() {
```
jay-hodgson: this looks a lot like the code up in the EntityPageTop.  should this instead be a method in EntityMetadata (where you can request it to get and show the storage location).

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadata.java
```diff
@@ -59,6 +71,9 @@ public void setEntityBundle(EntityBundle bundle, Long versionNumber) {
 			view.setRestrictionPanelVisible(en instanceof TableEntity
 					|| en instanceof Folder);
 		}
+		if (en instanceof Folder || en instanceof Project) {
```
jay-hodgson: Please move the type check to configureStorageLocation, and pass the entity in.  Then test to verify that it only asks for the upload destinations when entity is a Folder or Project.

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadata.java
```diff
@@ -94,4 +109,30 @@ public void clear() {
 		view.clear();
 	}
 	
+	 public void configureStorageLocation(String containerEntityId) {
+		 synapseClient.getUploadDestinations(containerEntityId, new AsyncCallback<List<UploadDestination>>() {
```
jay-hodgson: Could be more explicit about the view state here.  That is, instead of relying on view.clear to hide the upload destination panel, this method could initially set visibility to false, and then show if it successfully determines the destination.

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadata.java
```diff
@@ -94,4 +109,30 @@ public void clear() {
...
+						description += externalUploadDestination.getBaseKey();
+					};
+					view.setUploadDestinationText(description);
+				}
+				view.setUploadDestinationPanelVisible(true);
+			}
+
+			@Override
+			public void onFailure(Throwable arg0) {
+				view.setUploadDestinationPanelVisible(false);
```
jay-hodgson: Lost the error.  At the very least, log the exception to the js console.

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/EntityPageTopTest.java
```diff
@@ -158,6 +162,7 @@ public void testConstruction(){
 	
 	@Test
 	public void testConfigureWithProject(){
+		AsyncMockStubber.callSuccessWith(null).when(mockSynapseClientAsync).getUploadDestinations(anyString(), any(AsyncCallback.class));
```
jay-hodgson: remove

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/EntityPageTopTest.java
```diff
@@ -180,6 +185,7 @@ public void testConfigureWithProject(){
 	
 	@Test
 	public void testConfigureWithProjectWikiToken(){
+		AsyncMockStubber.callSuccessWith(null).when(mockSynapseClientAsync).getUploadDestinations(anyString(), any(AsyncCallback.class));
```
jay-hodgson: remove

PR #1935, opened 2015-10-11
Pre-appending @ will redirect to PeopleSearch in the Header Search

Njanderson:


src/main/java/org/sagebionetworks/web/client/widget/search/HomeSearchBox.java
```diff
@@ -50,7 +51,9 @@ public void clearState() {
     
 	@Override
 	public void search(String value) {
-		if(searchAll) {
+		if (value.charAt(0) == '@') {
```
jay-hodgson: Looks like a straight forward (if not secret) solution for advanced Synapse users, but I don't see the prerequisite precondition checks (search term is not null and not empty), nor the corresponding tests.

src/main/java/org/sagebionetworks/web/client/widget/search/HomeSearchBox.java
```diff
@@ -11,6 +11,7 @@
 import org.sagebionetworks.web.client.DisplayUtils;
```
jay-hodgson: Please remove HomeSearchBox (and new test).

PR #1940, opened 2015-10-14
Bumped version of markdown processor

Njanderson:


pom.xml
```diff
@@ -26,6 +26,7 @@
 		<org.json.version>20090211</org.json.version>
 		<jackson.version>1.8.1</jackson.version>
 		<log4j.version>2.0-beta8</log4j.version>
+		<markdown.version>2015-10-14-2cea4d8-1295</markdown.version>
```
jay-hodgson, line 29: Version should be 2015-10-14-d491778-18 (based on hudson output).  Please update to fix build.

PR #1946, opened 2015-10-23
SWC-2736: make it obvious when portal is not pointing to the production environment.  also merge release-110 down to develop.

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
```diff
@@ -22,6 +24,8 @@
 
 public class Header implements HeaderView.Presenter, IsWidget {
 
+	public static final String WWW_SYNAPSE_ORG = "www.synapse.org";
```
xschildw, line 27: Check that we have it when redirected from https://synapse.org

PR #1949, opened 2015-10-29
SWC-2731:  add integration with ORC ID (in alpha mode only until ORCID exposed in User Profile bundle)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/ErrorPresenter.java
```diff
@@ -55,6 +55,9 @@ public void onSuccess(LogEntry result) {
 				jsniUtils.consoleError(result.getLabel());
 				jsniUtils.consoleError(result.getMessage());
 				jsniUtils.consoleError(result.getStacktrace());
+				if (!synAlert.isUserLoggedIn()) {
+					synAlert.showMustLogin();
```
jay-hodgson, line 59: Don't indicate that logging in will gain access to the resource.  Instead, suggest that logging in may help resolve this error (no promises).  "If you feel this error is due to anonymous browsing, please log in".

src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -688,6 +693,26 @@ private void showView(Profile place) {
 		projectSynAlert.clear();
 		teamSynAlert.clear();
 		String token = place.toToken();
+		if (token.startsWith("message/")) {
```
jay-hodgson, line 696: Instead of a generic message, use a set of predefined messages instead.  Look for something like ORC_ID_SUCCESS.

src/main/java/org/sagebionetworks/web/server/servlet/oauth2/OAuth2AliasServlet.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.sagebionetworks.client.SynapseClient;
+import org.sagebionetworks.repo.model.LogEntry;
+import org.sagebionetworks.repo.model.oauth.OAuthProvider;
+import org.sagebionetworks.repo.model.oauth.OAuthValidationRequest;
+import org.sagebionetworks.repo.model.principal.PrincipalAlias;
+import org.sagebionetworks.util.SerializationUtils;
+import org.sagebionetworks.web.shared.WebConstants;
+
+public class OAuth2AliasServlet extends OAuth2SessionServlet {
```
jay-hodgson, line 18: Change so that both OAuth2AliasServlet and OAuth2SessionServlet share the same parent (or are composed with the same object)

PR #1952, opened 2015-10-30
change of UI when not looking at production website

jay-hodgson:


jay-hodgson:
screen shot 2015-10-30 at 4 46 02 pm

PR #1956, opened 2015-11-04
Swc-2749: implement a reusable file list widget (necessary for user verification feature)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/upload/FileHandleLink.java
```diff
@@ -0,0 +1,93 @@
...
+	}
+	
+	public FileHandleLink configure(String fileHandleId, String fileName, CallbackP<String> fileClickedCallback) {
+		this.fileClickedCallback = fileClickedCallback;
+		this.fileHandleId = fileHandleId;
+		fileHandleLink.setText(fileName);
+		return this;
+	}
+	
+	public FileHandleLink setFileClickedCallback(CallbackP<String> fileClickedCallback) {
```
brucehoff, line 64: setFileClickedCallback(CallbackP<String> fileClickedCallback)  is not needed

PR #1957, opened 2015-11-05
SWC-2754:  wrap new Synapse java client calls in a new servlet, and use the new UserBundle in the profile page

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -230,45 +239,34 @@ public void invoke() {
...
-			public void onSuccess(String passingRecordJson) {
-				try {
-					view.hideLoading();
-					PassingRecord passingRecord = new PassingRecord(adapterFactory.createNew(passingRecordJson));
-					if (passingRecord != null)
-						view.addCertifiedBadge();
-					view.setProfile(profile, isOwner);
-				} catch (JSONObjectAdapterException e) {
-					onFailure(e);
+				initializeShowHideProfile(isOwner);
```
brucehoff, line 250: iniitialize BEFORE the call to the async thingy

brucehoff:
Great job!  And thanks for keeping the code reviews small!!!

PR #1967, opened 2015-11-17
SWC-2763: Sage specific ACT place

jay-hodgson:
Note that the backend will respond with errors if someone should attempt to use the service calls referenced by this page (if not in the ACT).


src/main/java/org/sagebionetworks/web/client/presenter/BulkPresenterProxy.java
```diff
@@ -210,6 +211,11 @@ public void onSuccess() {
 					ErrorPresenter presenter = ginjector.getErrorPresenter();
 					presenter.setPlace((ErrorPlace) place);
 					presenter.start(panel, eventBus);
+				} else if (place instanceof ACTPlace) {
+					// user's profile page
```
jay-hodgson, line 215: remove comment

PR #1988, opened 2015-11-25
SWC-737:  Add Cytoscape JS wiki widget

jay-hodgson:


kimyen:
include Cytoscape version in the widget param

PR #1989, opened 2015-11-25
SWC-2784: add More Info button that opens up a new browser window whe…

jay-hodgson:
…re we can describe what Bridge data is and why one would want access to it


src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -264,6 +265,8 @@
 	Button submitProfileValidationButton;
 	@UiField
 	Button verificationApprovedButton;
+	@UiField
+	Button bridgeDataAccessButton;
```
brucehoff: whyGetValidatedButton

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -343,7 +346,12 @@ public void onClick(ClickEvent event) {
 				presenter.goTo(new Search(""));
 			}
 		});
-		
+		bridgeDataAccessButton.addClickHandler(new ClickHandler() {
+			@Override
+			public void onClick(ClickEvent event) {
+				DisplayUtils.newWindow("#!Help:BridgeDataAccess", "_blank", "");
```
brucehoff: Help:WhyGetValidated

brucehoff:
Awesome job!!

PR #2015, opened 2015-12-14
Swc 2787

Njanderson:


src/main/java/org/sagebionetworks/web/client/SynapseClient.java
```diff
@@ -545,4 +546,6 @@ Activity getOrCreateActivityForEntityVersion(String entityId,
 
 	Boolean isTeamMember(String userId, Long groupPrincipalId)
 			throws RestServiceException;
+
```
jay-hodgson: Already available in another servlet, please remove.

Njanderson: Gotcha!

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -82,6 +90,31 @@ public void onSuccess(UserSessionData userSessionData) {
 					cookies.setCookie(CookieKeys.USER_LOGIN_TOKEN, userSessionData.getSession().getSessionToken(), tomorrow);
 					currentUser = userSessionData;
 					callback.onSuccess(userSessionData);
+					
+					// Attempt to get UserBundle information for user being set
```
jay-hodgson: Clear user bundle before any remote async call.  And only callback after finishing init of userbundle (callback.onSuccess(userSessionData);).  Change login call to return user session data and user bundle (unless usersessiondata is a portal transfer object, then just add a UserBundle field to it).

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -145,6 +157,16 @@ public String getCurrentUserSessionToken() {
 	}
 	
 	@Override
+	public UserBundle getCurrentUserBundle() {
```
john-hill, line 160: This method seems to only be called by tests.  If that is the case it needs to be removed.

src/main/java/org/sagebionetworks/web/client/UserAccountService.java
```diff
@@ -35,5 +36,8 @@
 	
 	public PublicPrincipalIds getPublicAndAuthenticatedGroupPrincipalIds();
 	
-	public StorageUsageSummaryList getStorageUsage();
+	public StorageUsageSummaryList getStorageUsage();
+
+	public UserLoginBundle getUserLoginBundle(String sessionToken) throws RestServiceException;
+
```
john-hill, line 42: Is the UserLoginBundle a new class?

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
Njanderson, line 160: This method will be the foundation of upcoming work on ACTs: 

The user bundle can tell us if the current user is part of the ACT. 
One step that I think we need to make this work is that on Synapse log in, we should kick off a process to ask for the user bundle of the current user and store it in a global place (authenticationController?). When asking for the current user bundle, it should verify that it is set and matches the current user id from the session data.
**Once that's done, the main feature can be implemented.** I'll draw my thoughts on a whiteboard and attach them shortly, but I think a new command would be available from the Tools menu (that would only be shown if the current user is in the ACT). When clicked, a dialog would pop up where the ACT member would select the Access Requirement and User.

src/main/java/org/sagebionetworks/web/client/UserAccountService.java
Njanderson, line 42: Yes, I have since added the class in question. Thanks!

PR #2018, opened 2015-12-30
Removed the addition of a dummy item to every tree item

Njanderson:


jay-hodgson:
I think removing this causes problems with alignment in the files browser.  Could you attach a screenshot where nested folders are expanded next to sibling folders/files?

PR #2019, opened 2015-12-31
Initial attempt at SEO

Njanderson:


src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/FilesTab.java
```diff
@@ -307,6 +307,10 @@ public void setTargetBundle(EntityBundle bundle) {
 		view.clearActionMenuContainer();
 		//Preview
 		view.setPreviewVisible(isFile);
+		
+		fileTitleBar.clearState();
+		folderTitleBar.clearState();
```
Njanderson: I would like to make sure that I understand the lifecycle enough to make sure that state is always cleared, otherwise you may have a previous file or folder's search schema. The alternative is to completely clear out both title bars, then only re-add them when you want a certain titlebar.

jay-hodgson:
The crawler gets a different version of the page, it never encounters the action entity page.  See org.sagebionetworks.web.server.servlet.filter.CrawlFilter

As a side note, the anonymous user should never get a direct link to the file either (we don't allow for anonymous download).  It should instead be a link to the file entity page.

PR #2025, opened 2016-01-07
Discussion Tab

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -131,6 +136,7 @@ public void invoke(Tab t) {
 		};
 		wikiTab.setTabClickedCallback(showProjectInfoCallback);
 		adminTab.setTabClickedCallback(showProjectInfoCallback);
+		discussionTab.setTabClickedCallback(showProjectInfoCallback);
```
jay-hodgson: Do you want the project info to be shown?  If not, use the other callback (showHideProjectInfoCallback)

PR #2028, opened 2016-01-07
SWC-2830: change behavior of editing provenance in the web client.  c…

jay-hodgson:
…reates a new Activity, and (on save) updates the activity values and associates with the entity.


jay-hodgson:
Unable to move forward with this change.  See https://sagebionetworks.jira.com/browse/SWC-2830 for discussion.  
If version is bumped, then it breaks provenance for any downstream entities (they could point to broken or missing provenance).  But Brian thinks that the system would be inconsistent if it didn't bump the version, since the change to provenance would not be reflected in the version history.

PR #2044, opened 2016-01-14
Discussion Forum

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadListWidget.java
```diff
@@ -4,57 +4,68 @@
...
 		this.ginInjector = ginInjector;
 		this.discussionForumClientAsync = discussionForumClientAsync;
+		this.synAlert = synAlert;
 		view.setPresenter(this);
+		view.setAlert(synAlert.asWidget());
 	}
 
-	@Override
 	public void configure(String forumId) {
 		view.clear();
```
jay-hodgson, line 43: move synAlert.clear() to here (before the async call).

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidget.java
```diff
@@ -34,7 +39,28 @@ public void configure(DiscussionThreadBundle bundle) {
 		view.setActiveUsers(bundle.getActiveAuthors().toString());
 		view.setNumberOfReplies(bundle.getNumberOfReplies().toString());
 		view.setNumberOfViews(bundle.getNumberOfViews().toString());
-		view.setLastActivity(bundle.getLastActivity().toString());
+		view.setLastActivity(gwtWrapper.getFormattedDateString(bundle.getLastActivity()));
+		view.setAuthor(bundle.getCreatedBy());
+		view.setCreatedOn(gwtWrapper.getFormattedDateString(bundle.getCreatedOn()));
+		if (bundle.getNumberOfReplies() > 0) {
```
jay-hodgson, line 45: I think the click handler to show replies should always be set in the view.  And if 0 replies, then simply don't show the reply count.

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidget.java
```diff
@@ -1,23 +1,29 @@
...
 import com.google.inject.Inject;
 
-public class ThreadWidget implements ThreadWidgetView.Presenter{
+public class DiscussionThreadWidget implements DiscussionThreadWidgetView.Presenter{
 
-	ThreadWidgetView view;
+	DiscussionThreadWidgetView view;
 	PortalGinInjector ginInjector;
+	GWTWrapper gwtWrapper;
+	boolean areRepliesConfigure;
```
jay-hodgson, line 15: I don't understand why areRepliesConfigure exists, and assume that this will go away as implementation progresses.

kimyen, line 15: When a user click on DiscussionTab, the first 10 thread in the project is loaded. However, it's replies are not loaded. The first time, user click on "x replies" on a thread, the first n number of replies for that thread id loaded. However, clicking on the same element "x replies" from that point on, only the toggle functionality is triggered. We will not load the first n replies again. So I added a check to make sure I only load them and add them once. To load more replies from that point, user will use "load more" button.

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidget.java
kimyen, line 45: I thought of this, but haven't changed the code to do that yet because I was testing the display of replies. I will do this after I added NewReplyModal and configure ReplyWidget correctly.

src/main/java/org/sagebionetworks/web/client/widget/discussion/modal/NewDiscussionThreadModal.java
```diff
@@ -71,25 +70,37 @@ public void onSave() {
...
-		return (messageMarkdown != null && !messageMarkdown.equals(""));
-	}
-
-	private boolean isValidTitle(String threadTitle) {
-		return (threadTitle != null && !threadTitle.equals(""));
+	public static ValidationResult validate(String threadTitle, String messageMarkdown) {
+		String errorMessages = "";
+		ValidationResult result = new ValidationResult();
+		result.setValidity(true);
+		if (threadTitle == null || threadTitle.equals("")) {
```
jay-hodgson, line 91: I like this idea, and think it could be used across the app.  
What do you think about adding a new method to ValidationResult:

```
ValidationResult requiredField(String fieldName, String fieldValue)
```

My suggestion uses <a href='https://en.wikipedia.org/wiki/Method_chaining'>method chaining</a>, so that in your case validate() would become:

```
ValidationResult result = new ValidationResult();
result.requiredField("Title", threadTitle)
      .requiredField("Message", messageMarkdown);
return result;
```

kimyen, line 91: That sounds great! I will do that. In what package should the ValidationResult class be?

jay-hodgson, line 91: Not sure, how about a new package:
org.sagebionetworks.web.client.validation

PR #2050, opened 2016-01-16
Code Review Changes

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidget.java
```diff
@@ -99,14 +101,13 @@ public void toggleThread() {
...
+		configureReplies();
 		view.toggleReplies();
 	}
 
 	public void configureReplies() {
 		synAlert.clear();
+		view.clearReplies();
 		offset = 0L;
 		if (order == null) {
 			order = DEFAULT_ORDER;
```
jay-hodgson, line 113: the getRepliesForThread is going to need to be in another method once you hook up the onLoadMoreReplies() action.

PR #2051, opened 2016-01-19
Discussion forum

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidgetViewImpl.java
```diff
@@ -147,11 +147,6 @@ public void setMessage(String message) {
 	}
 
 	@Override
-	public void setActiveUsers(String activeUsers){
```
jay-hodgson, line 150: I would expect a change to view.clear() (to remove all from the activeUsers container).

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidget.java
```diff
@@ -59,12 +60,17 @@ public Widget asWidget() {
 	public void configure(DiscussionThreadBundle bundle) {
 		view.clear();
```
jay-hodgson, line 65: Since we know that there will be a single author, just inject a UserBadge in the constructor, and add it to the view.

src/main/java/org/sagebionetworks/web/client/widget/discussion/ReplyWidget.java
```diff
@@ -1,26 +1,39 @@
...
 	ReplyWidgetView view;
+	GWTWrapper gwtWrapper;
+	PortalGinInjector ginInjector;
 
 	@Inject
 	public ReplyWidget(
-			ReplyWidgetView view
+			ReplyWidgetView view,
+			GWTWrapper gwtWrapper,
+			PortalGinInjector ginInjector
```
jay-hodgson: Same, just inject UserBadge once in constructor (and add widget to the view).  Then configure the UserBadge in configure()

src/main/resources/org/sagebionetworks/web/client/widget/discussion/ReplyWidgetViewImpl.ui.xml
```diff
@@ -2,10 +2,17 @@
...
-		</b:Column>
-	</b:Row>
+	xmlns:bh="urn:import:org.gwtbootstrap3.client.ui.html"
+	xmlns:t="urn:import:org.sagebionetworks.web.client.view.bootstrap.table">
+	<bh:Div addStyleNames="margin-top-10 light-border-bottom" width="100%" pull="RIGHT" >
+		<t:Table width="100%">
+			<t:TableData>
+				<bh:Span ui:field="author"></bh:Span>
+			</t:TableData>
+			<t:TableData width="130px" addStyleNames="visible-lg visible-md">
```
jay-hodgson: Need to autohide "created on" on small devices (current set up looks like it will hide on < 992px)?  Consider 'hidden-xs' if more appropriate.

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidgetViewImpl.java
kimyen, line 150: activeUsers is only set in configure() and there is a view.clear() in configure(). This method has been removed.

PR #2056, opened 2016-01-20
Discussion forum

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidget.java
```diff
@@ -104,17 +113,68 @@ public void onSuccess(DiscussionThreadBundle result) {
...
+		discussionForumClientAsync.getThreadUrl(threadId, new AsyncCallback<MessageURL>(){
+
+			@Override
+			public void onFailure(Throwable caught) {
+				synAlert.handleException(caught);
+			}
+
+			@Override
+			public void onSuccess(MessageURL result) {
+				requestBuilder.configure(RequestBuilder.GET, result.getMessageUrl());
```
jay-hodgson, line 140: For code clarity, I suggest refactoring this into it's own method (it takes the result.getMessageUrl() String as input).

PR #2060, opened 2016-01-21
Discussion forum

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/discussion/modal/NewDiscussionThreadModalViewImpl.java
```diff
@@ -103,4 +106,19 @@ public void setAlert(Widget w) {
...
 	}
+
+	@Override
+	public void setSendingRequestVisible(boolean visible) {
+		sendingRequest.setVisible(visible);
+	}
+
+	@Override
+	public void setSaveButtonEnabled(boolean enabled) {
+		saveButton.setEnabled(enabled);
```
jay-hodgson: Use saveButton.state().loading() (and saveButton.state().reset()), and set the dataLoadingText button parameter in the ui xml.

src/main/webapp/Portal.css
```diff
@@ -847,6 +847,10 @@ a.shortBtn {
 	margin-bottom: 0px !important;
 }
 
+.margin-bottom-5-imp {
```
jay-hodgson: don't like to see new !important styles.  if the margin-bottom-5 is not working, then let's define a more specific selector.

PR #2065, opened 2016-01-26
SWC-2857: set content type to binary file, and retry in web client until uploader is closed.

jay-hodgson:


jay-hodgson:
Waiting for the backend change in order to indicate that the file part content type will be set (and should be included in the signature)...

PR #2066, opened 2016-01-26
Discussion forum

kimyen:


src/main/java/org/sagebionetworks/web/server/servlet/DiscussionMessageServlet.java
```diff
@@ -0,0 +1,93 @@
...
+			URL resolvedUrl = null;
+			if (type.equals(WebConstants.THREAD_TYPE)) {
+				resolvedUrl = client.getThreadMessageUrl(messageKey);
+			} else if (type.equals(WebConstants.REPLY_TYPE)) {
+				resolvedUrl = client.getReplyMessageUrl(messageKey);
+			} else {
+				throw new IllegalArgumentException("Do not support type "+type);
+			}
+			response.sendRedirect(resolvedUrl.toString());
+		} catch (SynapseException e) {
```
jay-hodgson: What is the behavior if an unsupported reply_type is passed in?  It would not be caught by this try/catch.  Change to catch Exception?

src/main/java/org/sagebionetworks/web/server/servlet/DiscussionMessageServlet.java
```diff
@@ -0,0 +1,93 @@
...
+			response.sendError(HttpServletResponse.SC_BAD_REQUEST, e.getMessage());
+		}
+	}
+
+	/**
+	 * Create a new Synapse client.
+	 *
+	 * @return
+	 */
+	private SynapseClient createNewClient(String sessionToken) {
```
jay-hodgson, line 83: We should refactor this into shared code, I think it's been copied in each httpservlet.

PR #2085, opened 2016-02-03
ask anonymous to login before creating a new thread/reply

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadWidget.java
```diff
@@ -204,7 +214,12 @@ public void configureReplies() {
 
 	@Override
 	public void onClickNewReply() {
-		newReplyModal.show();
+		if (!authController.isLoggedIn()) {
+			view.showErrorMessage(DisplayConstants.ERROR_LOGIN_REQUIRED);
```
jay-hodgson, line 218: Change to have more specific message.

PR #2086, opened 2016-02-03
SWC-2873: support single thread view, and add generic area token params for the discussion tab

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/place/ParameterizedToken.java
```diff
@@ -0,0 +1,68 @@
+package org.sagebionetworks.web.client.place;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+import com.google.gwt.core.client.GWT;
+
+public class ParameterizedToken {
+	public static final String DEFAULT_TOKEN = "0";
```
jay-hodgson, line 10: change to "default", or something...

PR #2096, opened 2016-02-08
SWC-2877: add help support place

jay-hodgson:
(behind alpha until discussion feature is release)

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/SynapseForumPresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/SynapseForumPresenter.java
@@ -1,0 +1,171 @@
+package org.sagebionetworks.web.client.presenter;
+
+import java.util.HashMap;
+
+import org.sagebionetworks.repo.model.discussion.Forum;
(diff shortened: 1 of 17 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/SynapseForumPresenter.java
```diff
@@ -0,0 +1,172 @@
...
+		showForum(globalApplicationState.getSynapseProperty(WebConstants.FORUM_SYNAPSE_ID_PROPERTY));
+	}
+	
+
+	@Override
+	public void setPlace(SynapseForumPlace place) {
+		this.place = place;
+		this.view.setPresenter(this);
+//		String threadId = place.getParam(DiscussionTab.THREAD_ID_KEY);
+		isCurrentUserModerator = DisplayUtils.isInTestWebsite(cookies);
```
jay-hodgson, line 98: always hide in forum place?

jay-hodgson, line 98: or base on place param

PR #2104, opened 2016-02-12
Refactor

kimyen:


src/main/java/org/sagebionetworks/web/client/presenter/SynapseForumPresenter.java
```diff
@@ -34,50 +27,40 @@
...
-	private String forumId;
 	
 	MarkdownWidget wikiPage;
 	WikiPageKey pageKey;
 	Boolean isCurrentUserModerator;
 	CookieProvider cookies;
-	
+	ForumWidget forumWidget;
+
+	private String areaToken = null;
```
jay-hodgson: areaToken is never set.  Please expose a method in ParameterizedPlace to get the ParameterizedToken (instead of recreating based on the place.toToken())

src/main/java/org/sagebionetworks/web/client/view/SynapseForumViewImpl.java
```diff
@@ -115,4 +68,9 @@ public void setWikiWidget(Widget w) {
 		wikiContainer.clear();
 		wikiContainer.add(w);
 	}
+
+	@Override
+	public void setForumWidget(Widget widget) {
+		forumWidgetContainer.add(widget);
```
jay-hodgson, line 74: Thank you for this refactor, this is so much better!

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadListWidget.java
```diff
@@ -78,21 +83,23 @@ public void onFailure(Throwable caught) {
...
+									configure(forumId, isCurrentUserModerator, emptyListCallback);
+								}
+							});
 							view.addThread(thread.asWidget());
 						}
 						offset += LIMIT;
 						long numberOfThreads = result.getTotalNumberOfResults();
 						view.setLoadingVisible(false);
 						view.setLoadMoreButtonVisibility(offset < numberOfThreads);
-						if (numberOfThreads > 0) {
```
jay-hodgson, line 88: I would expect the thread list widget to own the "empty" message UI (but I suspect you'll still need to call back to the parent so that visibility of other UI, like create thread, is changed).

PR #2125, opened 2016-02-26
Fully working with both unit tests in my nodeJS module as well as an …

Njanderson:
…in-Synapse integration test of every case I could think of


src/main/webapp/js/markdown-it/markdown-it-math.js
```diff
@@ -0,0 +1,152 @@
+/*! markdown-it-math 3.0.2 https://github.com/runarberg/markdown-it-math @license MIT */
+(function(f){if(typeof exports==="object"&&typeof module!=="undefined"){module.exports=f()}else if(typeof define==="function"&&define.amd){define([],f)}else{var g;if(typeof window!=="undefined"){g=window}else if(typeof global!=="undefined"){g=global}else if(typeof self!=="undefined"){g=self}else{g=this}g.markdownitMath = f()}})(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
+'use strict';
```
jay-hodgson: Please load minified version instead.  We'll need to swap in the full versions when debugging, but in general users should be async loading the compressed markdown it files.

jay-hodgson:
Thanks.  
Later, we should consider injecting the full (non-minified) js files from the server when going into alpha mode (as a possible trigger) to make our lives easier.

Njanderson:
That would be cool! I removed some debugging help already, too. There used
to be a different renderer for block and inline that just gave an arbitrary
class to differentiate block and inline.

PR #2132, opened 2016-02-29
Support external images in discussion

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/entity/MarkdownEditorWidgetViewImpl.java
```diff
@@ -385,4 +388,9 @@ public void setVideoCommandsVisible(boolean visible) {
 		videoLink.setVisible(visible);
 		videoButton.setVisible(visible);
 	}
+
+	@Override
+	public void setExternalImageButtonVisible(boolean visible) {
+		externalImageButton.setVisible(visible);
```
jay-hodgson, line 395: I realize it's not currently necessary (since this is set in the ui.xml), but could you add externalImageButton.setVisible(false) in the view.clear() method for me?

src/main/java/org/sagebionetworks/web/client/widget/entity/editor/ExternalImageConfigViewImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+	@Override
+	public void initView() {
+		urlField.setValue("");
+		nameField.setValue("");
+	}
+
+	@Override
+	public void checkParams() throws IllegalArgumentException {
+		if (!ValidationUtils.isValidUrl(urlField.getValue(), false))
```
jay-hodgson: I realize this is a copy, but feel it my obligation to point out that validation should not occur in the view impl (should be in the presenter, and tested).

src/main/java/org/sagebionetworks/web/client/widget/entity/registration/WidgetRegistrarImpl.java
```diff
@@ -61,6 +61,8 @@ public WidgetEditorPresenter getWidgetEditorForWidgetDescriptor(WikiPageKey wiki
 			presenter = ginInjector.getProvenanceConfigEditor();
 		} else if (contentTypeKey.equals(WidgetConstants.IMAGE_CONTENT_TYPE)) {
 			presenter = ginInjector.getImageConfigEditor();
+		} else if (contentTypeKey.equals(WidgetConstants.EXTERNAL_IMAGE_CONTENT_TYPE)) {
+			presenter = ginInjector.getExternalImageConfigEditor();
```
jay-hodgson, line 65: Will probably also want to call registerWidget in initWithKnownWidgets() for this new content type (see the dialog title in the new editor).

PR #2136, opened 2016-03-01
SWC-2918: autolink syn ids, and send most links to a new window by default.

jay-hodgson:
Also includes a refactor (move this out of SynapseJSNIUtils), and support for \@username notation (will autolink to user profile page, where it attempts to resolve the username2userid).


jay-hodgson:
included in https://github.com/Sage-Bionetworks/SynapseWebClient/pull/2137

PR #2137, opened 2016-03-01
SWC-2935 (and SWC-2918)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/MarkdownItImpl.java
```diff
@@ -0,0 +1,281 @@
...
+					return 0;
+				},
+				normalize : function(match) {
+					match.url = '#!Profile:' + match.url.replace(/^@/, '');
+				}
+			});
+
+			$wnd.md.linkify.add('syn', {
+				validate : function(text, pos, self) {
+					var tail = text.slice(pos);
```
xschildw, line 63: Use synRE below.

PR #2205, opened 2016-04-11
SWC-3009: from code review

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/place/ParameterizedPlace.java
```diff
@@ -23,6 +23,13 @@ public void putParam(String key, String value) {
 		parameterizedToken.put(key, value);
 		updateToken();
 	}
+	
+	public void setParameterizedToken(ParameterizedToken token) {
+		parameterizedToken = token;
+		updateToken();
```
kimyen: remove token as member variable

PR #2230, opened 2016-04-19
SWC-3018: ability to specify a filter to the entity finder

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
```diff
@@ -50,51 +55,95 @@ public void clearState() {
...
+				};
+			});
 		}
 	}
+	
+	public boolean validateEntityTypeAgainstFilter(Entity entity) {
+		boolean isCorrectType = true;
+		synAlert.clear();
+		switch (filter) {
+			case CONTAINER:
```
jay-hodgson, line 102: Move entity type check logic to Filter.

PR #2248, opened 2016-04-25
version scripts to update cache

jay-hodgson:


jay-hodgson:
Portal.html is cached as well (though it has no cache related headers).  So just deploying this change could make things much worse (since the cached Portal.html references js files that have been renamed).

jay-hodgson:
https://sagebionetworks.jira.com/browse/SWC-3038

PR #2261, opened 2016-04-28
SWC-3048: override OSX hidden scrollbars, and clean up scrollbars

jay-hodgson:
(now that I can see them!)


jay-hodgson:
Same as https://github.com/Sage-Bionetworks/SynapseWebClient/pull/2262 but pointing to develop.

PR #2292, opened 2016-05-12
SWC-3058: support project filters (including projects from a specific team)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/place/Profile.java
```diff
@@ -14,24 +15,56 @@
...
+	
 	private String token;
 	private String userId;
+	private String teamId;
 	private ProfileArea area;
+	private ProjectFilterEnum projectFilter;
 	private boolean noRestartActivity;
 	
 	public Profile(String token) {
 		this.token = token;
```
jay-hodgson: token.toLowerCase().  What about tokenizing token??

PR #2307, opened 2016-05-20
Cache portal properties

jay-hodgson:


xschildw:
Need to fix LocalStorage before pulling this in.

PR #2312, opened 2016-05-24
SWC-3095: when profile page is initialized, replace state instead of pushing new items onto the browser history

jay-hodgson:
on tab click (or project filter), push place into browser history.

### DIFF
--- a/src/test/java/org/sagebionetworks/web/unitclient/presenter/ProfilePresenterTest.java
+++ b/src/test/java/org/sagebionetworks/web/unitclient/presenter/ProfilePresenterTest.java
@@ -333,9 +333,6 @@ public class ProfilePresenterTest {
 		profilePresenter.setPlace(place);
 		verify(mockUserProfileClient).getUserBundle(anyLong(), anyInt(), any(AsyncCallback.class));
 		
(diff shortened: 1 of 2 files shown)


src/test/java/org/sagebionetworks/web/unitclient/presenter/ProfilePresenterTest.java
```diff
@@ -1289,15 +1285,26 @@ public void testCertificationBadgeClicked() {
...
+		profilePresenter.updateArea(ProfileArea.CHALLENGES, pushState);
+		verify(mockGlobalApplicationState).replaceCurrentPlace(any(Profile.class));
+	}
 
 	@Test
 	public void testUpdateAreaNoChange() {
 		profilePresenter.setPlace(place);
 		when(place.getArea()).thenReturn(ProfileArea.PROJECTS);
-		profilePresenter.updateArea(ProfileArea.PROJECTS);
+		boolean pushState = true;
```
xschildw, line 1306: Set to false and use below?

jay-hodgson, line 1306: Oops, yes.  In fact, the test verification does not make sense now that the place changer isn't used!

PR #2314, opened 2016-05-26
SWC-3100: add scope widget (entity container list widget).

jay-hodgson:
Rework wizard to have ability to create a TableEntity as well.  Change table/view listing UI a bit (shrink rows slightly, and add icon to distinguish type).

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizard.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizard.java
@@ -8,26 +8,34 @@ import com.google.gwt.user.client.ui.Widget;
 import com.google.inject.Inject;
 
 /**
- * Wizard used to create a new File View
+ * Wizard used to create a new Table or View
  * @author Jay
  *
  */
-public class CreateFileViewWizard {
+public class CreateTableViewWizard {
 	
 	ModalWizardWidget modalWizardWidget;
-	CreateFileViewWizardStep1 step1;
+	CreateTableViewWizardStep1 step1;
+	public enum TableType {
+	    table,
+	    view
+	}
 
 	@Inject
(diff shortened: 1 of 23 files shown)


src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityContainerListWidgetViewImpl.java
```diff
@@ -0,0 +1,93 @@
...
+	}
+	
+	@Override
+	public Widget asWidget() {
+		return widget;
+	}
+
+	@Override
+	public void addEntity(final String id, String name, boolean showDeleteButton) {
+		final Div row = new Div();
```
jay-hodgson, line 55: Pull into it's own widget.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizard.java
```diff
@@ -8,26 +8,34 @@
...
 		this.modalWizardWidget = modalWizarWidget;
-		this.modalWizardWidget.setTitle("Create File View");
 		this.modalWizardWidget.setModalSize(ModalSize.LARGE);
 		this.step1 = step1;
 		this.modalWizardWidget.configure(this.step1);
 	}
 
-	public void configure(String parentId) {
-		this.step1.configure(parentId);
+	public void configure(String parentId, TableType type) {
```
jay-hodgson, line 32: Clear on configure!

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizard.java
```diff
@@ -8,26 +8,34 @@
...
+		if (TableType.view.equals(type)) {
```
jay-hodgson, line 33: Next without container definition should function.

PR #2319, opened 2016-06-07
SWC-3100: use new Table interface instead of FileView and TableEntity

jay-hodgson:
also update the file view icon, and expose the Table wizard in alpha mode (on Add Table button click)


john-hill:
Looks good.

PR #2325, opened 2016-06-09
Empty Docker Tab under Alpha Mode

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -370,6 +401,11 @@ public void configureDiscussionTab() {
 		}
 		discussionTab.configure(projectId, projectHeader.getName(), discussionAreaToken, canModerate, moderatorIds);
 	}
+
+	public void configureDockerTab() {
+		String projectId = projectHeader.getId();
```
jay-hodgson, line 406: Current entity could be the repoentity, so it will need both (like the tables tab).  Something like:
`dockerTab.setProject(projectHeader.getId(), projectBundle, projectBundleLoadError);`
`dockerTab.configure(entity, entityUpdateHandler, tablesAreaToken);`

src/test/java/org/sagebionetworks/web/unitclient/place/SynapsePlaceTest.java
```diff
@@ -105,6 +106,18 @@ public void testDiscussionVersionCase() {
 		Assert.assertNull(place.getAreaToken());
 		Assert.assertEquals(testToken, tokenizer.getToken(place));
 	}
+
+	@Test
+	public void testDockerVersionCase() {
+		String testToken = testEntityId + VERSION_DELIMITER + testVersionNumber + DOCKER_DELIMITER;
```
jay-hodgson, line 112: I wonder if versioned docker area case is a real world use case (since they will currently live at the project level, which is not versionable).

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/tabs/DockerTabTest.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	@Test
+	public void testConfigureNotInAlpha() {
+		String entityId = "syn1"; 
+		String entityName = "Docker project test";
+		String areaToken = "a=b&c=d";
+		when(mockCookies.getCookie(DisplayUtils.SYNAPSE_TEST_WEBSITE_COOKIE_KEY)).thenReturn(null);
+		tab.configure(entityId, entityName, areaToken);
+		verify(mockTab).setTabListItemVisible(false);
```
jay-hodgson, line 62: Is this the right place to hide the tab list item?  Since configure of the tab is deferred, does that mean the Docker tab will flash visible in some cases with this code?  What do you think about setting the tab list item to invisible (unless in alpha mode) in EntityPageTop?

src/test/java/org/sagebionetworks/web/unitclient/place/SynapsePlaceTest.java
kimyen, line 112: Discussion does not have version either. What do you think about removing both tests (testDiscussionVersionCase and testDockerVersionCase)?

jay-hodgson, line 112: Agree that they can be removed.

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/tabs/DockerTabTest.java
kimyen, line 62: Agree. I will move it to EntityPageTop.

PR #2329, opened 2016-06-10
Docker layout

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/DockerTab.java
```diff
@@ -48,15 +51,16 @@ public DockerTab(
 		this.view = view;
 		this.tab = tab;
 		this.dockerTitleBar = dockerTitleBar;
-		this.dockerListWidget = dockerListWidget;
+		this.dockerListRepoWidget = dockerListRepoWidget;
```
jay-hodgson, line 54: missing?
this.dockerRepoWidget = dockerRepoWidget;

PR #2340, opened 2016-06-15
SWC-3120: ability to create view after refactor (in stack-140)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep1.java
```diff
@@ -69,10 +70,11 @@ private void createFileViewEntity(final String name) {
 		modalPresenter.setLoading(true);
 		Table table;
 		if (TableType.view.equals(tableType)) {
-			table = new FileView();
+			table = new EntityView();
 			List<String> scopeIds = entityContainerList.getEntityIds();
-			((FileView)table).setScopeIds(scopeIds);
-			table.setEntityType(FileView.class.getName());
+			((EntityView)table).setScopeIds(scopeIds);
+			table.setEntityType(EntityView.class.getName());
```
john-hill, line 76: not sure if we should set entity type.

PR #2356, opened 2016-06-21
SWC-3109: scroll down to load more threads/ replies

kimyen:


src/main/java/org/sagebionetworks/web/client/widget/discussion/SingleDiscussionThreadWidget.java
```diff
@@ -158,8 +159,28 @@ public void invoke() {
...
 	}
-	
+
+	public void checkForInViewAndLoadData() {
+		if (!view.isLoadMoreAttached()) {
+			//Done, view has been detached and widget was never in the viewport
+			return;
+		} else if (view.isLoadMoreInViewport() && view.getLoadMoreVisibility()) {
+			//try to load data!
+			loadMore();
```
jay-hodgson, line 177: If we get into loadMore once, what causes the code to recheck later (even if the loadMoreVisibility is set to true)?

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadListWidget.java
```diff
@@ -59,12 +66,31 @@ public void configure(String forumId, Boolean isCurrentUserModerator, Set<Long>
...
+			}
+		};
 		loadMore();
 		DiscussionThreadCountAlert threadCountAlert = ginInjector.getDiscussionThreadCountAlert();
 		view.setThreadCountAlert(threadCountAlert.asWidget());
 		threadCountAlert.configure(forumId);
 	}
 
+	public void checkForInViewAndLoadData() {
+		if (!view.isLoadMoreAttached()) {
```
jay-hodgson, line 82: Shouldn't this loop also end if the load more is no longer visible?  (!view.getLoadMoreVisibility()).

---

Nevermind the comment above.  If there's an error in the rpc (getThreadsForForum()), then the error will be shown by the synAlert, and I do think that this loop should continue.  Every 2 seconds, it will retry the rpc until success.

jay-hodgson:
Approved the code, but I'd like to see a demo before merging it in.  Nice work Kim (as usual).

PR #2357, opened 2016-06-23
SWC-3135:  Add area to show EntityView scope, and ability to edit.

jay-hodgson:
### New Scope area (collapse)

expanded scope area
### Edit Scope

edit


john-hill:
Looks Good!

PR #2361, opened 2016-06-23
SWC-3140: show wiki page title in confirm dialog (unless root wiki page)

jay-hodgson:
And go to parent wiki page on successful delete.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -15,6 +15,7 @@ import org.sagebionetworks.repo.model.Versionable;
 import org.sagebionetworks.repo.model.auth.UserEntityPermissions;
 import org.sagebionetworks.repo.model.docker.DockerRepository;
 import org.sagebionetworks.repo.model.table.Table;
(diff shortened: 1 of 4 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -849,25 +852,45 @@ public void invoke() {
...
-				postConfirmedDeleteWiki();
+			public void onSuccess(V2WikiPage page) {
+				// Confirm the delete with the user.
+				final String parentWikiId = page.getParentWikiId();
+				String confirmMessage;
+				if (parentWikiId == null) {
+					confirmMessage = ARE_YOU_SURE_YOU_WANT_TO_DELETE+THE_ROOT_WIKI_PAGE_AND_ALL_SUBPAGES;
+				} else {
+					confirmMessage = ARE_YOU_SURE_YOU_WANT_TO_DELETE+"the wiki page \""+page.getTitle()+"\" and all subpages?";
+				}
```
brucehoff, line 869: what if there is no title?

jay-hodgson, line 869: It was only recently that I became aware that this was something that the backend supports.  I've changed it so that it shows the wiki ID in this case (in this confirmation message, and in the wiki subtree)

PR #2378, opened 2016-07-01
SWC-3156

kimyen:


src/main/resources/org/sagebionetworks/web/client/widget/discussion/ForumWidgetViewImpl.ui.xml
```diff
@@ -10,8 +10,21 @@
 					<b:Button ui:field="newThreadButton" type="PRIMARY" pull="RIGHT">New Thread</b:Button>
 				</b:Tooltip>
 				<bh:Span ui:field="subscribeButtonContainer" addStyleNames="right margin-right-10" />
+				<b:Button ui:field="deletedThreadButton" type="DEFAULT" pull="RIGHT" icon="TRASH_O" addStyleNames="right margin-right-10" >Deleted Threads</b:Button>
 			</b:Column>
 		</b:Row>
+		<bh:Div ui:field="deleteThreadsArea" addStyleNames="highlight-box padding-top-0-imp margin-bottom-20" visible="false">
```
jay-hodgson, line 16: This could be converted into a format that supports HelpWidget (now, or later), see SettingsViewImpl.ui.xml for an example.

PR #2384, opened 2016-07-06
SWC-3156: change icon

kimyen:


jay-hodgson:
Please create another for release-143 branch

PR #2390, opened 2016-07-07
Refactor "register for Synapse" widget

jay-hodgson:
In preparation for home page redesign, remove duplicate code in the RegisterAccount place


jay-hodgson:
### Registration widget on the home page

screen shot 2016-07-07 at 2 58 23 pm

### Register page now uses the same widget:

screen shot 2016-07-07 at 2 57 32 pm

PR #2393, opened 2016-07-08
SWC-3172: show login widget in Synapse Alert instead of button.

jay-hodgson:
Has lazy construction (via gin injector) of LoginWidget, since the SynapseAlert only needs it when showLogin() is called.


jay-hodgson:
screen shot 2016-07-08 at 4 03 31 pm

src/main/java/org/sagebionetworks/web/client/PortalGinInjector.java
```diff
@@ -399,4 +400,6 @@
 
 	// docker
 	public DockerRepoWidget createNewDockerRepoWidget();
+	
+	public LoginWidget getLoginWidget();
```
kimyen, line 404: check to see if LoginWidget is singleton

PR #2406, opened 2016-07-13
SWC-3183: add discussion moderator permission level

jay-hodgson:
set of access types READ and MODERATE


src/main/java/org/sagebionetworks/web/client/widget/sharing/AccessControlListEditor.java
```diff
@@ -119,11 +119,12 @@ private void initViewPrincipalIds(){
 		}
 	}
 	public PermissionLevel[] getPermList() {
-		return new PermissionLevel[] {PermissionLevel.CAN_VIEW, PermissionLevel.CAN_EDIT, PermissionLevel.CAN_EDIT_DELETE, PermissionLevel.CAN_ADMINISTER};
```
jay-hodgson, line 122: only show moderate if Project

PR #2408, opened 2016-07-14
SWC-3053: support viewing a single thread reply (url place area parameter)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/discussion/SingleDiscussionThreadWidget.java
```diff
@@ -363,7 +374,47 @@ public void invoke() {
...
+				replyWidget.configure(bundle, isCurrentUserModerator, moderatorIds, refreshCallback, replyClickedCallback);
+				view.addReply(replyWidget.asWidget());
+			}
+		});
+	}
+
+	@Override
+	public void onClickShowAllReplies() {
+		setReplyId(null);
+		configureReplies();
```
jay-hodgson, line 408: need to watch for new replies

PR #2413, opened 2016-07-15
focus based does not work for longer clicks in popover

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/HelpWidget.java
```diff
@@ -68,16 +79,20 @@ public void setHelpMarkdown(String md) {
 	
 	public void setHref(String fullHelpHref) {
 		if (DisplayUtils.isDefined(fullHelpHref)) {
-			this.moreHelpHTML = "<form action=\""+SafeHtmlUtils.htmlEscape(fullHelpHref)+"\" target=\"_blank\"><button class=\"btn btn-primary btn-xs right\" type=\"submit\">More info</button></form>";
+			this.moreHelpHTML = "<button class=\"btn btn-primary btn-xs right\" onClick=\"window.open('"+SafeHtmlUtils.htmlEscape(fullHelpHref)+"');"+closePopoverJs+"\">More info</button>";
 		}
 	}
```
xschildw, line 85: Not needed

src/main/java/org/sagebionetworks/web/client/widget/HelpWidget.java
```diff
@@ -35,6 +36,9 @@
  *
  */
 public class HelpWidget implements IsWidget {
+	// 10 second delay, then auto hide
```
xschildw, line 39: Not needed.

PR #2418, opened 2016-07-18
SWC-3081: change UI for reset query

jay-hodgson:
to a button, that stands out on the danger background


jay-hodgson:
## Before

reset query - before

## After

reset query - after

PR #2419, opened 2016-07-19
SWC-3052: refactor preview so that it's available wherever markdown is written

jay-hodgson:
Also speed up modal dialog popup (by removing the animation), changing the preview so that it replaces the editor inline, and make the discussion related post buttons (for thread and reply) show up in a consistent position.


jay-hodgson:
## wiki - edit mode

wiki - edit mode

## wiki - preview mode

wiki - preview mode

## thread - edit mode

thread - edit mode

## thread - preview mode

thread - preview mode

## reply - edit mode

reply - edit mode

## reply - preview mode

reply - preview mode

PR #2420, opened 2016-07-19
SWC-3197: make copying text more convenient (and functional) across browsers

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/file/Md5Link.java
```diff
@@ -29,9 +30,11 @@ public void onClick(ClickEvent event) {
 		});
 		DisplayUtils.addTooltip(md5Link, md5, Placement.BOTTOM);
 		this.add(md5Link);
+		this.add(copyTextModal.asWidget());
```
jay-hodgson, line 33: Doesn't simple panel only support a single widget?

PR #2470, opened 2016-08-17
SWC-3237: also render the evaluation ID

jay-hodgson:


jay-hodgson:
screen shot 2016-08-17 at 3 01 13 pm

PR #2477, opened 2016-08-22
SWC-3114: expose ability to update table schema

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep2.java
```diff
@@ -34,6 +40,7 @@
 	Table entity;
 	TableType tableType;
 	SynapseClientAsync synapseClient;
+	JobTrackingWidget jobTrackingWidget;
```
john-hill: Since not using the widget just use the AsynchJobTracker.

john-hill, line 44: show the job tracking widget

PR #2478, opened 2016-08-23
SWC-3243: show docker commit tags

kimyen:
screen shot 2016-08-23 at 10 38 33 am


src/main/resources/org/sagebionetworks/web/client/widget/docker/DockerRepoWidgetViewImpl.ui.xml
```diff
@@ -27,6 +27,7 @@
 					</b:Panel>
 				</b:Column>
 				<b:Column size="XS_12, MD_6">
+					<bh:Div ui:field="dockerCommitListContainer" addStyleNames="highlight-box" />
```
jay-hodgson: instead of highlight-box method, see FilesTabViewImpl.ui.xml (to support contextual help).  this also applies to the provenance container.

PR #2511, opened 2016-09-08
SWC-3265,SWC-2241,SWC-515,SWC-962: Rewrite entity list widget

jay-hodgson:
Old widget could cause throttling problems for the client.  Rewrite so that it's composed of a number of widgets (UserBadge, SelectionToolbar, FileDownloadButton) to fix a number of old issues. Improves UI behavior (responsive layout, load entity bundle iff view is attached, configured, and in view).


jay-hodgson:
Kept the basic functionality of the widget.

### Old editor

old editor

### New editor

new editor

### Old renderer

old renderer

### New renderer

new renderer

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityListRowBadge.java
```diff
@@ -0,0 +1,213 @@
...
+		} else if (view.isInViewport()) {
+			//try to load data!
+			getEntityBundle();
+		} else {
+			//wait for a few seconds and see if we should load data
+			gwt.scheduleExecution(invokeCheckForInViewAndLoadData, DisplayConstants.DELAY_UNTIL_IN_VIEW);
+		}
+	}
+	@Override
+	public void viewAttached() {
```
kimyen, line 99: try making an interface for the view and abstract class for presenter

PR #2514, opened 2016-09-09
SWC-3245: change 'associated to' to 'associated with'

jay-hodgson:
…and fix field label inconsistency.


brucehoff:
Looks great!

PR #2518, opened 2016-09-13
SWC-3270: Add a helper class to support lazy loading of data (only load when attached, configured, and in the viewport)

jay-hodgson:
Refactor only, code was copied in 2 places.


jay-hodgson:
Approved by Xa

PR #2519, opened 2016-09-13
SWC-3266

kimyen:
restore a deleted thread


kimyen:
test against a local PLFM

src/main/java/org/sagebionetworks/web/client/widget/discussion/SingleDiscussionThreadWidget.java
```diff
@@ -190,6 +201,7 @@ private void configureView(DiscussionThreadBundle bundle) {
 		view.setDeletedThreadVisible(isDeleted);
 		view.setReplyContainerVisible(!isDeleted);
 		view.setCommandsVisible(!isDeleted);
+		view.setRestoreIconVisible(isDeleted);
```
jay-hodgson, line 204: Is there no permission check here because you can only configure the widget with a deleted thread bundle if the user is a moderator?  If this isn't the case, then I would expect an "else { view.setRestoreIconVisible(isCurrentUserModerator)}; }" below...

PR #2528, opened 2016-09-15
Refactor selection toolbar presenter logic

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/SelectionToolbarPresenter.java
```diff
@@ -0,0 +1,114 @@
+package org.sagebionetworks.web.client.widget;
```
jay-hodgson: Refactor so that this is a helper widget (encapsulating selection toolbar presenter logic, but not imposing hierarchy).

PR #2530, opened 2016-09-16
Add a table entity column renderer (lazy load based on what is in the viewport)

jay-hodgson:


jay-hodgson:
screen shot 2016-09-16 at 9 47 16 am

jay-hodgson:
Approved by Xa

PR #2531, opened 2016-09-16
convert getEntity to getEntityHeader when possible (per John)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/cell/EntityIdCellRendererImpl.java
```diff
@@ -37,12 +40,15 @@ public void invoke() {
...
-					entityName = entity.getName();
-					view.setIcon(EntityTypeUtils.getIconTypeForEntity(entity));
-					view.setLinkText(entityName);
+				public void onSuccess(ArrayList<EntityHeader> results) {
+					if (results.size() == 1) {
+						EntityHeader entity = results.get(0);
+						entityName = entity.getName();
+						view.setIcon(EntityTypeUtils.getIconTypeForEntityClassName(entity.getType()));
+						view.setLinkText(entityName);
+					}
```
brucehoff, line 55: else (if results.size()!=1):  do whatever you do onFailure

PR #2536, opened 2016-09-20
SWC-3271

kimyen:
only show AddExternalRepo button to users who can create repo


src/test/java/org/sagebionetworks/web/unitclient/widget/docker/DockerRepoListWidgetTest.java
```diff
@@ -69,6 +72,8 @@ public void setUp() {
 		projectId = "syn123";
 		when(mockProjectBundle.getEntity()).thenReturn(mockProject);
 		when(mockProject.getId()).thenReturn(projectId);
+		when(mockProjectBundle.getPermissions()).thenReturn(mockUserEntityPermissions);
+		when(mockUserEntityPermissions.getCanAddChild()).thenReturn(true);
```
jay-hodgson, line 76: canAddChild is the correct permission to check in this case?

PR #2543, opened 2016-09-22
SWC-3292: add Google map to team page in alpha mode

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/googlemap/GoogleMapViewImpl.java
```diff
@@ -0,0 +1,198 @@
...
+
+	@Override
+	public void showMap(String data) {
+		JSONArray jsonArray = (JSONArray) JSONParser.parseStrict(data);
+		Element markerPopupContentEl = markerPopupContent.getElement();
+		Element el = googleMapCanvas.getElement();
+		JavaScriptObject map = _createMap(el);
+		JavaScriptObject bounds  = _getBounds();
+		
+		userBadges.clear();
```
jay-hodgson, line 107: no need to clear user badges here

PR #2573, opened 2016-10-07
SWC-3325 Changed xml so that Created By tab is shown at default zoom

kmaltergott:


src/main/resources/org/sagebionetworks/web/client/widget/entity/renderer/EntityListWidgetViewImpl.ui.xml
```diff
@@ -26,7 +26,7 @@
 						<t:TableHeader width="165px" addStyleNames="visible-md visible-lg">
 							<bh:Span text="Date" />
 						</t:TableHeader>
-						<t:TableHeader width="310px" addStyleNames="visible-xl">
+						<t:TableHeader width="310px" addStyleNames="visible-md visible-lg">
```
jay-hodgson, line 29: I believe this will hide the column on xl devices (please see the table in the CSS responsive layout utils section of the bootstrap docs).

kmaltergott: How should I go about changing this then? Leaving in the visible-xl and adding visible-lg? Or should we set the width of the columns to be percentage-based instead of pixel-based? 
Also, shouldn't the other columns have visible-xl included as well - it seems strange that we would hide Created On and Note columns for extra large devices

jay-hodgson: Forgot we extended the bootstrap classes, visible-xl is our own media rule.  I'll take a look at what you have today...

PR #2575, opened 2016-10-10
Swc 3274

kmaltergott:


src/main/java/org/sagebionetworks/web/client/widget/user/UserBadgeViewImpl.java
```diff
@@ -187,6 +187,13 @@ public void setHref(String href) {
 	public void clearHref() {
 		usernameLink.setHref(HasHref.EMPTY_HREF);
 	}
+	
+	@Override
+	public void openNewWindow(String url) {
+		usernameLink.setTarget("_blank");
+		DisplayUtils.newWindow(url, "", "");
```
jay-hodgson: no need to set usernameLink target.

kmaltergott: If I don't set the target, then the link doesn't open a new window when clicked

jay-hodgson: Really interesting that the anchor target is used after the click event has taken place!  Well, I don't trust that every browser is going to handle this properly, please set the name value in the js Window open() method.
DisplayUtils.newWindow(url, "_blank", "");

src/main/java/org/sagebionetworks/web/client/widget/user/UserBadgeViewImpl.java
```diff
@@ -187,6 +187,18 @@ public void setHref(String href) {
...
+	
+	@Override
+	public void openNewWindow(String url) {
+		DisplayUtils.newWindow(url, "_blank", "");
+		
+	}
+	
+	@Override
+	public void setOpenNewWindow(boolean value) {
+		usernameLink.setTarget(value ? "_blank" : "");
```
jay-hodgson: Please move this logic (the ternary operator is really an if/else) to the presenter and test.

PR #2578, opened 2016-10-12
SWC-3327: only show container download instructions if entity is a folder

jay-hodgson:


jay-hodgson:
Previously merged into develop.  Merging to release-155

PR #2582, opened 2016-10-13
Swc 3322

kmaltergott:
I'm not sure about the configure method I added in VideoWidget - should it be overriding a method defined in WidgetRendererPresenter the way the existing configure method is?  Or should I leave it as simply a public method?

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/PreviewWidgetView.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/PreviewWidgetView.java
@@ -1,5 +1,7 @@
 package org.sagebionetworks.web.client.widget.entity;
 
+import org.sagebionetworks.web.client.widget.entity.renderer.VideoWidget;
+
 import com.google.gwt.event.dom.client.ErrorEvent;
 import com.google.gwt.user.client.ui.IsWidget;
 import com.google.gwt.user.client.ui.Widget;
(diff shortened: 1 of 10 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/PreviewWidgetView.java
```diff
@@ -14,6 +16,7 @@
 	public void setImagePreview(String fullFileUrl, String previewUrl);
 	public void setCodePreview(String text);
 	public void setTextPreview(String text);
+	public void setVideoPreview(VideoWidget vw);
```
jay-hodgson: Please change to a generic setPreviewWidget(Widget w); (and do the videoWidget.asWidget() in the presenter)

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/VideoWidget.java
```diff
@@ -33,12 +34,19 @@ public void configure(final WikiPageKey wikiKey, final Map<String, String> widge
...
 		String width = descriptor.get(WidgetConstants.VIDEO_WIDGET_WIDTH_KEY);
 		String height = descriptor.get(WidgetConstants.HEIGHT_KEY);
-		view.configure(wikiKey,	mp4SynapseId, oggSynapseId, webmSynapseId, width, height, authenticationController.isLoggedIn(), wikiVersionInView,
-				authenticationController.getCurrentXsrfToken());
+		view.configure(mp4SynapseId, oggSynapseId, webmSynapseId, width, height, authenticationController.getCurrentXsrfToken());
 		//set up view based on descriptor parameters
 		descriptor = widgetDescriptor;
 	}
 	
+	public void configure(String synapseId, String filename, int width, int height) {
```
jay-hodgson, line 42: New configure method looks ok.  
The WidgetRendererPresenter.configure() is implemented if it is to be supported in the wiki markdown system.

PR #2584, opened 2016-10-15
SWC-3331

kimyen:


jay-hodgson:
Thanks for refactoring this into a standalone widget!

PR #2604, opened 2016-10-31
cache wiki page in session storage

jay-hodgson:


jay-hodgson:
reviewed/approved by Xa.

PR #2610, opened 2016-11-03
SWC 3309 Changed wording of dialog to include all notification types

kmaltergott:


src/main/resources/org/sagebionetworks/web/client/view/SignedTokenViewImpl.ui.xml
```diff
@@ -20,7 +20,7 @@
 				<b:ModalBody>
 					<g:SimplePanel ui:field="unsubscribeUserBadgeContainer" />
 					<bh:Paragraph addStyleNames="margin-top-10">
-						You'll miss out on invitations to join a team, requests to join your team, and Challenge submission notifications.
+						You’ll miss out on invitations to join a team, requests to join your team, invitations to access data, requests to view your projects,  user verification status notifications, discussion forum notifications, and challenge submission notifications.
 					</bh:Paragraph>
```
jay-hodgson, line 24: Looks like the ascii apostrophe (0x27) was also changed to a unicode right single quotation mark (U+2019).

PR #2611, opened 2016-11-03
Swc 3355

kmaltergott:


src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/FilesTab.java
```diff
@@ -318,7 +318,9 @@ public void setTargetBundle(EntityBundle bundle) {
 		breadcrumb.configure(bundle.getPath(), EntityArea.FILES);
 		view.clearActionMenuContainer();
 		//Preview
-		view.setPreviewVisible(isFile);
+		if (!bundle.getFileHandles().isEmpty()) {
```
jay-hodgson: If Folder, it used to call setPreviewVisible(false).  With this code change it wouldn't (so if you go to a file page, and then to a folder, it would likely still show the preview panel.

PR #2632, opened 2016-11-18
Swc 3378 Updated table widget to allow multiple sorting icons

kmaltergott:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TablePageWidget.java
```diff
@@ -89,6 +91,12 @@ public void configure(QueryResultBundle bundle,
 		types = ColumnModelUtils.buildTypesForQueryResults(QueryBundleUtils.getSelectFromBundle(bundle), bundle.getColumnModels());
 		// setup the headers from the types
 		List<IsWidget> headers = new ArrayList<IsWidget>();
+		Set<String> sortedHeaders = new HashSet<String>();
```
jay-hodgson: Let's use a map &lt;String, SortItem&gt; to avoid ever iterating over the sortList to find the right item.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TablePageWidget.java
```diff
@@ -97,13 +105,9 @@ public void configure(QueryResultBundle bundle,
...
 				sth.configure(type.getName(), pageChangeListener);
 				headers.add(sth);
-				if(sort != null){
-					if(headerName.equals(sort.getColumn())){
-						if(SortDirection.DESC.equals(sort.getDirection())){
-							sth.setIcon(IconType.SORT_DESC);
-						}else{
-							sth.setIcon(IconType.SORT_ASC);
-						}
+				if(sortList != null){
```
jay-hodgson: can remove (sortList != null) check, go straight to sortHeadersMap.containsKey(headerName)

PR #2642, opened 2016-11-22
Swc 3361 Added extra message to be appended in case of folder being deleted

kmaltergott:


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1075,8 +1075,13 @@ public void onFailure(Throwable caught) {
 	
 	@Override
 	public void onDeleteEntity() {
-		// Confirm the delete with the user.
-		view.showConfirmDialog(CONFIRM_DELETE_TITLE,ARE_YOU_SURE_YOU_WANT_TO_DELETE+this.enityTypeDisplay+" "+this.entity.getName(), new Callback() {
+		// Confirm the delete with the user. Mention that everything inside folder will also be deleted if this is a folder entity.
+		String display = ARE_YOU_SURE_YOU_WANT_TO_DELETE+this.enityTypeDisplay+" "+this.entity.getName()+"?";
+		if (this.enityTypeDisplay.equals("Folder")) {
+			display += DELETE_FOLDER_EXPLANATION;
```
jay-hodgson, line 1081: Rather see (entity instanceof Folder) instead of testing the display text, otherwise this looks good.

jay-hodgson: Also needs a test verifying it contains the special Folder delete confirmation message.

PR #2660, opened 2016-12-06
Swc 3393

kmaltergott:


src/test/java/org/sagebionetworks/web/unitclient/widget/entity/download/UploaderTest.java
```diff
@@ -613,17 +612,11 @@ public void testGetSelectedFilesTextMultipleFiles() {
 		when(synapseJsniUtils.getMultipleUploadFileNames(anyString())).thenReturn(new String[]{"file1", "file2"});
 		assertEquals("2 files", uploader.getSelectedFilesText());
 	}
-	
+
 	@Test
-	public void testUwrapException() {
+	public void testUploadFailed() {
 		String message = "meow there's an error.";
-		Exception ex = new Exception("exception to log");
```
kmaltergott, line 620: Didn't seem like this test was necessary since we won't be passing a wrapped exception to the uploadError method anymore...

PR #2672, opened 2016-12-08
Swc 3392: Etags do not match

kmaltergott:


src/main/java/org/sagebionetworks/web/shared/exceptions/ConflictingUpdateException.java
```diff
@@ -0,0 +1,17 @@
+package org.sagebionetworks.web.shared.exceptions;
+
+import com.google.gwt.user.client.rpc.IsSerializable;
+
+public class ConflictingUpdateException extends RestServiceException implements IsSerializable {
```
jay-hodgson, line 5: "implements IsSerializable" is not necessary (because it extends RestServiceException)

PR #2681, opened 2016-12-13
SWC-3117: add handling for nav syntax

jay-hodgson:


jay-hodgson:
https://sagebionetworks.jira.com/browse/SWC-3117
Based on feedback, it's been suggested we adopt r-studio flexdashboard YAML page header values rather than introduce new (unique to Synapse) markdown syntax.

PR #2716, opened 2017-01-18
SWC-3455: get column models for annotations found within view scope

jay-hodgson:
branched from #2714, so that PR should be reviewed first.


src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep2.java
```diff
@@ -102,6 +110,24 @@ public void onSuccess(List<ColumnModel> columns) {
...
+		synapseClient.getPossibleColumnModelsForViewScope(scope, nextPageToken, new AsyncCallback<ColumnModelPage>() {
+			@Override
+			public void onFailure(Throwable caught) {
+				presenter.setErrorMessage(caught.getMessage());
+			}
+			@Override
+			public void onSuccess(ColumnModelPage columnPage) {
+				editor.addColumns(columnPage.getResults());
+				if (columnPage.getNextPageToken() != null) {
+					getPossibleColumnModelsForViewScope(columnPage.getNextPageToken());
```
kimyen, line 126: the button state should be "saving" while we are getting all column models

PR #2727, opened 2017-01-26
SWC-3461, SWC-3462

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/QueryResultEditorWidget.java
```diff
@@ -251,6 +262,8 @@ public void onComplete(AsynchronousResponseBody response) {
 							view.showErrorDialog(errors);
 						}
 						view.showMessage(CHANGES_SUBMITTED_TITLE, CHANGES_SUBMITTED_MESSAGE);
+						Date now = new Date();
+						clientCache.put(tableId + VIEW_RECENTLY_CHANGED_KEY, "true", now.getTime() + MESSAGE_EXPIRE_TIME);
```
kimyen: Jay: only put this in the client cache if it's a view

PR #2731, opened 2017-01-30
SWC-3464:  subscriber count/list widget

jay-hodgson:
to be shown at the forum level, and the thread level.


src/main/java/org/sagebionetworks/web/client/widget/discussion/SubscribersWidget.java
```diff
@@ -0,0 +1,127 @@
...
+			}
+		});
+	}
+
+	@Override
+	public void onClickSubscribersLink() {
+		// show the dialog and start getting the subscribers
+		loadMoreWidgetContainer.clear();
+		nextPageToken = null;
+		loadMoreWidgetContainer.configure(new Callback() {
```
kimyen, line 80: could be move out of this method

PR #2740, opened 2017-02-09
Swc 3466

kmaltergott:


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
```diff
@@ -95,6 +117,23 @@ public void onSuccess(Entity result) {
 		}
 	}
 	
+	private void multiSelectionOkClicked() {
+		//check for valid selection
+		if (view.getMultiEntityText().isEmpty()) {
+			synAlert.showError(DisplayConstants.PLEASE_MAKE_SELECTION);
+		} else {
+			lookupMultiEntity(view.getMultiEntityText(), new AsyncCallback<PaginatedResults<EntityHeader>>() {
```
jay-hodgson: please clean up by removing this asynccallback.

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
```diff
@@ -58,6 +65,13 @@ public void configure(boolean showVersions, SelectedHandler<Reference> handler)
 		configure(EntityFilter.ALL, showVersions, handler);
 	}
 	
+	public void configureMultiSelect(boolean showVersions, SelectedHandler<List<Reference>> handler) {
```
jay-hodgson: I don't think showVersions is necessary in this case, but we should validate based on a filter.

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
```diff
@@ -140,6 +179,32 @@ public void onFailure(Throwable caught) {
...
+			String s = entities[i];
+			Reference r = new Reference();
+			r.setTargetId(s);
+			rList.add(r);
+		}
+		selectedMultiEntity.setReferences(rList);
+		synapseClient.getEntityHeaderBatch(selectedMultiEntity, new AsyncCallback<PaginatedResults<EntityHeader>>() {
+			@Override
+			public void onSuccess(PaginatedResults<EntityHeader> result) {
+				callback.onSuccess(result);
```
jay-hodgson: Validate based on filter here.

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
jay-hodgson: Please verify that selecting a single Synapse entity (when multi is configured) properly calls back with a list containing the single item.

PR #2744, opened 2017-02-14
SWC-1376: render html file in preview

jay-hodgson:
If in the collaborators team, trust any html.  If not, sanitize (whitelist tags/attributes allowed) - if sanitized version is the same as unsanitized then render.  If sanitation changed html, then show the raw html code instead (never half-render html).

Using html5 srcdoc, and falling back to writing the content if iframe attribute is unsupported (IE/Edge).
Caching (servlet side) membership of collaborators team.  Expires after an hour (will be refreshed on the next call after expiration).

Client now pulls data directly from s3 rather than proxy through the servlet (fixed header to match signature).

Tested on Chrome/FF/Edge/Safari/IE. FF is slow to render compared to others, and I found some example ipynb html export files were not rendered well on IE.


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -555,6 +555,116 @@ private final static native void _showTwitterFeed(String dataWidgetId,
...
+				      return html;
+				    }
+				}
+			};
+			$wnd.xss = new $wnd.filterXSS.FilterXSS(options);
+	}-*/;
+	
+	private final static native String _sanitizeHtml(String html) /*-{
+		try {
+			return $wnd.xss.process(html);
```
jay-hodgson, line 661: what happens if an error occurs during processing?

PR #2746, opened 2017-02-16
SWC-3481:  Allow restricted editing to default file view columns.  Do not block editing of columns with these names in Tables (only File Views)

jay-hodgson:
Locking down column Name, Type, and Size.  See https://sagebionetworks.jira.com/browse/PLFM-4239 for more info.


src/test/java/org/sagebionetworks/web/unitclient/widget/table/modal/fileview/FileViewDefaultColumnsTest.java
```diff
@@ -52,8 +58,6 @@ public void testGetDefaultColumnsWithIds() {
 		fileViewDefaultColumns.getDefaultColumns(isClearIds, mockCallback);
 		verifyNoMoreInteractions(mockSynapseClient);
 		verify(mockCallback, times(2)).onSuccess(columns);
-		
-		verifyZeroInteractions(mockColumn);
```
kimyen: verify that callback.onSuccess()

PR #2753, opened 2017-02-23
Swc 3466 updated

kmaltergott:
Hi Jay, 
Sorry this PR is so long - also, you should delete the older request that is open for 3466.

Thanks,
Karen

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
@@ -35,6 +35,7 @@ public class EntityFinder implements EntityFinderView.Presenter, IsWidget {
 	AuthenticationController authenticationController;
 	ClientCache cache;
 	private SelectedHandler<Reference> selectedHandler;
+	private SelectedHandler<List<Reference>> selectedMultiHandler;
 	private EntityFilter filter;
 	private SynapseAlert synAlert;
 	@Inject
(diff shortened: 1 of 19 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
```diff
@@ -63,82 +65,122 @@ public void configure(EntityFilter filter, boolean showVersions, SelectedHandler
...
+		selectedEntity.addAll(selected);			
+		
 	}
 	
 	@Override
 	public void okClicked() {
 		synAlert.clear();
 		//check for valid selection
-		if (selectedEntity == null || selectedEntity.getTargetId() == null) {
+		if (selectedEntity == null || selectedEntity.get(0).getTargetId() == null) {
```
jay-hodgson: When nothing has been selected, selectedEntity contains one Reference with a null target Id?  I would have thought selectedEntity (which I think should be renamed to selectedEntities) would be empty.

src/main/resources/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderViewImpl.ui.xml
```diff
@@ -46,6 +47,18 @@
 										</b:InputGroupButton>
 									</b:InputGroup>
 								</g:FlowPanel>
+								<g:FlowPanel ui:field="enterMultiIdWidgetContainer">
+									<b:Heading size="H4">
+										<bh:Text>Enter List of Synapse Ids</bh:Text>
+									</b:Heading>
+									<b:InputGroup>
+										<b:TextBox ui:field="synapseMultiIdTextBox" placeholder="Enter Synapse IDs (i.e. 'syn123.4, syn567, syn789.0')"
+											autoComplete="false" />
```
jay-hodgson: Change "syn789.0" to "syn789.1" to show a valid version in the example.

PR #2755, opened 2017-02-23
SWC-3493: getEntityVersions now 0-based index

jay-hodgson:
and java client methods require limit and offset.  when gathering wiki header tree, return all.  same for access requirements (>50 access requirements on an entity/team!)


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/APITableWidget.java
```diff
@@ -323,7 +323,7 @@ public String getPagedURI(String uri) {
 		boolean isNodeQueryService = isNodeQueryService(uri);
 		if (isSubmissionQueryService || isNodeQueryService) {
 			//the node query service's first element is at index 1! (submission query service first element is at index 0)
-			Long firstIndex = isSubmissionQueryService ? ServiceConstants.DEFAULT_PAGINATION_OFFSET : ServiceConstants.DEFAULT_PAGINATION_OFFSET_NO_OFFSET_EQUALS_ONE;
+			Long firstIndex =  ServiceConstants.DEFAULT_PAGINATION_OFFSET;
```
jay-hodgson: @kimyen : please let me know if the node query service index still starts at 1, or has been fixed to start at index 0.

PR #2769, opened 2017-03-03
Swc 3507

kmaltergott:


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinder.java
```diff
@@ -172,7 +172,7 @@ public void lookupEntity(String entityId, final AsyncCallback<List<EntityHeader>
 	}
 
 	private void processEntities(String entityId) {
-		String[] entities = entityId.split(",");
+		String[] entities = entityId.split("[,|\\s]\\s*");
```
jay-hodgson: This also splits on the character '|' (does not act as an OR inside the character class definition.
Please change to `"[,\\s]\\s*"`

PR #2791, opened 2017-03-20
fix data access submission status UI

jay-hodgson:


jay-hodgson:
screen shot 2017-03-20 at 1 46 51 pm

PR #2816, opened 2017-04-11
SWC-3518: from code review

jay-hodgson:


jay-hodgson:
Support case where another person has requested access on your behalf.
access requested by another user

PR #2818, opened 2017-04-13
SWC-3518 code review changes.  Update Download button to use new AR place in alpha mode

jay-hodgson:
Deprecate LicensedDownloader (move direct download logic to the FileDownloadButton).  Use RestrictionInformation to determine if we should use the LicensedDownloader (or direct user to ARs place if in alpha).


jay-hodgson:
Removes support for PostMessage access requirement type, adds support for wiki page ACT access requirement, and (in alpha mode) adds a button to the ACT AR join team page that opens up a new window to the Access Requirements page (where the user can request access using the new workflow).

PR #2828, opened 2017-04-19
SWC-2054

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/upload/ImageUploadWidget.java
```diff
@@ -0,0 +1,138 @@
...
+	public void reset() {
+		view.setInputEnabled(true);
+		view.showProgress(false);
+		synAlert.clear();
+		view.resetForm();
+	}
+	
+	private void doMultipartUpload(final FileMetadata fileMeta, JavaScriptObjectWrapper blob) {
+		// The uploader does the real work
+		multipartUploader.uploadFile(fileMeta.getFileName(), "image/jpeg", blob.get(),
```
jay-hodgson: may not be the correct content type.

PR #2877, opened 2017-05-19
SWC-3612: use AR place for access requirements.

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadButton.java
```diff
@@ -215,7 +199,7 @@ public void onFailure(Throwable caught) {
 	
 	@Override
 	public void onLicensedDownloadClick() {
```
kimyen, line 201: is this method used?

PR #2896, opened 2017-06-05
SWC-3628:  support old ACT Access Requirement and new Managed ACT Access Requirement

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ACTAccessRequirementWidget.java
```diff
@@ -120,60 +108,20 @@ public void onSuccess(String rootWikiId) {
 		createAccessRequirementButton.configure(ar);
 		deleteAccessRequirementButton.configure(ar);
 		revokeUserAccessButton.configure(ar);
```
jay-hodgson, line 110: does this still work?

PR #2950, opened 2017-07-05
SWC-3677: new SelfSignAccessRequirement widget

jay-hodgson:
and change the create AR wizard to create this type (when ACT wants to create a click-wrap).


src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/SelfSignAccessRequirementWidget.java
```diff
@@ -0,0 +1,205 @@
...
+			}
+		});
+		createAccessRequirementButton.configure(ar);
+		deleteAccessRequirementButton.configure(ar);
+		subjectsWidget.configure(ar.getSubjectIds(), true);
+		manageAccessButton.configure(ar);
+		lazyLoadHelper.setIsConfigured();
+	}
+	
+	public void setDataAccessSubmissionStatus(BasicAccessRequirementStatus status) {
```
jay-hodgson, line 117: rename to setStatus()

PR #2979, opened 2017-07-19
SWC-3609: support direct upload/download from client.

jay-hodgson:
(with s3-like external object store)


src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadButton.java
```diff
@@ -198,8 +216,32 @@ public void onFailure(Throwable caught) {
...
+				view.showS3DirectDownloadDialog();
+			}
+		};
+		awsSdk.getS3(accessKeyId, secretAccessKey, objectStoreFileHandle.getBucket(), objectStoreFileHandle.getEndpointUrl(), s3Callback);	
+	}
+	
+	@Override
+	public void onAuthenticatedS3DirectDownloadClicked() {
+		ExternalObjectStoreFileHandle objectStoreFileHandle = (ExternalObjectStoreFileHandle) dataFileHandle;
+		String presignedUrl = awsSdk.getPresignedURL(objectStoreFileHandle.getFileKey(), objectStoreFileHandle.getBucket(), s3);
```
jay-hodgson: Respect "Download file as" value.

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadButton.java
```diff
@@ -198,8 +216,32 @@ public void onFailure(Throwable caught) {
...
+		popupUtilsView.openInNewWindow(presignedUrl);
```
jay-hodgson: @zimingd , is there a way to tell Synapse (for auditing purposes) that the user has downloaded one of these new external object store files?

PR #3044, opened 2017-08-29
SWC-3778: direct calls from js to repo services.

jay-hodgson:
(hop over gwt rpc proxy servlet)


src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -0,0 +1,368 @@
...
+	public void getEntityBundle(String entityId, int partsMask, final AsyncCallback<EntityBundle> callback) {
+		getEntityBundleForVersion(entityId, null, partsMask, callback);
+	}
+
+	public void getEntityBundleForVersion(String entityId, Long versionNumber, int partsMask, final AsyncCallback<EntityBundle> callback) {
+		String url = getRepoServiceUrl() + ENTITY_URI_PATH + "/" + entityId + ENTITY_BUNDLE_PATH + partsMask;
+		if (versionNumber != null) {
+			url += REPO_SUFFIX_VERSION + "/" + versionNumber;
+		}
+		CallbackP<JSONObjectAdapter> constructCallback = new CallbackP<JSONObjectAdapter>() {
```
jay-hodgson, line 208: Keep looking into how we could remove the need for a constructor callback

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -0,0 +1,368 @@
...
+				}
+			};
+			doGet(url, wrapCallback(constructCallback, callback));
+		} else {
+			getVersionOfV2WikiPageAsV1WithWikiPageId(key, versionNumber, callback);
+		}
+	}
+	
+	private void getVersionOfV2WikiPageAsV1WithWikiPageId(WikiPageKey key, Long versionNumber, final AsyncCallback<WikiPage> callback) {
+		String url = getRepoServiceUrl() + "/" +
```
jay-hodgson, line 310: Could this be converted to use v2 api?  (SWC could bundle into v1 object models if necessary)

PR #3046, opened 2017-08-30
SWC-1105: use new type filter

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -130,7 +139,7 @@ public void onResponseReceived(Request request,
 					if (statusCode == SC_OK) {
 						try {
 							JSONObjectAdapter jsonObject = jsonObjectAdapter.createNew(response.getText());
-							callback.onSuccess(jsonObject);
+							callback.onSuccess(jsFactory.newInstance(responseType, jsonObject));
```
jay-hodgson, line 142: Throw checked exception in newInstance if type not recognized.  Then send error to original callback.

PR #3049, opened 2017-09-06
SWC-3483: test that file name meets requirements for entity name

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/upload/FileHandleUploadWidgetImpl.java
```diff
@@ -107,7 +109,15 @@ public void onFileSelected() {
 		if (fileMetaArr != null) {
 			FileMetadata fileMeta = fileMetaArr[0];
 			boolean isValidUpload = validator == null || validator.isValid(fileMeta);
-			if (isValidUpload) {
+			boolean isValidFilename = AbstractFileValidator.isValidFilename(fileMeta.getFileName());
+			if (!isValidFilename) {
+				view.showError(WebConstants.INVALID_ENTITY_NAME_MESSAGE);
+				Callback invalidFileCallback = validator.getInvalidFileCallback();
```
jay-hodgson: what if validator is null here?

PR #3073, opened 2017-09-18
SWC-3783: on delete of external object store file, request "s3" credentials and attempt to delete in object store.

jay-hodgson:


jay-hodgson:
Bad idea, write garbage collector script instead (dependent on new service call to identify if filehandle can be safely swept).

PR #3118, opened 2017-10-06
SWC-3839

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
```diff
@@ -25,7 +25,7 @@
 public class Header implements HeaderView.Presenter, IsWidget {
 
 	public static final String N_A = "n/a";
-	public static final String ANONYMOUS = "anonymous";
+	public static final String ANONYMOUS = "VISITOR_UNIQUE_ID";
 	public static final String SYNAPSE_ORG = "@synapse.org";
 	public static final String GET_SATISFACTION_SUPPORT_SITE = "http://support.sagebase.org";
```
meredithslota, line 30: What is this GET_SATISFACTION_SUPPORT_SITE string for? Is that a thing we have?

meredithslota: cc: @jay-hodgson

PR #3150, opened 2017-10-27
SWC-3862

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/QueryResultEditorWidget.java
```diff
@@ -97,7 +96,6 @@ public void showEditor(QueryResultBundle bundle, TableType tableType, Callback c
 		view.setButtonToolbarVisible(isTable);
 		view.showEditor();
 		this.tableId = QueryBundleUtils.getTableId(bundle);
-		this.etagColumnId = getEtagColumnId();
```
jay-hodgson: add back in (and only clean up if etag column is present)

PR #3157, opened 2017-10-30
move debug parameters to another gwt module

jay-hodgson:
built if you set a property in the maven build settings


hhu94:
Looks good, tested on a local branch and it works.

PR #3188, opened 2017-11-09
SWC-3881

jay-hodgson:


jay-hodgson:
Reviewed/approved by Xa earlier today.

PR #3201, opened 2017-11-17
SWC-3919: push place into browser history…

jay-hodgson:
when an entity is clicked in files/tables/docker


jay-hodgson:
This also fixes an issue with the latest FF (57) when clicking on a folder while scrolled down in the file browser.

PR #3240, opened 2017-12-11
merge release-208 into release-209

jay-hodgson:


jay-hodgson:
verified build (after complex merge)

PR #3251, opened 2017-12-14
SWC-3963: for file download, use fha servlet instead of old file handle servlet

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
```diff
@@ -111,8 +107,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
 			fha.setAssociateObjectId(objectId);
 			fha.setAssociateObjectType(FileHandleAssociateType.valueOf(objectType));
 			fha.setFileHandleId(fileHandleId);
-			URL resolvedUrl = client.getFileURL(fha);
-			response.sendRedirect(resolvedUrl.toString());
+			URL url = client.getFileURL(fha);
+			response.sendRedirect(response.encodeRedirectURL(url.toString()));
```
jay-hodgson: remove encodeRedirectURL (already encoded by the backend service)

PR #3255, opened 2017-12-14
SWC-3965: failure to score password strength should not be logged to …

jay-hodgson:
…cloudwatch


jay-hodgson:
@xschildw reviewed, merging into develop (release-210)

PR #3272, opened 2018-01-09
SWC-3972

jay-hodgson:


jay-hodgson:
Code reviewed with Xa.  Approved so merging.

PR #3273, opened 2018-01-10
SWC-3983: possible that widgets are loading...

jay-hodgson:
before user session data has been reloaded from local storage?


jay-hodgson:
Code reviewed with Xa.  Approved so merging.

PR #3328, opened 2018-02-08
SWC-2367: set visibility of div rather than depending on bootstrap collapse panel

jay-hodgson:


jay-hodgson:
@xschildw reviewed/approved

PR #3357, opened 2018-02-22
SWC-4064

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -405,7 +412,52 @@ public void invoke() {
...
+	
+	public static void confirm(
+			String trustedHtmlMessage,
+			Callback cancelCallback,
+			Callback okCallback
+			) {
+		
+		DialogOptions options = DialogOptions.newOptions(trustedHtmlMessage);
+		options.setCloseButton(false);
+		options.addButton(BUTTON_CANCEL, LINK_BUTTON_STYLE, () -> {
```
jay-hodgson, line 446: double check that if cancel is clicked with no callback the confirmation dialog is hidden.

src/main/resources/org/sagebionetworks/web/client/widget/docker/modal/AddExternalRepoModalViewImpl.ui.xml
```diff
@@ -15,7 +15,7 @@
 		</b:ModalBody>
 		<b:ModalFooter>
 			<b:Button ui:field="saveButton" type="PRIMARY" dataLoadingText="Saving...">Save</b:Button>
```
jay-hodgson, line 17: is cancel button to the left of save in this dialog?

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
jay-hodgson: confirmed

PR #3364, opened 2018-02-27
SWC-4068

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/StuAlertViewImpl.java
```diff
@@ -60,50 +51,68 @@ public void showInfo(String tile, String message) {
...
+		if (widget != null) {
+			httpCode403.setVisible(false);
+			httpCode404.setVisible(false);
+			requestAccessUI.setVisible(false);
+			requestLoadingUI.setVisible(false);
+		}
 	}
 	
+	private void lazyConstruct() {
+		if (widget == null) {
```
jay-hodgson, line 69: jit construction saves ~150KB (stu image downloads)

PR #3369, opened 2018-02-28
SWC-4081: need entity click handler in tree for performance

jay-hodgson:
should not reload project


jay-hodgson:
Reviewed by @xschildw

PR #3383, opened 2018-03-06
SWC-4055

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/WikiDiffPresenter.java
```diff
@@ -0,0 +1,183 @@
...
+			public void onSuccess(WikiPage result) {
+				callback.invoke(result.getMarkdown());
+			}
+		});
+	}
+	
+	@Override
+	public void setPlace(WikiDiff place) {
+		this.place = place;
+		this.view.setPresenter(this);
```
jay-hodgson, line 149: no need to set presenter here

src/main/java/org/sagebionetworks/web/client/presenter/WikiDiffPresenter.java
```diff
@@ -0,0 +1,183 @@
...
+		globalAppState.pushCurrentPlace(place);
+		
+		view.setVersion1(version1);
+		view.setVersion2(version2);
+		
+		// refresh wiki versions available (kick off recursive call to get all versions, 30 at a time)
+		if (wikiVersionHistory == null) {
+			wikiVersionHistory = new ArrayList<>();
+			Long offset = 0L;
+			getVersions(offset);
```
jay-hodgson, line 94: This might be a huge number, we should set a max (thanks Xa!)

PR #3423, opened 2018-03-21
SWC-4124

jay-hodgson:


jay-hodgson:
Docker still has this issue (after the fix) since it is not optimized (does a page load)

PR #3426, opened 2018-03-22
SWC-4116: avoid cors for current stack status request

jay-hodgson:


xschildw:
So it's essentially going back to what we had before? Any idea why it only fails in certain conditions?

jay-hodgson:
Yes, essentially reverting.  Currently unknown.  Based on your description, a request for the stack status that contains a session token while the stack is in RO mode fails preflight.  Hopefully it's specific to this case...

PR #3476, opened 2018-04-25
SWC-4184: if anonymous, then send to the login place…

jay-hodgson:
… since you must first log in to delete or update entity.  Also lazy construct the QuizInfoDialog.


jay-hodgson:

PR #3477, opened 2018-04-26
SWC-4186: store Synapse properties in a js map instead of web storage.

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -127,11 +127,11 @@ public void onSuccess(HashMap<String, String> properties) {
 				for (String key : properties.keySet()) {
```
jay-hodgson: This should not be here, it should use SynapseProperties map.

jay-hodgson:

PR #3492, opened 2018-05-08
Clean up

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/docker/DockerCommitListWidget.java
```diff
@@ -114,7 +111,7 @@ public void onClick(ClickEvent event) {
 							}
 						}
 						offset += LIMIT;
-						commitsContainer.setIsMore(offset < numberOfCommits);
+						commitsContainer.setIsMore(!(result.size() < LIMIT));
```
jay-hodgson, line 114: This looks wrong.  Add test (or fix test if present)

src/main/webapp/Portal.html
```diff
@@ -138,9 +138,6 @@ <h2 class="center" style="color:white; padding-top: 180px">
 			<div id="alertWrapper">
 				<div id="alertContainer"></div>
 			</div>
-			<div id="headerPanel"></div>
-			<div id="rootPanel"></div>
-			<div id="footerPanel"></div>
 			<div id="initialLoadingUI" style="padding-top: 180px">
```
jay-hodgson, line 141: if there's a problem with the initial loading UI being above the rest of the content, let's just remove it from the DOM after load!

PR #3538, opened 2018-05-31
SWC-4240: add special url handling for digitalhealth and mHealth

jay-hodgson:


jay-hodgson:
(do not merge.  change is currently being considered.)

PR #3567, opened 2018-06-15
SWC-4230: scrolling container to the top should be deferred…

jay-hodgson:
… to allow markdown editor to resize to the current viewport


jay-hodgson:
@xschildw  reviewed/approved.

PR #3583, opened 2018-06-21
SWC-4272: change user suggestion oracle delay from 750ms to 250ms

jay-hodgson:


jay-hodgson:
Reviewed by @xschildwachter

PR #3585, opened 2018-06-22
SWC-4278: do not logout if Synapse ToU is unsigned

jay-hodgson:


jay-hodgson:

PR #3616, opened 2018-07-12
SWC-4073

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/LoadMoreWidgetContainer.java
```diff
@@ -97,5 +97,6 @@ public boolean remove(Widget w) {
 	 */
 	public void setIsProcessing(boolean isProcessing) {
 		this.isProcessing = isProcessing;
+		view.setLoadMoreVisibility(true);
```
xschildw, line 100: LoadMoreVisibility should be set to isProcessing

PR #3624, opened 2018-07-13
SWC-4315: update center text md plugin to use version that outputs css class

jay-hodgson:
instead of style


jay-hodgson:
@xschildw  reviewed/approved

PR #3634, opened 2018-08-06
SWC-4227

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/team/InviteWidget.java
```diff
@@ -69,69 +114,40 @@ public void setRefreshCallback(Callback teamUpdatedCallback) {
...
+		this.invitationMessage = invitationMessage;
+		doSendInvites();
+	}
+	
+	/**
+	 * Recursively process the user invitations (emails, then Synapse users).
+	 */
+	public void doSendInvites() {
+		currentlyProcessingUser = currentlyProcessingEmail = null;
+		if (!inviteEmails.isEmpty()) {
```
jay-hodgson, line 131: should be called inviteEmailAddresses

PR #3658, opened 2018-09-08
SWC-4319

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/search/SearchBoxViewImpl.java
```diff
@@ -19,33 +18,47 @@
...
 			public void onKeyDown(KeyDownEvent event) {
 				if (event.getNativeEvent().getKeyCode() == KeyCodes.KEY_ENTER) {
-					searchButton.click();
+					executeSearch();
 	            }					
 			}
-		});				
+		});
+	    searchField.addBlurHandler(event -> {
+	    	searchFieldInactive();
```
jay-hodgson, line 44: clicking on the search icon while active should execute the search.

jay-hodgson: could we delay making inactive for a second?

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TablePageViewImpl.java
```diff
@@ -58,6 +59,10 @@
 	Div topScrollDiv;
 	@UiField
 	Div tableDiv;
+	
+	public static final String ACTIVE_FACET_WIDGET_STYLES = "pull-left-unless-xs margin-right-10 panel panel-default padding-10 facetsWidget-active";
```
jay-hodgson, line 63: hiding and then showing facets is broken.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TablePageWidget.java
```diff
@@ -47,7 +47,7 @@
 	TableType tableType;
```
jay-hodgson, line 47: back out these changes to fix facets

PR #3684, opened 2018-09-25
SWC-4283

jay-hodgson:


jay-hodgson:
(Not to be pulled in until branch code is evaluated on a parallel portal stack.)

PR #3696, opened 2018-10-01
remove all uses of old DOI widget and bring new one out of alpha mode

nickgros-sage:
🚫🚫🚫DO NOT MERGE until we begin DOI migration 🚫🚫🚫

delete the old DOI widget later


jay-hodgson:
You can also remove the v1 DOIWidget (and view and view impl) in this PR (deleting code is the best!)

nickgros:
fixed merge issues + build is clean

nickgros:
this now includes changes from #3748 (so it can be merged on top of them easily)

PR #3722, opened 2018-10-12
bump portal css filename version

jay-hodgson:
since production is also using Portal10.css.  looking forward to SWC-4283 branch, where we don't have to deal with this anymore!


jay-hodgson:
reviewed/approved by @xschildw

PR #3754, opened 2018-11-02
pull new doi widget out of alpha and remove old code

nickgros:


jay-hodgson:
** Not to be merged until next week (when release-244 is production and we're ready to patch for doi migration).  **  

Nice work @nickgros

PR #3770, opened 2018-11-13
SWC-4197: add additional UI to support Not Set and All

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/facets/FacetColumnResultRangeWidget.java
```diff
@@ -79,14 +95,20 @@ public Double getNewMax() {
...
-			facetColumnRangeRequest.setMin(Double.toString(newMin));	
-		}
-		Double newMax = getNewMax();
-		if (newMax != null) {
-			facetColumnRangeRequest.setMax(Double.toString(newMax));
+		if (view.isNotSet()) {
+			facetColumnRangeRequest.setMin(TableConstants.NULL_VALUE_KEYWORD);
+			facetColumnRangeRequest.setMax(TableConstants.NULL_VALUE_KEYWORD);
+		} else if (!view.isAnyValue()) {
+			Double newMin = getNewMin();
```
jay-hodgson: Just get the String value from the view!

PR #3866, opened 2019-01-29
PORTALS-367: allow any requested headers if on synapse.org subdomain

jay-hodgson:


jay-hodgson:
issue is in the SRC call, not with the servlet response.

PR #3918, opened 2019-02-20
SWC-4479

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/LoadMoreWidgetContainer.java
```diff
@@ -91,12 +68,9 @@ public void clear() {
...
 	}
-
-	/*
-	 * for test only
-	 */
-	public void setIsProcessing(boolean isProcessing) {
-		this.isProcessing = isProcessing;
-		view.setLoadMoreVisibility(true);
+	
+	public void setIsProcessing(boolean isProcessing)
```
jay-hodgson: missing code, should update view with isProcessing()

PR #3939, opened 2019-03-06
SWC-4234

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -156,6 +157,7 @@ public void onSuccess(EntityBundle bundle) {
 						view.showErrorMessage(DisplayConstants.ERROR_NO_LINK_DEFINED);
 					}
 				}
+				entityId2BundleCache.put(bundle.getEntity().getId(), bundle);
```
xschildw: On line 185, use the cache call.

PR #3954, opened 2019-03-13
SWC-3981: use KaTeX

jay-hodgson:


src/main/webapp/Portal.html
```diff
@@ -94,6 +94,7 @@
 				head.load(cdnEndpoint + "js/linkify.min.js");
 				head.load(cdnEndpoint + "js/linkify-string.min.js");
 				head.load(cdnEndpoint + "js/SRC/synapse-react-client.production.styles.css");
+				head.load("//cdn.jsdelivr.net/npm/katex@0.10.1/dist/katex.min.css");
```
jay-hodgson: move this to Synapse cdn

PR #3958, opened 2019-03-19
SWC-3334: use Synapse Table widgets (sortable column headers and cell renderers)

jay-hodgson:


jay-hodgson:
Remove option to show "Row Numbers"

PR #3974, opened 2019-03-26
null checks in view

jay-hodgson:


jay-hodgson:
merging to get the build started...

PR #4029, opened 2019-04-29
SWC-4849: place tooltips on the right to give more space

jay-hodgson:


jay-hodgson:

PR #4030, opened 2019-04-30
SWC-4851: find the column config column in the response column set, a…

jay-hodgson:
…nd use the corresponding row values.


jay-hodgson:

PR #4051, opened 2019-05-16
SWC-4873: prevent lastpass autofill icon (sticks around in SPA)

jay-hodgson:


xschildw (approved):
So that's disabling LP for our user/pwd input fields?

PR #4056, opened 2019-05-21
PORTALS-494: leave portals config cookie until...

jay-hodgson:
keep showing the bumper bar until user makes it out of a transition page (like registration, or login, or password reset)


jay-hodgson:

PR #4077, opened 2019-07-02
SWC-4915: special case pdf.js iframe load, and add a test.

jay-hodgson:


jay-hodgson:
Merging change into release-271 (staging) branch, and will merge down to develop after code review (and any necessary changes)

PR #4087, opened 2019-07-17
SWC-4920: update aws js sdk library

jay-hodgson:


jay-hodgson:
Approved by @xschildw

PR #4096, opened 2019-07-23
SWC-4933, and remove head.js

jay-hodgson:
and update dependencies: pico, jsplumb, ...


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -769,7 +769,11 @@ public void unmountComponentAtNode(Element el) {
 		_unmountComponentAtNode(el);
 	}
 	public final static native void _unmountComponentAtNode(Element el) /*-{
-		return $wnd.ReactDOM.unmountComponentAtNode(el);
+		try {
```
jay-hodgson, line 772: This should fix SWC-4930 (it's the only instance where window.ReactDOM is referenced outside of a try/catch block).

jay-hodgson:
still in progress

xschildw (approved):

PR #4102, opened 2019-08-02
SWC-4933

jay-hodgson:
New approach.  Revert portal.html changes.  Reintroduce dependency version updates, and bake in head.js load to avoid additional request.


xschildw (approved):
OK

PR #4120, opened 2019-08-23
SWC-4951: 403 response may return after stu alert was cleared.  make …

jay-hodgson:
…sure it is clear if call is successful


src/main/java/org/sagebionetworks/web/client/widget/entity/WikiPageWidget.java
```diff
@@ -97,7 +97,6 @@ public WikiPageWidget(WikiPageWidgetView view,
 		this.historyWidget = historyWidget;
 		this.markdownWidget = markdownWidget;
 		this.wikiSubpages = wikiSubpages;
-		this.stuAlert = stuAlert;
```
jay-hodgson, line 100: Removed unnecessary code - this assignment already exists on line 96.

xschildw (approved):

PR #4123, opened 2019-08-28
SWC-4955: keep track of requests, and add ability to cancel pending r…

jay-hodgson:
…equests


jay-hodgson:
replacing with patch to staging branch: https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4124

PR #4145, opened 2019-09-11
PORTALS-678: also remove ':' from Pendo sanitized url

jay-hodgson:


jay-hodgson:
Screen Shot 2019-09-10 at 5 04 04 PM

PR #4148, opened 2019-09-13
SWC-4928

jay-hodgson:


jay-hodgson:
@xschildwachter reviewed/approved.

PR #4255, opened 2020-01-16
SWC-5082: clear last place on explicit logout.  clear session storage on genera…

jay-hodgson:
…l logout (save last place for UX).


jay-hodgson:
@xschildw has reviewed and approved the changes after going through them.

xschildw (approved):

PR #4271, opened 2020-02-06
SWC-5104: show maintenance message popup for 5 seconds

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/SynapseStatusDetector.java
```diff
@@ -14,6 +14,9 @@
 	DateTimeFormat iso8601DateFormat = DateTimeFormat.getFormat(PredefinedFormat.ISO_8601);
 	DateTimeUtils dateTimeUtils;
 
+	// how long the popup should be shown (in ms)
+	public static final int POPUP_DELAY_MS = 5000;
```
tschaffter, line 18: @jay-hodgson It would be more intuitive to define the delay in a "configuration" and have all Synapse popups refer to this configuration. If needed, a `POPUP_SHORT_DELAY_MS` and `POPUP_LONG_DELAY_MS` may be defined to better support popup with messages that takes a short/longer time to be read. Defining these values at a single place will avoid discrepancies in the behavior of the popups.

jay-hodgson: @tschaffter : This is a good idea.  Currently, the maintenance popup is the **only** case where an information popup automatically disappears, so I think we can wait until n=2 to refactor.

tschaffter: @jay-hodgson I agree if this is the first time a delay is introduced. But in Jira you said "Other popup messages already have a “delay” of nonzero (so they disappear)."

tschaffter: @jay-hodgson An example of existing popup with delay that you already have is the "You are not logged out".

PR #4275, opened 2020-02-11
SWC-5106

jay-hodgson:


jay-hodgson:
@xschildw reviewed/approved.  Merging

PR #4276, opened 2020-02-13
SWC-5111: show video if widget can access all video FileHandle info (…

jay-hodgson:
…for every video format that user wants to support)

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/VideoWidget.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/VideoWidget.java
@@ -8,6 +8,7 @@ import java.util.List;
 import java.util.Map;
 import org.sagebionetworks.repo.model.entitybundle.v2.EntityBundle;
 import org.sagebionetworks.repo.model.entitybundle.v2.EntityBundleRequest;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/VideoWidget.java
```diff
@@ -53,15 +63,48 @@ public void configure(String synapseId, String filename, int width, int height)
...
 	}
-
+	
 	private void configureFromSynapseFile(String mp4SynapseId, String oggSynapseId, String webmSynapseId, String width, String height) {
+		// Must be logged in in order to download the Synapse video file, unless it's open data.
 		if (!authenticationController.isLoggedIn()) {
-			// not logged in and attempting to download a Synapse video file.
-			view.showError(PLEASE_LOGIN_TO_VIEW_THIS_RESOURCE);
+			// Try to get the file handle information.  if successful, then show the video.
+			renderIfCanAccessAllFileHandles(mp4SynapseId, oggSynapseId, webmSynapseId, width, height);
```
xschildw: call unconditionally

xschildw (approved):

PR #4282, opened 2020-02-22
SWC-5087: fail early (as soon as the last modified on date changes of…

jay-hodgson:
… the file, fail)


src/main/java/org/sagebionetworks/web/client/widget/upload/MultipartUploaderImpl.java
```diff
@@ -177,6 +181,11 @@ public void attemptToUploadNextPart() {
 		// for each chunk that still needs to be uploaded, get the presigned url and upload to it
 		currentPartNumber++;
 
+		// sanity check - if last modified has changed, then fail.
+		long lastModifiedOn = synapseJsniUtils.getLastModified(blob);
+		if (lastModifiedOn != fileLastModifiedMs) {
+			uploadFailedDueToFileModification(new Date(lastModifiedOn));
```
xschildw, line 187: return here

PR #4286, opened 2020-02-27
SWC-5127: make synapseProperties map an instance variable

jay-hodgson:
and don't clear() the map on success/failure.


jay-hodgson:
@xschildw reviewed and approved.

PR #4299, opened 2020-03-06
SWC-5141: extra clear of replies container if offset is 0

jay-hodgson:
guard against previous execution that was not canceled


jay-hodgson:
@xschildw approved.

PR #4300, opened 2020-03-11
SWC-5146: ignore gwt Tree gain focus event

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityTree.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.google.gwt.user.client.ui.Tree;
+
+public class EntityTree extends Tree {
+	
+	public EntityTree(Resources resources) {
+		super(resources);
+	}
+  @Override
+  public void setFocus(boolean focus) {
```
jay-hodgson, line 11: This is the fix.

jay-hodgson:
Walked through with @xschildw , and he approved changes with demo.

PR #4304, opened 2020-03-13
SWC-5047: updated tablesorter lib with working sortEnd event, but UI …

jay-hodgson:
…does not update (concurrent style update?)


jay-hodgson:
Now that we have SubmissionViews, they should be used for Leaderboards (which makes this PR obsolete).

PR #4311, opened 2020-03-24
SWC-5160

jay-hodgson:


jay-hodgson:
- Rename hideButtons() to hideControls() (note that hideControls is only called by this Change User Access ACT dialog).
- Add ability to hideControls() to Self Sign Access Requirement widgets.  hideControls() sets the entire controlsContainer visibility to false.
- Instead of showing the text from the Access Requirement in ApproveUserAccessModal, use an AccessRequirementWidget (which handles wiki rendering as well).
- Modify tests to verify AccessRequirementWidget is configured with selected Access Requirement ID (and target subject).

xschildw (approved):

PR #4330, opened 2020-04-13
SWC-5170: check for plus sign up front

jay-hodgson:


jay-hodgson:
on second thought, new file name constraints should be introduced in the back end.

PR #4342, opened 2020-04-29
SWC-5184

jay-hodgson:


jay-hodgson:
@xschildw reviewed and approved.  Merging.

PR #4376, opened 2020-06-30
SWC-5240 (b): Additionally, redirect for ad knowledge portal too.

jay-hodgson:


jay-hodgson:
Did not get approval, closing.

PR #4405, opened 2020-07-29
SWC-5260: fix dropdown caret in signed-in UI

jay-hodgson:


jay-hodgson:
After this change (verified on Chrome and FF):
Screen Shot 2020-07-29 at 9 47 11 AM

zimingd (approved):

PR #4429, opened 2020-09-02
SWC-5288: set img source to the preview repo endpoint if showing a small user badge

jay-hodgson:


jay-hodgson:
Hold on.  If image is unavailable, then we don't handle it properly (default letter UI)..

xschildw (approved):

PR #4442, opened 2020-10-21
update SRC, and make sure node libraries know we're running in production

jay-hodgson:


xschildw (approved):
Why can't we use the logic that's just following the change where we set the cdnEndpoint
(assuming you need process.env.NODE_ENV=='development' at some point)?

xschildw (approved):

PR #4453, opened 2020-11-17
SWC-5723

nickgros:
* Add place for personal access token management
* Will need a new release of SRC before this feature is ready

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/view/PersonalAccessTokensViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/view/PersonalAccessTokensViewImpl.java
@@ -47,6 +47,7 @@ public class PersonalAccessTokensViewImpl extends Composite implements PersonalA
 
 	@Override
 	public void render() {
+		jsniUtils.unmountComponentAtNode(container.getElement());
 		_showPersonalAccessTokensComponent(container.getElement(), authController.getCurrentUserSessionToken());
 	}
 
@@ -72,28 +73,5 @@ public class PersonalAccessTokensViewImpl extends Composite implements PersonalA
 	public void setPresenter(PersonalAccessTokensPresenter presenter) {
 		this.presenter = presenter;
 		headerWidget.refresh();
-		render();
 	}
-
-
-	@Override
-	public void showErrorMessage(String message) {
-		DisplayUtils.showErrorMessage(message);
-	}
-
-	@Override
-	public void showLoading() {}
-
-	@Override
-	public void showInfo(String message) {
-		DisplayUtils.showInfo(message);
-	}
-
-
-	@Override
-	public void clear() {
-		jsniUtils.unmountComponentAtNode(container.getElement());
-		render();
-	}
-
 }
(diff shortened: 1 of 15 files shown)


src/main/java/org/sagebionetworks/web/client/view/PersonalAccessTokensViewImpl.java
```diff
@@ -0,0 +1,99 @@
...
+	}-*/;
+
+
+
+
+	@Override
+	public void setPresenter(PersonalAccessTokensPresenter presenter) {
+		this.presenter = presenter;
+		headerWidget.refresh();
+		render();
```
jay-hodgson: BulkPresenterProxy calls setPlace (which calls setPresenter(), which calls render()) and then start (which calls clear() which calls unmount() and then render()).  I think we have too many renders here.  Could you remove render() from view.setPresenter?  And then call view.render() in presenter.start() (instead of clear()), and change render() attempt to unmount before _showPersonalAccessTokensComponent()?  then you could delete clear() as well.

src/test/java/org/sagebionetworks/web/client/presenter/PersonalAccessTokensPresenterTest.java
```diff
@@ -0,0 +1,76 @@
...
+        // Method under test
+        presenter.setPlace(place);
+
+        // Called once in the constructor and once on setPlace invocation
+        verify(mockView, times(2)).setPresenter(presenter);
+    }
+
+    @Test
+    public void testStart() {
+        GWTMockUtilities.disarm();
```
jay-hodgson, line 50: Disarming due to the "new Widget()"?  If that's the reason, you could instead have the when(mockView.asWidget()) return a mock IsWidget.class to avoid this issue.  Then verify verify(mockPanel).setWidget is called with this mock object.

src/main/java/org/sagebionetworks/web/client/view/PersonalAccessTokensViewImpl.java
nickgros: I can't remove `clear` because PersonalAccessTokensView implements the SynapseView interface. Is it necessary to implement SynapseView? If so, I can just make `clear` an empty method.

src/test/java/org/sagebionetworks/web/client/presenter/PersonalAccessTokensPresenterTest.java
nickgros: `asWidget` must return a `Widget` (not an `IsWidget`). I can use `mock(Widget.class)` instead of `new Widget()`, but that still requires calling disarm.

src/main/java/org/sagebionetworks/web/client/view/PersonalAccessTokensViewImpl.java
jay-hodgson: Implementing SynapseView is not a requirement.

src/test/java/org/sagebionetworks/web/client/presenter/PersonalAccessTokensPresenterTest.java
jay-hodgson: That's right.  We must not be verifying the actual object in other tests.  Carry on 👍

jay-hodgson (approved):

PR #4454, opened 2020-11-18
load custom bootstrap theme after SRC css

nickgros:
This issue is supposed to resolve an issue caused by this https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/770; this need not be merged if the other PR is rejected.


nickgros:
The only issue that I actually saw (without this change) was that the alert text (e.g. at the top of the page when not on production SWC) was changed from white (as defined in the custom Bootstrap theme) to yellow (the default bootstrap alert text color, which would come from SRC if it pulls in its own Bootstrap configuration). Other issues may exist.

PR #4455, opened 2020-11-20
PORTALS-1760

jay-hodgson:
1.  Refactor: delete UserProfileModalWidget and move logic to UserProfileEditorWidget.
2.  Use js client to update UserProfile
3.  Restyle, make editor also the renderer (same form, but restyle readonly input elements)
4.  Integrate SRC AccountLevelBadge
Also noticed that the userprofile preview image call was happening a lot.  Setting the src directly to the backend service call (with redirect=true) significantly reduces the call count (in Chrome at least).  Doing it this way also speeds up user badge load, since we remove one rpc (to get the presigned URL) before configuring the UserCard.


jay-hodgson:
Make it so "Change email" and "Change password" open a new window to Settings.

src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
```diff
@@ -43,26 +55,70 @@
...
+		saveProfileButton.addClickHandler(event -> {
+			saveProfileButton.state().loading();
+			presenter.onSave();
+		});
+		cancelButton.addClickHandler(event -> {
+			presenter.onCancel();
+		});
+		linkRenderer.getElement().setAttribute("rel", "noreferrer noopener");
+		ClickHandler goToSettingsPage = event -> {
+			globalAppState.getPlaceChanger().goTo(new Profile(authController.getCurrentUserPrincipalId(), ProfileArea.SETTINGS));
```
jay-hodgson: open window instead

src/main/webapp/sass/swc.scss
```diff
@@ -128,12 +128,14 @@ end sticky footer
 		padding: 15px 0 30px 0;
 	}
 	.nav-tabs {
```
jay-hodgson, line 130: Also uppercase nav tab anchor text

jay-hodgson:
Pass in ORCID, show link if set.
Pass user profile owner ID to image widget, so it can be shown for other users.
More fixes to tab styles.

xschildw (approved):

PR #4458, opened 2020-11-25
SWC-5322: change registration page UI to be more like design

jay-hodgson:
Use a SynapseAlert for error handling, and rework the UI so it looks like the new design.


jay-hodgson:
Including these commits in the pledge and page progress integration PR:  https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4459

PR #4459, opened 2020-11-25
SWC-5323

jay-hodgson:


jay-hodgson:
Screenshots can be found in https://sagebionetworks.jira.com/browse/SWC-5322 (look for the latest comment that has screenshots)

PR #4464, opened 2020-12-04
SWC-5330, and update SRC to get the full Synapse terms/conditions copy

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
```diff
@@ -37,21 +38,24 @@
 import com.google.inject.Inject;
 
 public class LoginPresenter extends AbstractActivity implements LoginView.Presenter, Presenter<LoginPlace> {
+	public static final String CANCEL_TERMS_OF_USE_CONFIRM_MESSAGE = "Canceling now will log you out. You can always log in again to resume the registration process from where you left off.";
```
xschildw, line 41: Technically the user is logged in but he does not know that (the Login button is on). How about something like 'You will need to login in using the credentials you just created to resume the registration process.'?

xschildw (approved):
Per discussion with Jay, merging now to test func. We can fix text later if need be.

PR #4486, opened 2021-01-07
SWC-5364

nickgros:
Our custom checkboxes and radio buttons would wrap in such a way that label text would end up below the button, instead of with a 'hanging indent' style. 

This was fixed by converting the checkbox/radio div wrappers to flex containers, preventing flex wrap, and disabling shrink on the buttons. Some other tweaks were added to ensure the "selected" icons still lined up correctly. Checked on FF/Chrome/Safari/Edge

The issue is easiest to see here:

Before:
image

After:
image


jay-hodgson:
Nice!

PR #4499, opened 2021-01-19
SWC-5313: from review

jay-hodgson:


xschildw (approved):
Wondering if we should remove references to 'alpha' in general? I the target audience is internal I'd assume users know what we're talking about but external might not. Should we talk about 'early access' (like in games)?
Something like 'This mode gives you early access to features that are still in development. Note that we do not guarantee the absence of errors, that data created using the features may be lost and that the feature itself could change significantly (or even be removed).'

src/main/resources/org/sagebionetworks/web/client/widget/footer/FooterViewImpl.ui.xml
```diff
@@ -34,17 +34,17 @@
...
-									<span>Alpha Mode <strong>Off</strong></span>
+									<span>Experimental Mode <strong>Off</strong></span>
 								</b:Anchor>
 								<b:Anchor ui:field="debugOffLink"
 									addStyleNames="flexcontainer-column flexcontainer-justify-flex-end color-white">
-									<span>Alpha Mode <strong>On</strong></span>
+									<span>Experimental Mode <strong>On</strong></span>
 								</b:Anchor>
 								<w:HelpWidget
 									helpMarkdown="This mode is for alpha testing features only. Please note that the developer does not guarantee an absence of errors, and that the data created using alpha features may be lost during product upgrade."
```
xschildw, line 50: This ref to alpha also needs to change?

PR #4500, opened 2021-01-20
SWC-5378 

nickgros:
Add version dropdown and some informative copy text to DOI modal. This also fixes an issue where the client could try to create a versioned DOI for the current version of the table, which ends up creating a broken DOI.

Opening the DOI modal on a versionable entity:
image

Options on versionable entities: 
image

The help text:
image

When you open a non-versionable entity, you don't see the version section:

image

On versionable entities, the DOI information reloads when switching versions, so the user will always know if they will be overwriting data beforehand.


jay-hodgson:
Has this help text been reviewed by @stacey-w-taylor ?  
For example, I think the UI and user docs have historically tried to steer clear of the word "Entity", since this was a difficult abstraction to explain/grasp.

jay-hodgson:
> Has this help text been reviewed by @stacey-w-taylor ?
> For example, I think the UI and user docs have historically tried to steer clear of the word "Entity", since this was a difficult abstraction to explain/grasp.

I suppose we started doing this long ago for Entity Views (since the entity type utils display name returns Entity View), so this is unrelated to the current PR.

src/main/resources/org/sagebionetworks/web/client/widget/doi/CreateOrUpdateDoiModalViewImpl.ui.xml
```diff
@@ -20,6 +20,16 @@
 					addStyleNames="em" />
 				<bh:Div ui:field="doiEditorDiv">
 					<b:Form>
+						<b:FormGroup ui:field="versionForm">
+							<b:FormLabel>Version</b:FormLabel>
+							<w:HelpWidget
+									ui:field="versionHelpBox"
+									helpMarkdown="The version that the DOI should be minted for. Unversioned DOIs will always resolve to the newest version of the "
```
jay-hodgson: Recommend deleting the tail "of the " and replace with a ".".  Just in case it's shown in some unexpected edge case.

jay-hodgson (approved):

stacey-w-taylor:
@jay-hodgson Thanks for the heads up.  I think we should move away from the "Entity View" terminology in the UI and docs, but I agree with you that it's a bigger issue than this PR.

@nickgros I have a couple of copy edits for the text in these screenshots:

First line of the DOI Modal is: "Minting a DOI allows you to credit contributors to your work, and makes it easy to cite your work"

I suggest changing this line to: "Minting a DOI allows you to credit contributors and makes it easy to cite your work"

For the help text:

In general, this seems like more text/info than we typically put in a help text pop up. Should some of this move to the docs (with a link as in the Schema help text)?

"The version of the Entity View for which the DOI should be minted" -- I suggest changing to "A DOI can be associated with a specific version of an Entity View."

For the rest of the text, I'm not sure that some of our users will understand what "resolve" means in this context.  Either we should think about how to explain this differently/succinctly, or move this to the docs where we can have more room for context.

nickgros:
> I'm not sure that some of our users will understand what "resolve" means in this context

I think we can convey the same idea by replacing "resolve" with "link"

In general, I agree that this section should probably link to docs. I can look into updating the DOI docs page with some more thorough information about how versioning interacts with DOIs.

stacey-w-taylor:
> > I'm not sure that some of our users will understand what "resolve" means in this context
> 
> I think we can convey the same idea by replacing "resolve" with "link"
> 
> In general, I agree that this section should probably link to docs. I can look into updating the DOI docs page with some more thorough information about how versioning interacts with DOIs.

Thanks, Nick -- "link" instead of "resolve" is more intuitive, so thumbs up from me.

PR #4505, opened 2021-01-27
bump SRC dependency version and take new Evaluation Editors out of alpha test mode (SWC-5393)

zimingd:


src/test/java/org/sagebionetworks/web/unitclient/widget/evaluation/AdministerEvaluationsListTest.java
```diff
@@ -75,8 +76,11 @@ public void setup() throws Exception {
 		e1.setId("101");
 		e1.setCreatedOn(new Date());
 		evaluationResults.add(e1);
+		e1.setQuota(new SubmissionQuota());
```
zimingd, line 79: this tests interactions with the old evaluation editor

zimingd:
Taking it out of the test mode is still backwards compatible with the SubmissionQuota system, but ensures any newly created Evaluations will use the EvaluationRounds system

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTab.java
```diff
@@ -119,15 +119,9 @@ public void showCreateNewEvaluationEditor(String entityId){
...
-			actionMenuWidget.setActionListener(Action.ADD_EVALUATION_QUEUE, (Action action) -> {
-				showOldAddEvaluationQueueModal();
-			});
-		}
+		actionMenuWidget.setActionListener(Action.ADD_EVALUATION_QUEUE, (Action action) -> {
+			showCreateNewEvaluationEditor(entityId);
+		});
 	}
 
 	private void showOldAddEvaluationQueueModal() {
```
jay-hodgson, line 127: Is showOldAddEvaluationQueueModal() dead code after this change?

zimingd: No, it is still being used to be backwards compatible with Evaluations that have the old SubmissionQuotas.
I did a code review with @marcomarasca for the backend changes, and he suggested a better plan to allow users to opt-in to the conversion via a REST API call instead of a mass migration. 
On the UI side, I will need to add a button on the old evaluation editor to call that API, but given that it'll be thrown away eventually, it probably doesn't need to look pretty.

jay-hodgson (approved):

PR #4507, opened 2021-02-01
SWC-5375 - Add node, yarn, SCSS compiling to the build

nickgros:
This PR allows us to integrate SRC SCSS into the build step, which allows us to override SRC Sass variables. This is accomplished by pulling SRC (and the Sass compiler) from NPM at build-time.

Additionally:
* Moved styles in swc.scss to _core.scss
* swc.scss now imports other SCSS/CSS files such as core, SRC's styles, and our custom bootstrap 3 theme
* Removed the old Sass compilation step


package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "sass": "1.32.5",
+    "synapse-react-client": "1.15.64"
```
nickgros: This is where the SRC version will be managed from now on

Will need to release a new version of SRC after merging https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/883/files and bump here

pom.xml
```diff
@@ -287,26 +287,73 @@
 					<port>8989</port>
 				</configuration>
 			</plugin>
+
 			<plugin>
-				<groupId>com.github.warmuuh</groupId>
-				<artifactId>libsass-maven-plugin</artifactId>
-				<version>0.2.9-libsass_3.4.7</version>
+				<groupId>com.github.eirslett</groupId>
+				<artifactId>frontend-maven-plugin</artifactId>
```
nickgros, line 293: Pretty handy plugin to include Node/NPM/Yarn in the build process: https://github.com/eirslett/frontend-maven-plugin

package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "sass": "1.32.5",
+    "synapse-react-client": "1.15.65"
```
nickgros, line 4: This is where the SRC version will be managed from now on

nickgros:
TODO After this PR, update the developer bootstrap doc

The Sass section will be outdated. Can also include a section on using yarn link to use local SRC (just like in Portals)

jay-hodgson (approved):
I really like this direction.  
In theory, I think we could extend this pattern to many other dependencies referenced in Portal.html.  Perhaps then we could rely on GitHub to alert us when there's a critical update (since it will be looking at our package.json file from now on).

PR #4513, opened 2021-02-05
Fix build; consolidate compiled files to generated folder

nickgros:


jay-hodgson (approved):
Checked out nickgros:fix-build and verified that I could spin up a local synapse.org client (that pulls in the generated css and SRC js file).

zimingd:
can confirm `mvn gwt:run` works. Checking IDE

PR #4515, opened 2021-02-08
SWC-5342: New login page

emmanmills:
Apply new design to Synapse login page

Screen Shot 2021-02-08 at 2 17 01 PM


jay-hodgson (approved):
Looks great.  Just need to publish a new SRC and pull in that new version into SWC for this to work.  Thanks Emma!

PR #4523, opened 2021-02-25
SWC-5398 - Use new Entity Finder built in React

nickgros:


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2Impl.java
```diff
@@ -0,0 +1,348 @@
...
+        private boolean multiSelect = false;
+        private String initialContainerId = null;
+
+        private String modalTitle = "Find in Synapse";
+        private String promptCopy = "";
+        private String selectedCopy = "Selected";
+        private String confirmButtonCopy = "Select";
+
+        @Inject
+        public Builder(EntityFinderV2View view, GlobalApplicationState globalApplicationState, AuthenticationController authenticationController, SynapseAlert synAlert, SynapseJavascriptClient jsClient) {
```
jay-hodgson: An alternative approach, that would avoid the dependency injection mess, would be to create a builder for an EntityFinderConfiguration object instead (then I think you would have one configure(EntityFinderConfiguration) method to rule them all).  Then you could pull the builder into it's own class, and not require injection of the builder (just the EntityFinderV2).

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentDiv.java
```diff
@@ -12,4 +12,10 @@ protected void onUnload() {
 		ReactDOM.unmountComponentAtNode(this.getElement());
 		super.onUnload();
 	}
+
+	@Override
+	public void clear() {
```
nickgros, line 17: Provides a public method to unmount the component

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentDiv.java
```diff
@@ -12,4 +12,10 @@ protected void onUnload() {
...
+		ReactDOM.unmountComponentAtNode(this.getElement());
+		super.clear();
+	}
```
nickgros, line 20: Public method to unmount the component

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFilter.java
```diff
@@ -5,17 +5,29 @@
...
+	ALL(EntityType.values()),
+	ALL_DIRECTORY(project, folder, file, link),
+	CONTAINER(project, folder),
+	PROJECT(project),
+	FOLDER(folder),
+	FILE(file),
+	TABLE(table),
+	ALL_DIRECTORY_BUT_LINK(project, folder, file),
+	ALL_TABLES(table, entityview, submissionview),
+	PROJECT_OR_TABLE(project, table, entityview, submissionview);
```
nickgros: Changing the meaning of "ALL" here to really mean all entity types. Open to naming suggestion for "ALL_DIRECTORY" which is now what "ALL" was

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2Impl.java
```diff
@@ -0,0 +1,332 @@
...
+        @Override
+        public EntityFinder.Builder setInitialContainer(InitialContainer initialContainer) {
+            this.initialContainer = initialContainer;
+            return this;
+        }
+
+        @Override
+        public EntityFinder.Builder setSelectedHandler(SelectedHandler<Reference> handler) {
+            if (multiSelect) {
+                throw new IllegalArgumentException("Attempted to set handler for single selection in Entity Finder when multiselect is true");
```
nickgros: This exception and others are for the developer because it indicates a misconfiguration. A user should never see it. Is there a more appropriate exception type?

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1203,22 +1203,30 @@ private void onLink() {
...
+					createLink(selected.getTargetId(), entityFinder);
+				})
+				.setTreeOnly(true)
+				.setModalTitle("Create Link to " + entityTypeDisplay)
+				.setHelpMarkdown("Search or Browse to find a Project or Folder that you have access to, and place a symbolic link for easy access")
+				.setPromptCopy("Find a destination and place a link to " + entity.getId())
+				.setSelectedCopy("Destination")
+				.setConfirmButtonCopy("Create Link")
+				.build()
+				.show();
```
nickgros: Most of the changed files are configuring the new entity finder with specified copy

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/ProvenanceEditorWidget.java
```diff
@@ -31,26 +32,22 @@
...
 		this.synAlert = synAlert;
 		this.ginInjector = ginInjector;
-		this.entityFinder = entityFinder;
 		this.urlDialog = urlDialog;
 		this.eventBus = eventBus;
 		usedProvenanceList = ginInjector.getProvenanceListWidget();
 		executedProvenanceList = ginInjector.getProvenanceListWidget();
-		usedProvenanceList.setEntityFinder(entityFinder);
 		usedProvenanceList.setURLDialog(urlDialog);
-		executedProvenanceList.setEntityFinder(entityFinder);
```
nickgros, line 53: This finder is now configured in the list object

src/main/java/org/sagebionetworks/web/client/widget/entity/editor/EntityListConfigEditor.java
```diff
@@ -88,16 +89,26 @@ public void refresh() {
...
+				.setMultiSelect(true)
+				.setShowVersions(true)
+				.setSelectedMultiHandler(((selected, entityFinder) -> {
+					entityFinder.hide();
+					for (Reference ref : selected) {
+						EntityGroupRecord record = createRecord(ref.getTargetId(), ref.getTargetVersionNumber(), null);
+						entityListWidget.addRecord(record);
+					}
+					view.setButtonToolbarVisible(true);
+				}))
```
nickgros, line 108: This widget now supports selecting multiple entities

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -284,6 +284,9 @@ public static void showNotification(String message, String href, String buttonTe
 		settings.setType(notifyType);
 		if (timeout != null) {
 			settings.setDelay(timeout);
+		} else {
+			// TODO: For Entity Finder testing, we set notifications to stay open for 30s by default.
+			settings.setDelay(30 * 1000);
```
nickgros: TODO: Remove this before merge

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -823,10 +826,6 @@ public static FileHandle getFileHandle(EntityBundle bundle) {
 		return fileHandle;
 	}
 
-	public interface SelectedHandler<T> {
-		public void onSelected(T selected);
-	}
```
nickgros, line 828: This was only used by the Entity Finder

package.json
```diff
@@ -8,7 +8,8 @@
   	"react": "16.13.1",
   	"react-dom": "16.13.1",
     "sass": "1.32.5",
-    "synapse-react-client": "1.15.85",
+    "sanitize-html": "^1.18.4",
+    "synapse-react-client": "0.0.0-experimental-9587ccdd",
```
nickgros: TODO: Use release version before merge

package.json
```diff
@@ -8,7 +8,8 @@
   	"react": "16.13.1",
   	"react-dom": "16.13.1",
     "sass": "1.32.5",
-    "synapse-react-client": "1.15.85",
+    "sanitize-html": "^1.18.4",
```
nickgros, line 11: `sanitize-html` is omitted from the Rollup build and it's needed to sanitize annotations before we render them in HTML in the SRC EntityBadge. Currently it's only used in the EntityFinder, so it just needs to go in with it.

src/main/webapp/Portal.html
```diff
@@ -125,8 +125,8 @@
 			head.load(cdnEndpoint + "generated/react-jsonschema-form.js");
 			head.load(cdnEndpoint + "generated/universalCookie.min.js");
 			head.load(cdnEndpoint + "generated/react-bootstrap.min.js");
+			head.load(cdnEndpoint + "generated/sanitize-html.min.js");
 			head.load(cdnEndpoint + "generated/synapse-react-client.production.min.js");
-			head.load(cdnEndpoint + "js/sanitize-html.min.js");
```
nickgros, line 129: sanitize-html is now retrieved via yarn...where is this used so we can validate that this doesn't break whatever was using the old version?

src/main/webapp/sass/_core.scss
```diff
@@ -2213,6 +2213,12 @@ Looks for any child element under .modal-fullscreen that has a class that contai
 	width: 90%;
 }
 
+.modal-max-content [class*="modal-dialog"] {
+    width: max-content;
+    max-width: 95%;
+}
```
nickgros, line 2219: "Responsive" modal width: 


https://user-images.githubusercontent.com/17580037/114752868-c84ff980-9d24-11eb-87db-0c0d64c7aeb6.mov

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFilter.java
jay-hodgson: Confused by this.  In my mind DIRECTORY is the same as CONTAINER, what do you mean by DIRECTORY?

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2Impl.java
jay-hodgson: What ends up catching this?  Is it the general Uncaught JS Exception handler for the entire app?  If so, it will be logged in the console and sent to the repo client logs, but it also may have unintended side effects on the running web app.

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2ViewImpl.java
```diff
@@ -0,0 +1,271 @@
...
+			$wnd.ReactDOM.render(
+				$wnd.React.createElement($wnd.SRC.SynapseComponents.EntityFinder, props, null),
+				el
+			);
+		} catch (err) {
+			console.error(err);
+		}
+	}-*/;
+
+	public static JsArrayString toJsArray(List<String> list) {
```
jay-hodgson: This seems like it is of general use, and should be in a js utility class.  Perhaps convertFromJso and ReferenceJso too.

src/main/java/org/sagebionetworks/web/client/widget/entity/editor/EntityListConfigEditor.java
jay-hodgson: Thanks for enhancing this while you were here

src/main/java/org/sagebionetworks/web/client/widget/entity/editor/ProvenanceConfigViewImpl.java
```diff
@@ -37,30 +35,26 @@
...
-							showErrorMessage(DisplayConstants.PLEASE_MAKE_SELECTION);
-						}
+
+		entityFinderButton.addClickHandler(event -> entityFinderBuilder
+				.setInitialScope(EntityFinderScope.CURRENT_PROJECT)
+				.setInitialContainer(EntityFinder.InitialContainer.PROJECT)
+				.setHelpMarkdown("Search or Browse Synapse to find an item and display the Provenance Graph within the Wiki page")
+				.setPromptCopy("Find items to insert a Provenance Graph")
+				.setMultiSelect(true)
+				.setSelectableTypes(EntityFilter.ALL)
```
jay-hodgson: Not sure about showing provenance for a Link entity (makes sense for the target of the Link!)

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ImportTableViewColumnsButton.java
```diff
@@ -32,35 +30,33 @@
...
 	}
 
-	public void onTableViewSelected(String entityId) {
+	public void onTableViewSelected(String entityId, Long versionNumber) {
 		// get the column schema
 		EntityBundleRequest bundleRequest = new EntityBundleRequest();
 		bundleRequest.setIncludeEntity(true);
 		bundleRequest.setIncludeTableBundle(true);
-		jsClient.getEntityBundle(entityId, bundleRequest, new AsyncCallback<EntityBundle>() {
+		jsClient.getEntityBundleForVersion(entityId, versionNumber, bundleRequest, new AsyncCallback<EntityBundle>() {
```
jay-hodgson, line 63: another nice enhancement (pull in schema associated to a particular version)

src/main/webapp/Portal.html
jay-hodgson: I'd verify a few places.  As a user who is not on the special js/html/css team, try uploading an html file that has js, and verify that the preview looks ok.  But SynapseJSNIUtilsImpl.sanitizeHtml() uses [xss](https://www.npmjs.com/package/xss), which is still loaded from the js directory, so I think sanitize-html was just being loaded for SRC.

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFilter.java
nickgros: Before my changes, `ALL` was what we show in the EntityTree (projects, folders, files, links). `ALL` is obviously no longer a good name for this. It's kind of an arbitrary collection, so I'm not sure if there's a better name for it. For now, renaming to `PROJECT_FOLDER_FILE_LINK`

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2Impl.java
nickgros: Correct--it ends up getting logged in the console. I'm not super worried about performance because this should never happen unless something is misconfigured in code. I will also show the Jira Issue screen.

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -285,8 +285,8 @@ public static void showNotification(String message, String href, String buttonTe
 		if (timeout != null) {
 			settings.setDelay(timeout);
 		} else {
-			// TODO: For Entity Finder testing, we set notifications to stay open for 30s by default.
-			settings.setDelay(30 * 1000);
+			// By default, notifications stay open for 15s
+			settings.setDelay(15 * 1000);
```
nickgros, line 289: Halving the value, but making this change permanent after conversation with Duminda

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFilter.java
```diff
@@ -19,13 +20,13 @@
 
 public enum EntityFilter {
 	ALL(EntityType.values()),
-	ALL_DIRECTORY(project, folder, file, link),
+	ALL_BUT_LINK(project, folder, file, table, entityview, dockerrepo, submissionview),
+	PROJECT_FOLDER_FILE_LINK(project, folder, file, link),
```
nickgros, line 24: This is what's displayed in the EntityTree by default. It used to be called `ALL`, but it could still use a better name than this.

src/main/java/org/sagebionetworks/web/client/jsni/Reference.java
```diff
@@ -0,0 +1,33 @@
+package org.sagebionetworks.web.client.jsni;
+
+import com.google.gwt.core.client.JavaScriptObject;
+
+public class Reference extends JavaScriptObject {
```
jay-hodgson: Do we have a standard class naming convention to indicate js objects yet?  Like "ReferenceJsObject"

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/SynapseAlertImpl.java
```diff
@@ -88,7 +89,7 @@ public void handleException(Throwable ex) {
 			view.showError(DisplayConstants.ERROR_CONFLICTING_UPDATE + "\n" + message);
 		} else if (ex instanceof DeprecatedServiceException) {
 			view.showError(DisplayConstants.ERROR_DEPRECATED_SERVICE + "\n" + message);
-		} else if (ex instanceof UnknownErrorException) {
+		} else if (ex instanceof UnknownErrorException || ex instanceof WebClientConfigurationException) {
```
jay-hodgson, line 92: This seems appropriate.  It's an unexpected exception that we should go directly to development.

PR #4528, opened 2021-03-02
SWC-5354

jay-hodgson:
restyles FullWidthAlerts (like the Cookie notification), and changes to the styling of all Bootstrap Alerts in SWC.


src/main/java/org/sagebionetworks/web/client/view/LoginViewImpl.java
```diff
@@ -5,7 +5,7 @@
 import org.sagebionetworks.web.client.DisplayUtils;
 import org.sagebionetworks.web.client.SynapseJSNIUtils;
 import org.sagebionetworks.web.client.utils.Callback;
-import org.sagebionetworks.web.client.widget.InfoAlert;
```
jay-hodgson, line 8: Renamed InfoAlert to FullWidthAlert (since it's not always an Info)

src/main/java/org/sagebionetworks/web/client/widget/FullWidthAlert.java
```diff
@@ -0,0 +1,145 @@
...
+	@UiField
+	Icon icon;
+	@UiField
+	Span messageSpan;
+	@UiField
+	Strong messageTitle;
+	@UiField
+	Button primaryButton;
+	@UiField
+	Button secondaryButton;
```
jay-hodgson, line 35: Added the ability to configure a secondary button (buttons are not shown unless you set the button text)

src/main/java/org/sagebionetworks/web/client/widget/FullWidthAlert.java
```diff
@@ -0,0 +1,145 @@
...
+		primaryButton.setVisible(true);
+		primaryButton.setText(text.toUpperCase());
+	}
+	public void setPrimaryCTAButtonType(ButtonType type) {
+		primaryButton.setType(type);
+	}
+
+	public void setPrimaryCTAHref(String href) {
+		primaryButton.setHref(href);
+		addPrimaryCTAClickHandler(event -> {
```
jay-hodgson, line 100: Setting the href on the button does not hook up the event.

src/main/webapp/sass/_core.scss
```diff
@@ -3076,10 +3076,102 @@ div.border {
 	font-style:italic;
 }
 
-#body .fullWidthAlert {
-	@include border-radius(0px);
-	box-shadow: none;
-	padding: 7px 15px;
+#body .synapseAlert {
```
jay-hodgson, line 3079: See https://sagebionetworks.jira.com/browse/SWC-5354 for screenshots of what this looks like in practice.

src/main/webapp/sass/_core.scss
```diff
@@ -3076,10 +3076,102 @@ div.border {
...
+	&.global {
```
jay-hodgson, line 3080: if it also has the `global` class, then fix it to the bottom-left of the viewport with full width.

src/main/webapp/sass/_core.scss
```diff
@@ -3076,10 +3076,102 @@ div.border {
...
+		}
+		.primaryButton {
+			grid-column: 2 / span 2;
+			grid-row: 2 / span 1;
+		}
+	}	
+}
+
+// Bootstrap Alert overrides
+#body {
```
jay-hodgson, line 3138: override all alerts in synapse.org to have the new styles from the design (including the synapseAlerts that have special styling above).

src/main/java/org/sagebionetworks/web/client/widget/FullWidthAlert.java
```diff
@@ -0,0 +1,145 @@
...
+				break;
+			case SUCCESS: 
+			case DEFAULT: 
+			default: defaultIcon = IconType.CHECK;
+		}
+		icon.setType(defaultIcon);
+		w.setType(type);
+	}
+
+	public void setDismissable(boolean dismissable) {
```
nickgros, line 142: I'm guessing we typically use whatever the default value for `dismissable` is for the Alert component? Should we lift the default up to this level for visibility?

src/main/resources/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.ui.xml
```diff
@@ -9,8 +9,8 @@
 		type='org.sagebionetworks.web.client.SageImageBundle' />
 
 	<bh:Div addStyleNames="header">
-		<b:Alert ui:field="portalAlert" visible="false"
-			addStyleNames="fullWidthAlert flexcontainer-row background-color-white" type="INFO" height="50px"
+		<bh:Div ui:field="portalAlert" visible="false"
+			addStyleNames="padding-15 flexcontainer-row background-color-white"
```
nickgros, line 13: Is this no longer full-width at the top of the page, or is this maintaining its style some other way? Just double-checking because there is no screenshot

src/main/webapp/sass/_core.scss
nickgros: Just concerned that this might affect alerts created in SRC before those are modified to look like these alerts. Consider double checking e.g. the personal access tokens danger alert when you try to create a token with no name.

nickgros (approved):
Looks great! Few things just to double check, if you didn't verify them already

src/main/webapp/sass/_core.scss
jay-hodgson: Screen Shot 2021-03-02 at 1 49 25 PM

src/main/java/org/sagebionetworks/web/client/widget/FullWidthAlert.java
jay-hodgson: Not sure what you mean, could you please elaborate on your recommendation?

src/main/resources/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.ui.xml
jay-hodgson: Oh yeah, should have included a screenshot of this too.  It's a special case, the portal back bar.  It's no longer an Alert (because the new style conflicts with the portal back bar design):  
Screen Shot 2021-03-02 at 1 53 27 PM

src/main/webapp/sass/_core.scss
nickgros: Fantastic!

src/main/java/org/sagebionetworks/web/client/widget/FullWidthAlert.java
nickgros: It's not clear to me what the default value for `dismissable` is. Unless I missed something, it's not specified here or in the corresponding ui.xml. So my assumption is that gwtbootstrap3's Alert specifies the default value.

Just wondering if we should set a default value somewhere in our code (particularly if design has a preference). Ultimately not a big deal.

PR #4529, opened 2021-03-03
SWC-5431

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/PortalGinInjector.java
```diff
@@ -536,7 +538,7 @@
 
 	StringEditorCell createStringEditorCell();
 	
-	JSONListCellEditor createListCellEditor();
+	JSONListCellEditor createJSONListCellEditor();
```
jay-hodgson, line 541: Renamed injector method to be more clear (since we have a ListCellEditor as well)

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/cell/EnumFormCellEditor.java
```diff
@@ -73,33 +97,57 @@ public void fireEvent(GwtEvent<?> event) {
...
-		/*
-		 * The items include all passed values
-		 */
+		view.clear();
 		this.items = new ArrayList<String>(validValues.size());
 		for (String value : validValues) {
 			this.items.add(value);
 		}
-		view.configure(this.items);
+		if (validValues.size() > MAX_RADIO_BUTTONS) {
```
jay-hodgson, line 109: If configured with more than MAX_RADIO_BUTTONS items, then it uses a dropdown list view.  Otherwise, it uses the old radio buttons.

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/results/cell/EnumFormCellEditorImplTest.java
```diff
@@ -5,71 +5,121 @@
...
 	@Before
 	public void before() {
-		mockView = Mockito.mock(RadioCellEditorView.class);
-		editor = new EnumFormCellEditor(mockView);
+		when(mockGinInjector.createListCellEditorView()).thenReturn(mockListView);
+		when(mockGinInjector.createRadioCellEditorView()).thenReturn(mockRadioView);
+		editor = new EnumFormCellEditor(mockDivView, mockGinInjector);
 	}
 
+	private List<String> getValuesForDropdownList() {
```
jay-hodgson, line 44: Helper method, generates MAX_RADIO_BUTTONS + 1 values (to trigger the dropdown list renderer)

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/cell/IntegerCellEditorView.java
```diff
@@ -0,0 +1,12 @@
+package org.sagebionetworks.web.client.widget.table.v2.results.cell;
+
+import java.util.List;
+
+import com.google.gwt.event.dom.client.HasKeyDownHandlers;
+import com.google.gwt.user.client.TakesValue;
+import com.google.gwt.user.client.ui.Focusable;
+import com.google.gwt.user.client.ui.IsWidget;
+
+public interface IntegerCellEditorView extends IsWidget, TakesValue<Integer>, HasKeyDownHandlers, Focusable {
```
zimingd: When I see `IntegerCellEditorView`, I immediately think of an input box allowing only integer values.

I would describe this as a `ListIndexSelectionCellEditorView`? Or maybe `ListIndexChoiceCellEditorView` if `Selection` seems to imply the HTML `<select/>` tag?

zimingd (changes requested):
Logic looks great. `IntegerCellEditorView` just needs to be renamed to something less ambiguous.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/cell/IntegerCellEditorView.java
jay-hodgson: That's fair, renamed to `ListIndexSelectionCellEditorView`

zimingd (approved):

PR #4531, opened 2021-03-04
SWC-5373: Add maintenance page component. 

emmanmills:


jay-hodgson:
Approved, but it is dependent on a new SRC version.

src/main/java/org/sagebionetworks/web/client/presenter/DownPresenter.java
```diff
@@ -62,15 +58,12 @@ public void onSuccess(StackStatus status) {
 						break;
 					case READ_ONLY:
 					case DOWN:
-						// it's down, report the message and check again later
-						view.setMessage(status.getCurrentMessage());
```
jay-hodgson: Alternatively, we could try to surface the status message to the user in the react component.
Perhaps we pass a string into _createSRCDown() and use it for the message.  Called once during an attach event (with the string "We're busy updating Synapse for you and will be back soon."), and in setMessage().

emmanmills: Sure, I will work on that.

emmanmills: Actually, looking at the design's figma file, the message has 2 parts...a title (e.g. Sorry, Synapse is down for maintenance.) and another message (e.g. We’re busy updating Synapse for you and will be back soon). See https://sagebionetworks.jira.com/browse/SWC-5373. The original setMessage() seems to only provide the title part.

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -116,7 +116,7 @@ public static boolean isIgnoredErrorMessage(String error) {
 	public void handleUncaughtException(Throwable e) {
 		try {
 			if (!isIgnoredErrorMessage(e.getMessage())) {
-				GWT.debugger();
```
emmanmills: I comment this out because it causes the page to pause during loading

src/main/java/org/sagebionetworks/web/client/presenter/DownPresenter.java
emmanmills: I hardcoded the title but the smaller more information paragraph will be the error message from the backend.

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
jay-hodgson: This is a global uncaught js exception handler, and something that we need to track down before merging.  Please leave in (it's there on purpose)!

src/test/java/org/sagebionetworks/web/unitclient/presenter/DownPresenterTest.java
```diff
@@ -108,20 +104,4 @@ public void testRepoUpHasLastPlace() {
 		verify(mockGlobalAppState).back();
 	}
 
-	@Test
```
jay-hodgson, line 111: I think you should leave these two unit tests in (testRepoDown() and testCheckFailure())

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
emmanmills: Thanks, I revert the change here.

src/test/java/org/sagebionetworks/web/unitclient/presenter/DownPresenterTest.java
emmanmills: Thanks, I added it back.

src/test/java/org/sagebionetworks/web/unitclient/presenter/DownPresenterTest.java
```diff
@@ -107,8 +103,7 @@ public void testRepoUpHasLastPlace() {
 		presenter.checkForRepoDown();
 		verify(mockGlobalAppState).back();
 	}
-
-	@Test
+
```
jay-hodgson, line 106: Sorry to nitpick, I think this needs the `@Test`  annotation for it to be run.

jay-hodgson (approved):

src/test/java/org/sagebionetworks/web/unitclient/presenter/DownPresenterTest.java
emmanmills: No problem, I added it back.

PR #4536, opened 2021-03-08
SWC-5378 - fix DOIs for Views

nickgros:
The previous code worked for tables, but did not work for views. In this release (348), I added isLatestVersion to the backend, planning to backfill this later. We can use it here, and it should work for all entities.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -588,12 +588,8 @@ public class EntityActionControllerImpl implements EntityActionController, Actio
 		Entity entity = entityBundle.getEntity();
 		if (entity instanceof Versionable) {
 			VersionableEntity versionableEntity = ((VersionableEntity) entity);
-			if (versionableEntity instanceof Table
-					&& versionableEntity.getVersionLabel() != null
-					&& versionableEntity.getVersionLabel().equals("in progress")) {
-				// TODO: This is undefined behavior. Use `isLatestVersion` when PLFM-6583 is complete.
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -583,17 +583,22 @@ public void onFailure(Throwable caught) {
...
+	private Optional<Long> getVersionIfNotLatest() {
 		Long version = null;
 		Entity entity = entityBundle.getEntity();
-		if (entity instanceof Versionable) {
+		if (entity instanceof VersionableEntity) {
 			VersionableEntity versionableEntity = ((VersionableEntity) entity);
-			if (versionableEntity instanceof Table
-					&& versionableEntity.getVersionLabel() != null
-					&& versionableEntity.getVersionLabel().equals("in progress")) {
-				// TODO: This is undefined behavior. Use `isLatestVersion` when PLFM-6583 is complete.
```
nickgros, line 594: The above condition didn't hold for Views (they don't have the label, `"in progress"`), which caused the issue.

jay-hodgson: Looking for that hard-coded version label text should have been a red flag during code review, sorry I missed this.

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
nickgros: I think I explicitly told you it was there, so I think you saw it. I think I said I was sure it wouldn't break anything, and was wrong! My fault.

PR #4537, opened 2021-03-08
SWC-5436: update SRC (fixes success button colors), and rework synaps…

jay-hodgson:
…e alert layout to be a flexbox instead of css grid


jay-hodgson:
Screen Shot 2021-03-08 at 12 08 56 PM
Screen Shot 2021-03-08 at 12 09 12 PM

nickgros (approved):

PR #4543, opened 2021-03-12
SWC-5423: support Profile:edit token again, and go there from Settings

jay-hodgson:


jay-hodgson:
Today https://www.synapse.org/#!Profile:edit redirects to Settings.  And the Edit Profile in Settings redirects to the Profile tab, without the profile editor.  It's a mess.
Fix https://www.synapse.org/#!Profile:edit (now flips into edit mode).  And change the Edit Profile in Settings to go to this special token

src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -197,13 +197,14 @@ public String mayStop() {
 
 	public void editMyProfile() {
 		if (authenticationController.isLoggedIn()) {
-			goTo(new Profile(authenticationController.getCurrentUserPrincipalId(), ProfileArea.SETTINGS));
+			getUserProfileEditorWidget().setIsEditingMode(true);
```
jay-hodgson, line 200: Put into edit mode, and then viewMyProfile (which updates the user bundle, ...)

src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -197,13 +197,14 @@ public String mayStop() {
...
-			goTo(new Profile(authenticationController.getCurrentUserPrincipalId(), ProfileArea.SETTINGS));
+			getUserProfileEditorWidget().setIsEditingMode(true);
+			viewMyProfile();
 		} else {
 			view.showLoginAlert();
 		}
 	}
 
-	public void viewMyProfile(String area) {
+	public void viewMyProfile() {
```
jay-hodgson, line 207: The area was never referenced.

src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -273,6 +274,8 @@ private void getUserProfile() {
 		currentUserBundle = null;
 		int mask = PROFILE | ORC_ID;
 		Long currentUserIdLong = currentUserId != null ? Long.parseLong(currentUserId) : null;
+		// capture current state of user profile editor widget, to restore after reconfiguring with the current user profile.
+		boolean isEditing = getUserProfileEditorWidget().isEditingMode();
```
jay-hodgson, line 278: When we update the UserBundle, remember if the profile widget is in edit mode, and restore after we reconfigure.

src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
```diff
@@ -283,7 +283,7 @@ public Widget asWidget() {
 
 	@Override
 	public void onEditProfile() {
-		globalApplicationState.getPlaceChanger().goTo(new Profile(authenticationController.getCurrentUserPrincipalId(), ProfileArea.PROFILE));
+		globalApplicationState.getPlaceChanger().goTo(new Profile(Profile.EDIT_PROFILE_TOKEN));
```
jay-hodgson, line 286: Edit Profile button from the Settings tab goes to #!Profile:edit

PR #4547, opened 2021-03-17
SWC-5383: do not rely on click handler to add answer index into map

jay-hodgson:
which used to also recreate the answer set on click


xschildw (approved):
Thx for the explanations @jay-hodgson !

PR #4550, opened 2021-03-19
SWC-5469 - Refine notifications, add success/warning/danger

nickgros:
https://sagebionetworks.jira.com/browse/SWC-5469

This should only affect notifications pushed through DisplayUtils, not FullWidthAlerts.

image

### DIFF
--- a/src/main/webapp/sass/_core.scss
+++ b/src/main/webapp/sass/_core.scss
@@ -602,20 +602,6 @@ div.favoriteIcon i, button.favoriteIcon>i, i.favoriteIcon, a.favoriteIcon, a.fav
 	margin-right: auto;
 }
 
-.alert, #body .alert {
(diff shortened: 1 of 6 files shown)


src/main/webapp/sass/_core.scss
```diff
@@ -3155,6 +3144,17 @@ div.border {
 				color: #48896B;
 			}
 		}
+		.close {
+			font-size: 0px;
```
jay-hodgson, line 3148: The things we have to do!

jay-hodgson (approved):

src/main/webapp/sass/_core.scss
nickgros: I'm looking forward to deleting these hacks if we can ever manage to free ourselves from Bootstrap 3 😉

PR #4551, opened 2021-03-19
SWC-5371 and SWC-5373

jay-hodgson:


src/main/resources/org/sagebionetworks/web/client/widget/entity/controller/SynapseAlertViewImpl.ui.xml
```diff
@@ -1,16 +1,23 @@
...
-  	<bh:Div>
-		<bh:Div ui:field="loginWidgetContainer" />
-  		<b:Alert ui:field="alert" type="DANGER" addStyleNames="synapseAlert margin-top-10 margin-bottom-0-imp">
-  			<bh:Div addStyleNames="iconAndMessageArea">
-	  			<b:IconStack addStyleNames="iconArea font-size-20">
+  xmlns:w="urn:import:org.sagebionetworks.web.client.widget">
+
+	<bh:Div>
+		<bh:Div ui:field="loginWidgetContainer" >
+			<w:FullWidthAlert
```
jay-hodgson, line 10: Use a FullWidthAlert instead of the LoginWidget (which is really the LoginPage) in this case

src/main/resources/org/sagebionetworks/web/client/widget/entity/controller/SynapseAlertViewImpl.ui.xml
```diff
@@ -1,16 +1,23 @@
...
+				message="You will need to sign in to access this resource."
+				primaryCTAText="Sign In"
+				primaryCTAHrefTargetSelf="/#!LoginPlace:0"
```
jay-hodgson, line 13: Open in the same window.

src/main/resources/org/sagebionetworks/web/client/widget/entity/controller/SynapseAlertViewImpl.ui.xml
```diff
@@ -1,16 +1,23 @@
...
+				alertType="INFO"
+				global="false"
```
jay-hodgson, line 15: In line, not absolute positioning (fixed to the bottom of the screen).

src/main/java/org/sagebionetworks/web/client/SynapseStatusDetector.java
```diff
@@ -13,28 +20,52 @@
...
 		_getCurrentStatus(this);
 		_getUnresolvedIncidents(this);
 		_getScheduledMaintenance(this);
-
+		getSynapseStackStatus();
+		
 		gwt.scheduleFixedDelay(() -> {
 			_getCurrentStatus(this);
 			_getUnresolvedIncidents(this);
+			getSynapseStackStatus();
```
jay-hodgson, line 48: In addition to checking StatusPage for the status (and incidents), also check Synapse directly.  If we detect that we're in RO mode (or Down), redirect to the Down page (where it will click Back() when we're back in RW mode).

xschildw (approved):
Thx for the comments!

PR #4573, opened 2021-04-28
SWC-5555: release modal dialog when reconfigured

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/FilesTabViewImpl.java
```diff
@@ -161,6 +161,7 @@ public void setProvenance(Widget w) {
 		if (expandProvHandlerRegistration != null) {
 			expandProvHandlerRegistration.removeHandler();
 		}
+		modalDialogContainer.clear();
```
jay-hodgson, line 164: this causes modal detach which cleans it up.

xschildw (approved):

PR #4575, opened 2021-05-03
SWC-5372

nickgros:
https://sagebionetworks.jira.com/browse/SWC-5372

Some issues with tab ordering, but it's a start

https://user-images.githubusercontent.com/17580037/116930148-5f75e600-ac2d-11eb-82ef-4858955382f0.mov

### DIFF
--- a/src/main/webapp/sass/_core.scss
+++ b/src/main/webapp/sass/_core.scss
@@ -225,9 +225,6 @@ end sticky footer
 		padding-left: 0px;
 		min-width: 27px;
 	}
-	input[type="checkbox"], input[type="radio"] {
-		display:none;
-	}
 
 	.radio {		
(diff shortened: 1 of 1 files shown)


jay-hodgson:
Definitely a start, thanks @nickgros !  Why does this work (deleting the display:none style only), wouldn't this just show the real html control without the other hack (or is this using the hack from SRC?)?

jay-hodgson (approved):

nickgros:
> Wouldn't this just show the real html control without the other hack (or is this using the hack from SRC?)?

It's using the same hack (straight from the SRC SCSS) b/c the checkboxes are implemented in nearly the same way. The radio buttons are a bit different, but those were only missing focus styles.

PR #4577, opened 2021-05-04
SWC-5586

nickgros:
https://user-images.githubusercontent.com/17580037/117050448-58a9aa80-ace3-11eb-9950-fd57bf50ca1a.mov

### DIFF
--- a/src/main/webapp/sass/_core.scss
+++ b/src/main/webapp/sass/_core.scss
@@ -3432,3 +3432,11 @@ div.border {
 #body .EntityFinderSelectionPane.alert.alert-warning {
     border-left: none;
 }
+
+.entityFinder .synAlertPanel {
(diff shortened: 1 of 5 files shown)


src/main/webapp/sass/_core.scss
```diff
@@ -3431,4 +3431,12 @@ div.border {
...
     border-left: none;
+}
+
+.entityFinder .synAlertPanel {
+    display: flex;
+    > * {
+        width: 600px;
+        flex-grow: 1;
+    }
 }
```
nickgros, line 3442: Kinda hacky, but it lets us fill the modal width without going over (can't use `width: 100%` because parent has `width: max-content`)

jay-hodgson (approved):
Ok

PR #4579, opened 2021-05-04
SWC-5583

nickgros:
Before this, the react component would render after the modal was shown. Because the modal's size depends on its contents, it would appear small and then immediately become large, which was distracting. See clip in [the issue](https://sagebionetworks.jira.com/browse/SWC-5583).

In this change, we show the modal after the ReactDOM.render call completes (via a callback). The modal immediately appears with the final width.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2ViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2ViewImpl.java
@@ -4,6 +4,7 @@ import static org.sagebionetworks.web.client.utils.JavaScriptArrayUtils.convertT
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderV2ViewImpl.java
```diff
@@ -231,7 +224,8 @@ private static native void _showEntityFinderReactComponent(Element el, String se
 			};
 			$wnd.ReactDOM.render(
 				$wnd.React.createElement($wnd.SRC.SynapseComponents.EntityFinder, props, null),
-				el
+				el,
+				function() { onRender.@java.lang.Runnable::run()() }
```
jay-hodgson, line 228: What do you think about adding a "react component integration" section to the SWC readme that would contain generally useful solutions like this?

jay-hodgson (approved):

PR #4581, opened 2021-05-05
Guide for using JsInterop + React, refactor Entity Finder to use JsInterop

nickgros:
* Refactor Entity Finder to use JsInterop over JSNI
* Add a guide for using JsInterop to call an SRC component.


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderWidget.java
```diff
@@ -3,17 +3,18 @@
 import java.util.List;
 
 import org.sagebionetworks.repo.model.Reference;
+import org.sagebionetworks.web.client.jsinterop.EntityFinderScope;
 
-public interface EntityFinder {
+public interface EntityFinderWidget {
```
nickgros, line 8: I had to rename EntityFinder -> EntityFinderWidget or else it collided with SRC.SynapseComponents.EntityFinder JsInterop class

jay-hodgson:
This PR provides both documentation on using JsInterop, and an example of converting from JSNI to JsInterop!

jay-hodgson (approved):

PR #4584, opened 2021-05-15
SWC-5589

jay-hodgson:


jay-hodgson:
TODO: update SRC to version 1.16.2 once it's available on npm.
https://github.com/Sage-Bionetworks/Synapse-React-Client/pull/1043

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -64,62 +65,62 @@ public void onFailure(Throwable reason) {
...
-									ginjector.getVersionTimer().start();
-									// start timer to check for Synapse outage or scheduled maintenance
-									ginjector.getSynapseStatusDetector().start();
-									// Goes to place represented on URL or default place
-									historyHandler.handleCurrentHistory();
-									globalApplicationState.initializeDropZone();
-									// initialize the view default columns so that they're ready when we need them (do this by constructing that singleton object)
-									ginjector.getViewDefaultColumns();
-								}
+							ginjector.getAuthenticationController().checkForUserChange(() -> {
```
jay-hodgson, line 72: This actually speeds up app load if previously logged in (avoids page refresh).

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -64,62 +65,62 @@ public void onFailure(Throwable reason) {
 				@Override
 				public void onSuccess() {
 					try {
-						// start timer to check for user session state change (session expired, or user explicitly logged
-						// out)
-						ginjector.getSessionDetector().start();
```
jay-hodgson, line 69: session/token detection requires Synapse Properties to be loaded (for endpoints)

src/main/java/org/sagebionetworks/web/client/SessionDetector.java
```diff
@@ -22,7 +22,7 @@ public SessionDetector(AuthenticationController authController, GlobalApplicatio
 		this.authController = authController;
 		this.globalAppState = globalAppState;
 		this.gwt = gwt;
-		initializeSessionTokenState();
+		initializeAccessTokenState();
```
jay-hodgson, line 25: Lots of renaming (session token to access token)

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -407,36 +409,36 @@ private Request doDelete(String url, AsyncCallback callback) {
 		requestBuilder.configure(DELETE, url);
 		requestBuilder.setHeader(ACCEPT, APPLICATION_JSON_CHARSET_UTF8);
 		if (authController.isLoggedIn()) {
-			requestBuilder.setHeader(SESSION_TOKEN_HEADER, authController.getCurrentUserSessionToken());
```
jay-hodgson, line 410: change the header that's used for authorization (since we're using an access token now).

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -827,6 +829,10 @@ public void getUserProfile(String userId, AsyncCallback<UserProfile> callback) {
 		}
 	}
 
+	public FluentFuture<UserProfile> getMyUserProfile() {
+		return getUserProfile(null);
```
jay-hodgson, line 833: We load the current user profile during app load, and it also validates the current access token.  
During Google SSO login the call is made servlet side, and only the access token is sent back to the client, so the response to this call is also used to find out if the user still needs to sign the Synapse terms of use

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -1258,17 +1264,10 @@ public void getV2WikiOrderHint(WikiPageKey key, AsyncCallback<V2WikiOrderHint> c
 	}
 
 	public void login(LoginRequest loginRequest, AsyncCallback<LoginResponse> callback) {
-		String url = getAuthServiceUrl() + "/login";
+		String url = getAuthServiceUrl() + "/login2";
 		doPost(url, loginRequest, OBJECT_TYPE.LoginResponse, false, callback);
 	}
 
-	public void logout() {
```
jay-hodgson, line 1265: Service removed (just forget the token)

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -1534,16 +1533,20 @@ public void initSession(String token) {
...
 	 * 
 	 * @param token
 	 * @param callback
 	 */
 	public void initSession(String token, AsyncCallback<Void> callback) {
-		Session s = new Session();
-		s.setSessionToken(token);
-		String url = jsniUtils.getSessionCookieUrl();
-		doPost(url, s, OBJECT_TYPE.None, false, callback);
+		AccessTokenWrapper t = new AccessTokenWrapper();
```
jay-hodgson, line 1542: Created my own object to wrap the token.  It still has the same structure, but I don't rely on a backend json object that may go away.  Maintaining the structure (key is still "sessionToken") makes my life easier since other portals hosted on .synapse.org domains use the InitSessionServlet in www.synapse.org to access the current session token.  Eases migration from session token to access token for other portals.

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -1534,16 +1533,20 @@ public void initSession(String token) {
...
+		t.setToken(token);
+		String url = jsniUtils.getAccessTokenCookieUrl();
+		doPost(url, t, OBJECT_TYPE.None, false, callback);
+	}
+	
+	public FluentFuture<String> getAccessToken() {
```
jay-hodgson, line 1548: Wrote this as a FluentFuture because I thought I would want to run these two calls in parallel (getting the access token cookie value and the current user profile), but then realized the getMyProfile call depends on the access token (duh).

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -1675,21 +1678,6 @@ public void addTeamMember(String userId, String teamId, AsyncCallback<Void> cb)
 		doPut(url, null, OBJECT_TYPE.None, cb);
 	}
 
-	/**
-	 * If logged in, refresh the current session token to render it usable for another 24 hours
-	 */
-	public void refreshCurrentSessionToken() {
```
jay-hodgson, line 1681: Not supported for access tokens.

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -116,37 +123,52 @@ public void onFailure(Throwable caught) {
...
 	}
 
 	/**
-	 * Session cookie should be set before this call
+	 * Access token cookie should be set before this call
 	 * 
 	 * @param callback
 	 */
-	public void initializeFromExistingSessionCookie(AsyncCallback<UserProfile> callback) {
-		userAccountService.getCurrentUserSessionData(new AsyncCallback<UserSessionData>() {
```
jay-hodgson, line 130: UserSessionData is going away (it's session specific), so I needed to figure out how to do this in the new world during initialization.  First step is to get the access token (if there is one).  Call will fail if there is none, or if the access token is invalid (which is a new change to the InitSessionServlet doGet(), you can find below). 
If the access token is present (and valid), then get the current user profile.  If this fails because the terms of use have not been signed, then send the user to the ToU (like we did previously).

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -213,56 +234,56 @@ public UserProfile getCurrentUserProfile() {
...
 
 	@Override
 	public void checkForUserChange() {
-		String oldUserSessionToken = currentUserSessionToken;
-		initializeFromExistingSessionCookie(new AsyncCallback<UserProfile>() {
+		checkForUserChange(null);
+	}
+	
+	@Override
+	public void checkForUserChange(Callback optionalCallback) {
```
jay-hodgson: Added an optional callback, which is called after the access token has been initialized in the app.  This is useful for app initialization (in Portal.java).  If not set, then it has the same behavior as before (which the SessionDetector calls periodically).

src/main/java/org/sagebionetworks/web/server/servlet/DiscussionMessageServlet.java
```diff
@@ -60,13 +60,13 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
 	 *
 	 * @return
 	 */
-	private SynapseClient createNewClient(String sessionToken) {
+	private SynapseClient createNewClient(String accessToken) {
```
jay-hodgson, line 63: You'll see many of these too (for some reason this is reimplemented over and over again).  Rename sessionToken to accessToken, use the SynapseClient.setBearerAuthorizationToken() instead of setSessionToken(), and only set if accessToken is not null.

src/main/java/org/sagebionetworks/web/server/servlet/InitSessionServlet.java
```diff
@@ -32,6 +38,10 @@
 	private static final long serialVersionUID = 1L;
 	protected static final ThreadLocal<HttpServletRequest> perThreadRequest = new ThreadLocal<HttpServletRequest>();
 	public static final int ONE_DAY_IN_SECONDS = 60 * 60 * 24;
+	private SynapseProvider synapseProvider = new SynapseProviderImpl();
```
jay-hodgson, line 41: Needed to validate the access token when it is retrieved.

src/main/java/org/sagebionetworks/web/server/servlet/InitSessionServlet.java
```diff
@@ -47,18 +57,18 @@ public void service(ServletRequest arg0, ServletResponse arg1) throws ServletExc
...
-		// return the Set-Cookie response with the session token
+		// return the Set-Cookie response with the access token
 		try {
 			String sessionJson = IOUtils.toString(request.getReader());
 			JSONObjectAdapter adapter = new JSONObjectAdapterImpl(sessionJson);
-			Session s = new Session(adapter);
-			String sessionToken = s.getSessionToken();
-			if (sessionToken == null || sessionToken.isEmpty()) {
-				sessionToken = WebConstants.EXPIRE_SESSION_TOKEN;
+			AccessTokenWrapper s = new AccessTokenWrapper(adapter);
```
jay-hodgson, line 64: Use my AccessTokenWrapper instead of the backend Session object, but has the same key that I care about (sessionToken)

src/main/java/org/sagebionetworks/web/server/servlet/InitSessionServlet.java
```diff
@@ -114,16 +124,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
 		if (token != null) {
 			token = SimpleHtmlSanitizer.sanitizeHtml(token).asString(); // The token should not be HTML, but just in case (SWC-5504)
-			response.setContentType(WebConstants.TEXT_PLAIN_CHARSET_UTF8);
-			response.setHeader(WebConstants.CONTENT_TYPE_OPTIONS, WebConstants.NOSNIFF);
-			response.setStatus(HttpServletResponse.SC_OK);
-			response.getOutputStream().write(token.getBytes("UTF-8"));
-			response.getOutputStream().flush();
+			SynapseClient synapseClient = createNewClient(token);
+			try {
+				synapseClient.getMyProfile();
```
jay-hodgson: Validate

src/main/java/org/sagebionetworks/web/server/servlet/UserAccountServiceImpl.java
```diff
@@ -53,12 +53,12 @@ private void validateService() {
...
 
 	@Override
-	public void signTermsOfUse(String sessionToken, boolean acceptsTerms) throws RestServiceException {
+	public void signTermsOfUse(String accessToken) throws RestServiceException {
 		validateService();
 
 		SynapseClient synapseClient = createSynapseClient();
 		try {
-			synapseClient.signTermsOfUse(sessionToken, acceptsTerms);
+			synapseClient.signTermsOfUse(accessToken);
```
jay-hodgson, line 61: New call that takes an access token (and no boolean).

src/main/java/org/sagebionetworks/web/server/servlet/UserAccountServiceImpl.java
```diff
@@ -76,19 +76,6 @@ public void createUserStep1(NewUser newUser, String portalEndpoint) throws RestS
 		}
 	}
 
-	@Override
-	public UserSessionData getCurrentUserSessionData() throws RestServiceException {
```
jay-hodgson, line 80: UserSessionData is going away.

src/main/java/org/sagebionetworks/web/server/servlet/UserAccountServiceImpl.java
```diff
@@ -101,15 +88,15 @@ public String createUserStep2(String userName, String fName, String lName, Strin
 			accountSetup.setUsername(userName);
 			accountSetup.setPassword(password);
 			accountSetup.setEmailValidationSignedToken(emailValidationSignedToken);
-			Session s = client.createNewAccount(accountSetup);
```
jay-hodgson, line 104: Session is going away.  Use LoginResponse (and pass back access token instead of session token)

src/main/java/org/sagebionetworks/web/server/servlet/oauth2/OAuth2SessionServlet.java
```diff
@@ -61,8 +63,8 @@ public void validateUser(HttpServletRequest req, HttpServletResponse resp, OAuth
 			request.setAuthenticationCode(athenticationCode);
 			request.setProvider(provider);
 			request.setRedirectUrl(redirectUrl);
-			Session token = client.validateOAuthAuthenticationCode(request);
-			resp.sendRedirect(LOGIN_PLACE + token.getSessionToken());
+			LoginResponse token = client.validateOAuthAuthenticationCodeForAccessToken(request);
```
jay-hodgson, line 66: Use new service (supports access token)

src/main/java/org/sagebionetworks/web/server/servlet/oauth2/OAuth2SessionServlet.java
```diff
@@ -89,8 +91,8 @@ public void createAccountViaOauth(HttpServletRequest req, HttpServletResponse re
 			request.setProvider(provider);
 			request.setRedirectUrl(redirectUrl);
 			request.setUserName(username);
-			Session token = client.createAccountViaOAuth2(request);
-			resp.sendRedirect(LOGIN_PLACE + token.getSessionToken());
+			LoginResponse token = client.createAccountViaOAuth2ForAccessToken(request);
```
jay-hodgson, line 94: Use new service (supports access token)

src/test/java/org/sagebionetworks/web/client/SynapseJavascriptClientTest.java
```diff
@@ -917,8 +882,8 @@ public void testDeleteRequestsCannotBeCancelled() throws RequestException {
 		when(mockRequest2.isPending()).thenReturn(true);
 		when(mockGwt.getCurrentURL()).thenReturn(currentUrl);
 
-		client.logout();
-
+		client.deleteEntityById("syn111", null);
```
jay-hodgson, line 885: This test is verifying that a DELETE call cannot be canceled.  logout() has been removed from the codebase, so use another DELETE call.

package.json
```diff
@@ -9,7 +9,7 @@
   	"react-dom": "16.13.1",
     "sass": "1.32.5",
     "sanitize-html": "^1.18.4",
-    "synapse-react-client": "1.15.99",
+    "synapse-react-client": "1.16.2",
```
jay-hodgson, line 12: Uses Access Tokens

pom.xml
```diff
@@ -16,13 +16,13 @@
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<webappDirectory>${project.build.directory}/${project.build.finalName}</webappDirectory>
-		<synapse.version>354.0</synapse.version>
+		<synapse.version>2021-05-13-894-d467cd5</synapse.version>
```
jay-hodgson, line 19: Supports access tokens.

pom.xml
```diff
@@ -16,13 +16,13 @@
...
 		<gwtVersion>2.9.0</gwtVersion>
 		<org.springframework.version>4.3.18.RELEASE</org.springframework.version>
 		<guiceVersion>3.0</guiceVersion>
 		<ginVersion>2.1.2</ginVersion>
 		<xstream.version>1.3.1</xstream.version>
-		<schema-to-pojo.version>0.6.3</schema-to-pojo.version>
+		<schema-to-pojo.version>0.6.6</schema-to-pojo.version>
```
jay-hodgson, line 25: Has updates to JSON schema that are compatible with GWT compilation.

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -1258,17 +1264,10 @@ public void getV2WikiOrderHint(WikiPageKey key, AsyncCallback<V2WikiOrderHint> c
 	}
 
 	public void login(LoginRequest loginRequest, AsyncCallback<LoginResponse> callback) {
-		String url = getAuthServiceUrl() + "/login";
+		String url = getAuthServiceUrl() + "/login2";
```
jay-hodgson, line 1267: New service, uses access token

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
xschildw: Could we rename callback to webAppInitCallback to indicate it's only used for init?

src/test/java/org/sagebionetworks/web/client/SynapseJavascriptClientTest.java
xschildw: We want make sure that the delete does not added to the list of cancellable requests client.getRequests().

PR #4587, opened 2021-05-18
SWC-5581

nickgros:
Shows basic schema conformance in Entity badge in experimental mode

https://user-images.githubusercontent.com/17580037/118674501-d2539500-b7c7-11eb-8a99-723960b3df9d.mov

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
+++ b/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
@@ -122,6 +122,8 @@ import org.sagebionetworks.repo.model.principal.PrincipalAliasResponse;
 import org.sagebionetworks.repo.model.principal.TypeFilter;
 import org.sagebionetworks.repo.model.provenance.Activity;
 import org.sagebionetworks.repo.model.request.ReferenceList;
+import org.sagebionetworks.repo.model.schema.JsonSchemaObjectBinding;
+import org.sagebionetworks.repo.model.schema.ValidationResults;
 import org.sagebionetworks.repo.model.search.SearchResults;
 import org.sagebionetworks.repo.model.search.query.SearchQuery;
 import org.sagebionetworks.repo.model.subscription.SortByType;
@@ -295,7 +297,7 @@ public class SynapseJavascriptClient {
 	public static final String VIEW_COLUMN_MODEL_REQUEST = "/column/view/scope";
 	public static final String SCHEMA_TYPE_VALIDATION = "/schema/type/validation/";
 	public static final String DOWNLOAD_LIST_V2 = "/download/list/query/";
-	
+
 	public static final String ASYNC_START = "/async/start";
 	public static final String ASYNC_GET = "/async/get/";
 	public static final String AUTH_OAUTH_2 = "/oauth2";
@@ -359,7 +361,7 @@ public class SynapseJavascriptClient {
 	/**
 	 * For testing purposes only. Returns Requests associated to the given url. Requests are
 	 * periodically removed from this list once they leave the Pending state.
-	 * 
+	 *
 	 * @param forUrl
 	 * @return
 	 */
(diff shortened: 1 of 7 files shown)


src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -1847,10 +1849,53 @@ public void getAccessApprovalNotifications(String requirementId, List<String> re
...
+
+	/**
+	 * Retries (with exponential backoff) on mismatched eTag or 404
+	 * @param entityId
+	 * @param expectedEtag
+	 * @param cb
+	 */
+	public void getSchemaValidationResultsWithMatchingEtag(String entityId, String expectedEtag, AsyncCallback<ValidationResults> cb) {
+		getSchemaValidationResultsWithMatchingEtag(entityId, expectedEtag, INITIAL_RETRY_REQUEST_DELAY_MS, cb);
+	}
```
nickgros, line 1898: Schema validation is eventually consistent, where the eTag of the object will match the eTag in the validation results if it's up-to-date. So we want to retry (w/ backoff) on a 404 (assuming there is a schema binding) or eTag mismatch.

src/main/webapp/sass/_core.scss
```diff
@@ -3443,4 +3444,14 @@ div.border {
...
+
+.MissingAnnotations {
+    color: $synapse-gray-med;
+}
+.ValidAnnotations {
+ color: $synapse-green;
+}
+.InvalidAnnotations {
+  color: $synapse-yellow;
 }
```
nickgros: Could/should these be reused? Maybe .ValidIcon, .InvalidIcon, etc.

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadge.java
```diff
@@ -131,12 +133,52 @@ public void setEntityBundle(EntityBundle eb) {
...
+
+		// In experimental mode, check if there's a bound JSON Schema + check validity
+		if (DisplayUtils.isInTestWebsite(cookies)) {
+			final String finalAnnotationsHtml = annotationsHtml;
+			jsClient.getSchemaBinding(eb.getEntity().getId(), new AsyncCallback<JsonSchemaObjectBinding>() {
+				@Override
+				public void onSuccess(JsonSchemaObjectBinding result) {
+					boolean hasSchema = true;
+
+					jsClient.getSchemaValidationResultsWithMatchingEtag(eb.getEntity().getId(), eb.getEntity().getEtag(), new AsyncCallback<ValidationResults>() {
```
jay-hodgson, line 152: Based on how this works, it seems unlikely that the etag we pass here is out of date, so I think this is fine.  If this was more likely (if the schema validation happened on hover, for example), then this could get caught in an infinite loop (since the call keeps retrying until the etags match).

jay-hodgson:
I understand why this was done in SWC (since the file browser is implemented here so it's much more straight forward).  I'd like to see the annotation editor done in SRC though (for simple JSON schema).

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadge.java
```diff
@@ -131,12 +133,52 @@ public void setEntityBundle(EntityBundle eb) {
...
+							view.setError(caught.getMessage());
+						}
+					});
+
+				}
+
+				@Override
+				public void onFailure(Throwable caught) {
+					if (caught instanceof NotFoundException) { // The entity has no schema.
+						boolean hasSchema = false;
```
jay-hodgson, line 169: Thanks for the local variables to clarify parameters.

src/main/webapp/sass/_core.scss
jay-hodgson: You could use existing generalized classes.  .synapse-green, .synapse-yellow.  Probably need to make one for .synapse-gray-med

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadge.java
nickgros: Good catch, I didn't think of a race condition occurring in this order. I think here it’s not worth the effort to guard against it, but if I have to reimplement it in SRC, I’ll be cognizant of it.

PR #4592, opened 2021-05-21
SWC-5613: integrate UTIF.js

jay-hodgson:
lazy load the js when we attempt to preview an image/tiff file).
UTIF.js: https://www.npmjs.com/package/utif


xschildw (approved):
Reviewed with Jay.

PR #4602, opened 2021-06-15
Use a global react-query QueryClient

nickgros:
Creating a single query client and passing that in to our context object gives us the benefits of a shared query client despite components living in separate component trees.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,21 +1,22 @@
 {
   "dependencies": {
-    "jquery": "3.5.1",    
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -1,21 +1,22 @@
 {
   "dependencies": {
```
nickgros, line 2: `yarn add` sorted the dependencies. The only change is the addition of react-query

jay-hodgson (approved):

PR #4616, opened 2021-07-14
SWC-5663: revert update to sanitize-html

jay-hodgson:
and regenerate yarn.lock


jay-hodgson:
Going to pull this in to attempt to fix the build

PR #4637, opened 2021-08-10
SWC-5673

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -305,6 +312,7 @@ public boolean isEditing() {
 	@Override
 	public void setIsEditing(boolean isEditing) {
 		this.isEditing = isEditing;
+		ginInjector.getHeader().refresh();
```
jay-hodgson, line 315: Refresh the header whenever we change edit state.

src/main/java/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.java
```diff
@@ -83,6 +85,10 @@ public void clear() {
 
 
 	public void initClickHandlers() {
+		editModeNavBarClickBlocker.addClickHandler(event -> {
```
jay-hodgson, line 88: This is a layer on top of the nav bar where we can catch all click events (and set the background color to grey it out)

src/main/java/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.java
```diff
@@ -107,10 +113,13 @@ private void attachNavBar() {
 	
 	@Override
 	public void refresh() {
+		GlobalApplicationState globalAppState = ginInjector.getGlobalApplicationState();
 		if (globalAppState.getCurrentPlace() == null || globalAppState.getCurrentPlace() instanceof Home || globalAppState.getCurrentPlace() instanceof LoginPlace) {
 			detachNavBar();
 		} else {
 			attachNavBar();
+			// cover with a click catcher if in editing mode
+			editModeNavBarClickBlocker.setVisible(globalAppState.isEditing());
```
jay-hodgson, line 122: Nav bar is visible in this case.  Show the new layer if we're in editing mode.

xschildw (approved):

PR #4642, opened 2021-08-16
SWC-5632: ability to tell the entity cell renderer to also show the e…

jay-hodgson:
…ntity ID


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/cell/EntityIdCellRenderer.java
```diff
@@ -22,6 +22,7 @@
 	ClickHandler customClickHandler;
 	CellAddress address;
 	boolean hideIfLoadError;
+	boolean showEntityId = false;
```
jay-hodgson, line 25: By default, each EntityIdCellRenderer instance starts out with showEntityId = false.  But you can flip it to true (before setValue()!) to show it in the output link text

xschildw (approved):

PR #4644, opened 2021-08-17
SWC-5694 - In alpha mode, show new annotation components in modal

nickgros:
* Add SRC's EntityModal component.
* Show the modal's annotations page when opening Annotations in alpha mode.
* Remove widget that previously exposed the new annotation editor in alpha mode.

https://user-images.githubusercontent.com/17580037/129744409-57717a41-030e-4b7d-9ee0-28ebbf8f3102.mov


src/main/java/org/sagebionetworks/web/client/widget/entity/annotation/AnnotationEditorV2.java
```diff
@@ -1,23 +0,0 @@
-package org.sagebionetworks.web.client.widget.entity.annotation;
-
-import com.google.gwt.user.client.ui.Widget;
-import com.google.inject.Inject;
-
-public class AnnotationEditorV2 implements AnnotationEditorV2View.Presenter {
```
nickgros, line 6: This widget showed the new editor in its own modal in alpha mode. Since the editor is now contained in the EntityModal React component, this is no longer used.

src/main/java/org/sagebionetworks/web/client/widget/entity/annotation/AnnotationsRendererWidget.java
```diff
@@ -75,13 +71,6 @@ public Widget asWidget() {
...
 	public void onEdit() {
-		preflightController.checkUploadToEntity(bundle, () -> {
-			if (DisplayUtils.isInTestWebsite(cookies)) {
-				editorDialogV2.configure(bundle.getEntity().getId());
-			} else {
-				getEditAnnotationsDialog().configure(bundle);
-
-			}
-		});
+		preflightController.checkUploadToEntity(bundle, () -> getEditAnnotationsDialog().configure(bundle));
```
nickgros, line 74: Drop alpha mode behavior for showing the (now removed) annotation editor widget. With the other changes in this PR, if you're in alpha mode, you would never even see this (AnnotationsRendererWidget) widget, because the modal would open instead.

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadataViewImpl.java
```diff
@@ -20,125 +29,162 @@
...
+                        EntityModalProps.create(idField.getText(), showModal, () -> setAnnotationsVisible(false), "ANNOTATIONS", showTabs);
+                ReactDOM.render(
+                        React.createElementWithSynapseContext(
+                                SRC.SynapseComponents.EntityModal,
+                                props,
+                                propsProvider.getJsInteropContextProps()
+                        ),
+                        annotationsModalContainer.getElement()
+                );
+            }
```
nickgros, line 136: If the 'visible' variable changes we have to manually rerender with the changed `showModal` prop. This is super verbose without JSX!

jay-hodgson (approved):

PR #4650, opened 2021-08-19
SWC-5739: reintroduce the special @ user search to the entity search box

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/SearchUtil.java
```diff
@@ -19,7 +22,7 @@
 	 * @param place
 	 * @return
 	 */
-	public static Synapse willRedirect(Search place) {
+	public static Place willRedirect(Search place) {
```
jay-hodgson, line 25: Based on the search term, might return a Synapse place (if the query term starts with 'syn') or a PeopleSearch place (if it starts with '@')

src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
```diff
@@ -147,7 +147,7 @@ public void setSearchResults(SearchResults searchResults, String searchTerm) {
 		String facetNames = createShownFacets(searchResults);
 		String pageTitleSearchTerm = searchTerm != null && searchTerm.length() > 0 ? "'" + searchTerm + "' " : "";
 		synapseJSNIUtils.setPageTitle("Search: " + pageTitleSearchTerm + facetNames);
-		narrowResultsPanel.setVisible(true);
+		narrowResultsPanel.setVisible(searchResults.getFacets() != null);
```
jay-hodgson: Only show the "Narrow Results" panel if the search results have facets.

src/main/resources/org/sagebionetworks/web/client/view/PeopleSearchViewImpl.ui.xml
```diff
@@ -5,8 +5,11 @@
 	xmlns:b="urn:import:org.gwtbootstrap3.client.ui"
 	xmlns:bh="urn:import:org.gwtbootstrap3.client.ui.html">
 	<g:HTMLPanel>
-		<bh:Div addStyleNames="margin-left-15 margin-right-15">
-			<b:Heading size="H2" text="People Search" />
+		<bh:Div addStyleNames="pageHeader">
+			<b:Heading addStyleNames="pageHeaderTitle" size="H3" text="People Search"/>
+		</bh:Div>
```
jay-hodgson, line 10: Update header style on the People Search page to the newer style.

src/main/resources/org/sagebionetworks/web/client/view/SearchViewImpl.ui.xml
```diff
@@ -37,16 +37,18 @@
...
 					<div class="well">
 						<h5>Narrow Results</h5>
-						<bh:Italic>Searching for a user profile? Try our </bh:Italic>
-						<b:Anchor href="#!PeopleSearch:"
-							addStyleNames="margin-left-5 font-italic">People Search</b:Anchor>
-						<bh:Italic>.</bh:Italic>
 						<g:SimplePanel ui:field="facetPanel" />
 					</div>
 				</g:HTMLPanel>
+				<bh:Div addStyleNames="col-md-3 col-md-pull-9">
```
jay-hodgson, line 46: Pull out the PeopleSearch link from the Narrow Results panel.  Will be shown below facets if they are defined.

src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
nickgros: Is it always true that if this is not null, then it is also not empty?

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
nickgros: I'm assuming this is a list or some other collection

jay-hodgson: Based on testing, this was always null when empty.  But I added an empty check for good measure (and retested locally of course)

PR #4651, opened 2021-08-19
SWC-5680 - New toast system

nickgros:
Replace the old toast system with the new one in SRC. It even plays nicely with the new nav:

This addresses
* [SWC-5680](https://sagebionetworks.jira.com/browse/SWC-5680) - toast notifications in the new annotation editor. The logic for this is in SRC, but utilizes a new toast framework, which this PR supports
* [SWC-5740](https://sagebionetworks.jira.com/browse/SWC-5740) - prompt a confirmation before deleting a link

image


src/main/webapp/sass/_core.scss
```diff
@@ -134,7 +134,12 @@ a:focus{color:$synapse-blue;}
 }
 
 .SynapseNavDrawerIsShowing {
-	padding-left: 53px;
+    $-nav-drawer-width: 53px;
+	padding-left: $-nav-drawer-width;
+
+	.SynapseToastContainer {
+	    width: calc(100% - #{$-nav-drawer-width});
+	}
```
nickgros, line 142: Ensures the toast message doesn't cover the nav. Worked for me in FF/Chrome/Safari

nickgros, line 142: This ensures the toast message doesn't cover the nav. Worked for me in FF/Chrome/Safari

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -244,15 +265,7 @@ public static IsWidget getLoadingWidget(String message) {
...
 
 	public static void showSuccess(String message) {
-		showNotification(message, null, null, NotifyType.SUCCESS, IconType.CHECK_CIRCLE, null);
-	}
-
-	public static void showWarning(String message) {
-		showNotification(message, null, null, NotifyType.WARNING, IconType.WARNING, null);
-	}
-
-	public static void showDanger(String message) {
```
nickgros, line 254: These weren't being used

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -272,34 +285,29 @@ public static void showInfo(String message) {
...
 
-	public static void notify(String message, IconType iconType, NotifySettings settings) {
+	public static void notify(NotificationVariant variant, String title, String description, Long autoCloseInMs, String secondaryButtonText, String secondaryButtonHref) {
 		try {
-			String key = message + "/" + iconType.name();
+			String key = variant + "/" + title + "/" + description;
 			if (!recentNotificationMessages.contains(key)) {
 				recentNotificationMessages.add(key);
-				Notify.notify("", message, iconType, settings);
+				SRC.SynapseComponents.displayToast(variant.toString(), title, description, autoCloseInMs, null, null, secondaryButtonText, secondaryButtonHref);
```
nickgros, line 310: This is the key change--all SWC toasts now go through here

src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
```diff
@@ -18,6 +20,8 @@
 		public static ReactFunctionComponent<SchemaDrivenAnnotationEditorProps> SchemaDrivenAnnotationEditor;
 		public static ReactFunctionComponent<SynapseNavDrawerProps> SynapseNavDrawer;
 		public static ReactFunctionComponent<EmptyProps> FavoritesPage;
+		public static ReactFunctionComponent SynapseToastContainer;
+		public static native void displayToast(String variant, String title, String description, Long autoCloseInMs, String primaryButtonText, Callback onPrimaryButtonClick, String secondaryButtonText, String secondaryButtonHref);
```
nickgros, line 25: I'm not sure if I am ok with exporting a function `displayToast` from `SynapseComponents`, because it's not a component. But I also don't want to export a whole new object in SRC for one function

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadgeViewImpl.java
```diff
@@ -334,9 +334,7 @@ public void showUnlinkIcon() {
 		icon.setPull(Pull.RIGHT);
 		icon.setEmphasis(Emphasis.DANGER);
 		icon.addStyleName("imageButton");
-		icon.addClickHandler(event -> {
-			presenter.onUnlink();
-		});
+		icon.addClickHandler(event -> DisplayUtils.confirmDelete("Are you sure you want to delete this Link?", () -> presenter.onUnlink()));
```
nickgros, line 337: SWC-5740 - show a confirmation before unlinking an entity

I didn't realize we had `DisplayUtils.confirmDelete` until working on this. It makes this so easy, I just did it here.

src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
jay-hodgson: The signature should be pretty stable for this particular case (long history of showing these notifications), but I'm also weary in recommending this.

src/main/webapp/sass/_core.scss
jay-hodgson: Nice!  That explains why it is the correct width.  What makes it properly horizontally aligned?

jay-hodgson (approved):

src/main/webapp/sass/_core.scss
nickgros: The horizontal positioning is determined by its location in the DOM, (it's at the same level as the headerPanel, rootPanel, and footerPanel)--so whatever rules are positioning those properly are working here. If width is set to 100%, it's still properly left-aligned, but the right side of the toasts will go past the width of everything else.

PR #4652, opened 2021-08-19
SWC-5738

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/view/SearchViewImpl.java
```diff
@@ -377,22 +375,19 @@ private Panel getResult(int i, Hit hit) {
 		createdByBadge.configure(getSearchUserId(hit.getCreated_by()));
 		UserBadge modifiedByBadge = ginInjector.getUserBadgeWidget();
 		modifiedByBadge.configure(getSearchUserId(hit.getModified_by()));
-
+		attributionPanel.addStyleName("hitattribution");
```
jay-hodgson, line 378: Each part ("Created by X Modified By Y") used to have this css class name.  Now apply to entire attribution panel.  And change the style so that it doesn't make the text green, but rather italicized and with a proper margin to space out the hits.

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/schema/ColumnModelTableRowEditorWidgetTest.java
```diff
@@ -35,6 +35,7 @@
 	ColumnModelTableRowEditorWidgetImpl editor;
 	ColumnModel columnModel;
 	CellEditor mockStringEditor;
+	CellEditor mockStringListEditor;
```
jay-hodgson, line 38: New mock cell editor for StringList type column models

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/schema/ColumnModelTableRowEditorWidgetTest.java
```diff
@@ -78,6 +83,7 @@ public void testConfigure() {
 	public void testTypeChange() {
 		editor.configure(columnModel, null);
 		reset(mockView);
+		when(mockView.getMaxSize()).thenReturn("");
```
jay-hodgson, line 86: Since the mockView is fully reset, I need to re-initialize the default behavior when asking for the max size (in reality this is a text element in the dom, that will always return the empty string unless it contains a different value).

jay-hodgson:
Small change to search UI.
Before:
Screen Shot 2021-08-19 at 2 48 15 PM
After:
Screen Shot 2021-08-19 at 1 40 49 PM

xschildw (approved):

PR #4659, opened 2021-08-26
SWC-5748

nickgros:


package.json
```diff
@@ -16,7 +16,7 @@
     "react-transition-group": "2.6.0",
     "sanitize-html": "^1.18.4",
     "sass": "1.32.5",
-    "synapse-react-client": "2.0.31",
+    "synapse-react-client": "2.0.32",
```
nickgros, line 19: @jay-hodgson 
I need some CSS that didn't make it into 2.0.31. Are any of [these changes](https://github.com/Sage-Bionetworks/Synapse-React-Client/compare/v2.0.31...v2.0.32) inappropriate to add to staging? Looked fine to me

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -302,13 +303,14 @@ public static void notify(NotificationVariant variant, String title, String desc
...
+	public static void notify(NotificationVariant variant, String title, String description, int autoCloseInMs, String primaryButtonText, String primaryButtonHref) {
 		try {
 			String key = variant + "/" + title + "/" + description;
 			if (!recentNotificationMessages.contains(key)) {
 				recentNotificationMessages.add(key);
-				SRC.SynapseComponents.displayToast(variant.toString(), title, description, autoCloseInMs, null, null, secondaryButtonText, secondaryButtonHref);
-
+                SRC.SynapseComponents.displayToast(variant.toString(), title, description, autoCloseInMs, primaryButtonText, () -> {
+                    Window.Location.assign(primaryButtonHref);
+                }, null, null);
```
nickgros, line 313: Switch to use primary button rather than secondary. We use Window.Location.assign b/c I think we always want to open in the current window.

src/main/java/org/sagebionetworks/web/client/PopupUtilsViewImpl.java
```diff
@@ -18,7 +18,7 @@ public void showInfo(String message, Integer timeout) {
 
 	@Override
 	public void showInfo(String message, String href, String buttonText) {
-		DisplayUtils.showInfo(message, href, buttonText, 0);
+		DisplayUtils.showInfo(message, href, buttonText, 15000);
```
nickgros, line 21: This was causing some notifications to persist. Next I would really like to pare down to just a couple of ways to trigger toasts rather than having 25 methods across 3 classes that are all slightly different 🙂

src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
```diff
@@ -1,7 +1,5 @@
 package org.sagebionetworks.web.client.jsinterop;
 
-import org.sagebionetworks.web.client.utils.Callback;
-
```
nickgros, line 4: This Callback interface doesn't work in JsInterop, which caused errors when clicking the primary button

src/main/webapp/sass/_core.scss
```diff
@@ -3516,4 +3516,12 @@ div.border {
...
 	z-index: 1050;
 	background-color: rgba(0, 0, 0, 0.5);
+}
+
+// We may want to move this to SRC and apply this style to all sites
+// May also want to apply the style to all buttons, rather than just in this alert
+.FullWidthAlert {
+    button.btn {
+        text-transform: capitalize;
+    }
```
nickgros, line 3526: e.g. one string is `"view download list"` -- I'm not sure if we want to really style ALL primary buttons with CSS, but this is really a stopgap for these toasts because I don't want to try to fix each string that could get passed in

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
nickgros: image

jay-hodgson (approved):

PR #4663, opened 2021-08-27
SWC-5747 Account for displayToast signature change, some cleanup

nickgros:
Will need to get the latest (currently unreleased) version of SRC for this to work

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
@@ -266,7 +266,12 @@ public class DisplayUtils {
 	 */
 	public static void notify(String message, NotificationVariant variant, ToastMessageOptions options) {
 		try {
-			String key = variant + "/" + options.title + "/" + message;
+			String title = null;
+			if (options != null) {
+				title = options.title;
+			}
+
+			String key = variant + "/" + title + "/" + message;
 			if (!recentNotificationMessages.contains(key)) {
 				recentNotificationMessages.add(key);
                 SRC.SynapseComponents.displayToast(message, variant.toString(), options);
--- a/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
@@ -26,12 +26,9 @@ import org.gwtbootstrap3.client.ui.constants.Placement;
 import org.gwtbootstrap3.client.ui.constants.Trigger;
 import org.gwtbootstrap3.client.ui.html.Div;
 import org.gwtbootstrap3.client.ui.html.Text;
-import org.gwtbootstrap3.extras.animate.client.ui.constants.Animation;
 import org.gwtbootstrap3.extras.bootbox.client.Bootbox;
 import org.gwtbootstrap3.extras.bootbox.client.callback.SimpleCallback;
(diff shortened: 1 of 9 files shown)


src/main/java/org/sagebionetworks/web/client/PopupUtilsViewImpl.java
```diff
@@ -1,7 +1,8 @@
 package org.sagebionetworks.web.client;
 
-import org.gwtbootstrap3.client.ui.constants.IconType;
+import org.sagebionetworks.web.client.jsinterop.ToastMessageOptions;
 import org.sagebionetworks.web.client.utils.Callback;
+
 import com.google.gwt.user.client.Window;
 
 public class PopupUtilsViewImpl implements PopupUtilsView {
```
nickgros, line 8: What should we do with this class? Mostly an instance wrapper over static methods in DisplayUtils. Do we expect to have another implementation of this at some point?

src/main/java/org/sagebionetworks/web/client/jsinterop/ToastMessageOptions.java
```diff
@@ -0,0 +1,116 @@
...
+
+        ToastMessageOptions options = new ToastMessageOptions();
+        options.title = title;
+        options.autoCloseInMs = autoCloseInMs;
+        options.primaryButtonText = primaryButtonText;
+        options.onPrimaryButtonClick = onPrimaryButtonClick;
+        return options;
+    }
+
+    public static class Builder {
```
nickgros, line 65: Sorry, I'm obsessed w/ builder pattern.

I think this gives us the same benefit in Java as simplifying the `displayToast` signature does in JavaScript--no need to have code for setting seldom-used parameters to default/null when creating a toast

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -276,41 +247,34 @@ public static void showSuccess(String message) {
...
-	 * @param message
-	 */
-	public static void showInfo(String message, Integer timeout) {
-		showInfo(message, null, null, timeout);
-	}
-
-	public static void showInfo(String message, String href, String buttonText, Integer timeout) {
-		notify(NotificationVariant.INFO, null, message, timeout, buttonText, href);
+		notify(message, NotificationVariant.INFO);
 	}
```
nickgros, line 251: Pared down to just one `showInfo` method in DisplayUtils. If you need something more complicated, you can call `notify` directly.

src/main/java/org/sagebionetworks/web/client/jsinterop/ToastMessageOptions.java
jay-hodgson: I think this is an excellent fit for the builder pattern.

src/main/java/org/sagebionetworks/web/client/PopupUtilsViewImpl.java
jay-hodgson: I think it provides the necessary abstraction, since presenters cannot directly use the view js code in DisplayUtils.  Perhaps a more logical change would be to add a DisplayUtilsView interface, and switch all references from PopupUtilsView to DisplayUtilsView.

jay-hodgson (approved):

PR #4678, opened 2021-09-13
SWC-5774: move account level info below form

jay-hodgson:


jay-hodgson:
(merging to start new staging build before Tuesday, but will follow up with any necessary code changes)

PR #4680, opened 2021-09-14
SWC-5754: hide project UI and tabs if project entity cannot be loaded

jay-hodgson:
I think this will be better than an infinite spinner, a non-functional project tools menu, and tabs that don't work right.


jay-hodgson:
Screen Shot 2021-09-14 at 11 16 37 AM

nickgros (approved):

PR #4682, opened 2021-09-15
SWC-5771, SWC-5777: attempt to fix sizing bugs

jay-hodgson:


xschildw (approved):
I'll take your word for it...

PR #4691, opened 2021-10-07
SWC-5808: queue a check to see if the currently selected tab is hidden.

jay-hodgson:
If so, select another tab that has content.  In the following order - Wiki, Files, Tables, Docker, Discussion

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
@@ -509,7 +509,7 @@ public class GlobalApplicationStateImpl implements GlobalApplicationState {
 		view.initSRCEndpoints(repoUrl, portalUrl);
 
 		// Add a global click handler.
-		Event.addNativePreviewHandler(new Event.NativePreviewHandler() {
+		view.addNativePreviewHandler(new Event.NativePreviewHandler() {
 			public void onPreviewNativeEvent(NativePreviewEvent event) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -504,6 +508,17 @@ public void init(final Callback finalCallback) {
 		}
 		view.initSRCEndpoints(repoUrl, portalUrl);
 
+		// Add a global click handler.
```
jay-hodgson, line 511: See https://sagebionetworks.jira.com/browse/SWC-5812

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -467,6 +504,7 @@ private void fileChanged(EntityBundle bundle, Long version) {
 
 	public void showSelectedTabs() {
 		visibleTabCount = 0;
+		countTabContentChecked = 0;
```
jay-hodgson, line 507: This is where we start to determine what tabs are visible.  count is incremented by each callback.  On the last callback (when countTabContentChecked == EntityArea.values().length), we check to see if the currently selected area tab is visible.  If not, change the area to a visible area and reconfigure.

xschildw: Got it...

xschildw (approved):

PR #4699, opened 2021-10-12
SWC-5763: Add datasets tab and support datasets across shared table/view widgets

nickgros:
Tab is only visible in Experimental Mode

Tab w/ list:
image

Entity Page: 
image

Custom area:
image

Creating a new dataset:

https://user-images.githubusercontent.com/17580037/137015457-0b2da2c4-36ad-4095-9a13-bf1eb93b010a.mov

Modifying the scope:

https://user-images.githubusercontent.com/17580037/137017354-3587c2bc-ea4f-404d-a29e-da496f7c4ab9.mov


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -440,6 +441,8 @@ private void reconfigureActions() {
 		configureACTCommands();
 		configureTableCommands();
 		configureProjectLevelTableCommands();
+		// TODO: configureDatasetCommands()
```
nickgros, line 444: configureTableCommands works for now, so not messing with this yet

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/DatasetsTab.java
```diff
@@ -0,0 +1,268 @@
...
+import org.sagebionetworks.web.client.widget.table.v2.QueryTokenProvider;
+import org.sagebionetworks.web.client.widget.table.v2.TableEntityWidget;
+import org.sagebionetworks.web.client.widget.table.v2.results.QueryBundleUtils;
+import org.sagebionetworks.web.shared.WebConstants;
+import org.sagebionetworks.web.shared.WidgetConstants;
+
+import com.google.gwt.place.shared.Place;
+import com.google.inject.Inject;
+
+public class DatasetsTab implements DatasetsTabView.Presenter, QueryChangeHandler {
```
nickgros: Completely copy-pasted from TablesTab and then pretty much replaced all references of "Table" to "Dataset". Is there a better way to do this to avoid duplicate code?

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityContainerListWidget.java
```diff
@@ -109,7 +127,8 @@ public void onAddProject() {
 	 * 
 	 * @param id
 	 */
-	public void onAddProject(String id) {
+	public void onAddEntity(String id) {
+		// TODO: Support versions
```
nickgros, line 130: This widget only partially supports versions after these changes. It won't let you add two versions of the same entity, and the view doesn't show you a version.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityViewScopeWidget.java
```diff
@@ -60,19 +68,56 @@ public EntityViewScopeWidget(EntityViewScopeWidgetView view, SynapseJavascriptCl
...
+			viewScopeWidget.configure(references, false, tableType);
 			view.setEditButtonVisible(isEditable && tableType != null);
-			if (tableType == null) {
-				synAlert.consoleError("View type mask is not supported by web client, blocking edit." + currentView.getViewTypeMask());
-			}
+
 		}
+
 		view.setVisible(isVisible);
 	}
```
nickgros, line 120: Scope/dataset items basically work the same, but we access the items differently.

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -513,12 +553,18 @@ public void showSelectedTabs() {
...
+			}
 			getTabVisibilityCallback(EntityArea.TABLES, tablesTab.asTab()).onSuccess(true);
 			getTabVisibilityCallback(EntityArea.DOCKER, dockerTab.asTab()).onSuccess(true);
 		} else {
 			// otherwise only show the tabs only if content is present.
 			synapseJavascriptClient.isWiki(projectHeader.getId(), getTabVisibilityCallback(EntityArea.WIKI, wikiTab.asTab()));
 			synapseJavascriptClient.isFileOrFolder(projectHeader.getId(), getTabVisibilityCallback(EntityArea.FILES, filesTab.asTab()));
+			if (DisplayUtils.isInTestWebsite(cookies)) {
+				synapseJavascriptClient.isDataset(projectHeader.getId(), getTabVisibilityCallback(EntityArea.DATASETS, datasetsTab.asTab()));
+			}
```
nickgros, line 567: The visibility of the tab is the only thing gated by experimental mode

src/main/java/org/sagebionetworks/web/client/widget/table/TableListWidget.java
```diff
@@ -64,12 +65,13 @@ public void invoke() {
...
-	public void configure(EntityBundle parentBundle) {
+	public void configure(EntityBundle parentBundle, List<EntityType> typesToShow) {
 		if (currentRequest != null) {
 			currentRequest.cancel();
 		}
 		isInitializing = true;
 		this.parentBundle = parentBundle;
+		this.typesToShow = typesToShow;
 		loadData();
 	}
```
nickgros, line 76: The list of tables is now configurable so it can show a list of datasets, or a list of (table, entityview, submissionview), or however else we may configure it.

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/DatasetsTab.java
jay-hodgson: This could be a good candidate for a shared abstract class (where you would only see the differentiators in this non-abstract subclass).

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/DatasetsTabViewImpl.java
```diff
@@ -0,0 +1,137 @@
+package org.sagebionetworks.web.client.widget.entity.tabs;
```
jay-hodgson: If it's the same UI as the tables tab, it could at least share the same View.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityContainerListWidget.java
```diff
@@ -132,14 +154,27 @@ public void onSuccess(ArrayList<EntityHeader> entityHeaders) {
 	}
 
 	@Override
-	public void onRemoveProject(String id) {
-		entityIds.remove(id);
+	public void onRemoveEntity(String id) {
+		for (int i = 0; i < references.size(); i++) {
+			if (references.get(i).getTargetId().equals(id)) {
+				references.remove(i);
+				i--;
```
jay-hodgson: This works.  I think another option would be to iterate through the list backwards (from size-1 to 0) to avoid the correction to the index after removal.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityViewScopeWidget.java
```diff
@@ -113,9 +170,16 @@ public void onFailure(Throwable caught) {
 
 	@Override
 	public void onEdit() {
+		List<Reference> references = new ArrayList<>();
 		// configure edit list, and show modal
-		editScopeWidget.configure(currentView.getScopeIds(), true, tableType);
-		if (TableType.table.equals(tableType) || TableType.projects.equals(tableType)) {
+		if (currentView instanceof EntityView) {
+            references = getReferencesFromIdList(((EntityView) currentView).getScopeIds());
```
jay-hodgson: Few places where whitespace seems to be inconstant with the current code.

jay-hodgson (approved):

PR #4704, opened 2021-11-02
SWC-5836

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/SWC-5836)


src/main/resources/org/sagebionetworks/web/client/widget/entity/tabs/DatasetsTabViewImpl.ui.xml
```diff
@@ -1,38 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
```
nickgros, line 1: We're using one view for Tables and Datasets, this was unused and didn't get deleted.

jay-hodgson (approved):

PR #4712, opened 2021-11-15
SWC-5623: set container that should be used for tooltip to select the…

jay-hodgson:
… body element

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
@@ -622,6 +622,7 @@ public class DisplayUtils {
 		t.setIsHtml(true);
 		t.setIsAnimated(false);
 		t.setTrigger(Trigger.HOVER);
+		t.setContainer("body");
 		t.setWidget(widget);
 		return t;
 	}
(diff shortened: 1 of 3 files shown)


jay-hodgson:
Initially I thought it would be great to reuse @nickgros 's EntityBadgeIcons from SRC.  Plan to save that for the next step, when files browser is replaced by SRC.

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -622,6 +622,7 @@ public static Tooltip addTooltip(Widget widget, String tooltipText, Placement po
 		t.setIsHtml(true);
 		t.setIsAnimated(false);
 		t.setTrigger(Trigger.HOVER);
+		t.setContainer("body");
```
nickgros, line 625: This is some Bootstrap magic, right?

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
jay-hodgson: Yes

jay-hodgson: I think the default behavior is to attach the new (absolute positioned) tooltip element as a sibling to the target element, but this allows you to specify another element.

PR #4715, opened 2021-11-19
SWC-5849: pass in sso redirect URL that can be used for any of our su…

jay-hodgson:
…pport Oauth providers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "react-transition-group": "2.6.0",
     "sanitize-html": "^1.18.4",
     "sass": "1.32.5",
-    "synapse-react-client": "2.0.51",
+    "synapse-react-client": "2.0.56",
     "universal-cookie": "^4.0.3"
   }
 }
--- a/src/main/java/org/sagebionetworks/web/client/view/users/RegisterAccountViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/view/users/RegisterAccountViewImpl.java
@@ -55,8 +55,8 @@ public class RegisterAccountViewImpl extends Composite implements RegisterAccoun
 	Presenter presenter;
 	private Header headerWidget;
 	public static final String ROOT_PORTAL_URL = Window.Location.getProtocol() + "//" + Window.Location.getHost() + "/";
-	public static final String GOOGLE_OAUTH_CALLBACK_URL = RegisterAccountViewImpl.ROOT_PORTAL_URL + "Portal/oauth2callback?oauth2provider=GOOGLE_OAUTH_2_0";
-	public static final String GOOGLE_OAUTH_WITH_STATE_CALLBACK_URL = GOOGLE_OAUTH_CALLBACK_URL + "&state=";
+	public static final String OAUTH_CALLBACK_URL = RegisterAccountViewImpl.ROOT_PORTAL_URL + "Portal/oauth2callback?oauth2provider=";
+	public static final String GOOGLE_OAUTH_WITH_STATE_CALLBACK_URL = OAUTH_CALLBACK_URL + "GOOGLE_OAUTH_2_0&state=";
 
 	@Inject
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "react-transition-group": "2.6.0",
     "sanitize-html": "^1.18.4",
     "sass": "1.32.5",
-    "synapse-react-client": "2.0.51",
+    "synapse-react-client": "2.0.56",
```
jay-hodgson, line 19: Update to the latest SRC.

src/main/java/org/sagebionetworks/web/client/widget/login/LoginWidgetViewImpl.java
```diff
@@ -54,14 +54,14 @@ public void postLogin() {
...
-	   String googleSSORedirectUrl, SynapseContextProviderPropsJSNIObject wrapperProps) /*-{
+	   String ssoRedirectUrl, SynapseContextProviderPropsJSNIObject wrapperProps) /*-{
 		try {
 			function sessionCallback() {
 				loginWidgetView.@org.sagebionetworks.web.client.widget.login.LoginWidgetViewImpl::postLogin()();
 			}
 
 			var props = {
-				googleRedirectUrl : googleSSORedirectUrl,
+				ssoRedirectUrl : ssoRedirectUrl,
```
jay-hodgson, line 64: In the latest version of SRC, the LoginPage component has a prop named ssoRedirectUrl, and it will append the appropriate OAuth provider string based on what provider the user selects to log in with.

src/main/java/org/sagebionetworks/web/client/view/users/RegisterAccountViewImpl.java
```diff
@@ -55,8 +55,8 @@
 	Presenter presenter;
 	private Header headerWidget;
 	public static final String ROOT_PORTAL_URL = Window.Location.getProtocol() + "//" + Window.Location.getHost() + "/";
-	public static final String GOOGLE_OAUTH_CALLBACK_URL = RegisterAccountViewImpl.ROOT_PORTAL_URL + "Portal/oauth2callback?oauth2provider=GOOGLE_OAUTH_2_0";
-	public static final String GOOGLE_OAUTH_WITH_STATE_CALLBACK_URL = GOOGLE_OAUTH_CALLBACK_URL + "&state=";
+	public static final String OAUTH_CALLBACK_URL = RegisterAccountViewImpl.ROOT_PORTAL_URL + "Portal/oauth2callback?oauth2provider=";
```
jay-hodgson, line 58: Rename, and do not hardcode GOOGLE_OAUTH_2_0 as the provider

src/main/java/org/sagebionetworks/web/client/view/users/RegisterAccountViewImpl.java
```diff
@@ -55,8 +55,8 @@
 	Presenter presenter;
 	private Header headerWidget;
 	public static final String ROOT_PORTAL_URL = Window.Location.getProtocol() + "//" + Window.Location.getHost() + "/";
-	public static final String GOOGLE_OAUTH_CALLBACK_URL = RegisterAccountViewImpl.ROOT_PORTAL_URL + "Portal/oauth2callback?oauth2provider=GOOGLE_OAUTH_2_0";
-	public static final String GOOGLE_OAUTH_WITH_STATE_CALLBACK_URL = GOOGLE_OAUTH_CALLBACK_URL + "&state=";
+	public static final String OAUTH_CALLBACK_URL = RegisterAccountViewImpl.ROOT_PORTAL_URL + "Portal/oauth2callback?oauth2provider=";
+	public static final String GOOGLE_OAUTH_WITH_STATE_CALLBACK_URL = OAUTH_CALLBACK_URL + "GOOGLE_OAUTH_2_0&state=";
```
jay-hodgson, line 59: During registration, however, we currently only support Google-based user creation.  If we will support others, the registration widget (in SWC) will need to be updated, and this would also need to be changed.  We may want to reimplement the registration widget as a react component at that time.

xschildw (approved):

PR #4716, opened 2021-11-22
SWC-5820: add a Table type column

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
```diff
@@ -115,4 +121,18 @@ public static IconType getIconTypeForEntityClassName(String className) {
 
 		return icon;
 	}
+	
+	public static String getFriendlyTableTypeName(String className) {
+		String friendlyName = UNKNOWN_TABLE_TYPE;
```
jay-hodgson, line 126: When a new type is added but unrecognized, then it will initially be reported as an "Unknown Table Type" in the UI, until it is recognized here (Materialized Views is the upcoming example).

xschildw (approved):

PR #4717, opened 2021-11-23
SWC-5784

nickgros:
https://user-images.githubusercontent.com/17580037/143269805-04b75edc-54cd-45e5-9e35-f97c8528a598.mov


src/main/java/org/sagebionetworks/web/client/jsinterop/EntityFinderProps.java
```diff
@@ -52,7 +59,7 @@ public static EntityFinderProps create(
             List<EntityType> visibleTypesInList,
             List<EntityType> visibleTypesInTree,
             List<EntityType> selectableTypes,
-            String selectedCopy,
+            SelectedCopyHandler selectedCopy,
```
nickgros, line 62: This is the copy in the entity finder that says "Selected" or "Destination". We can now pass a `fn(int selectedCount)` to say "4 Items Selected"

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -589,4 +604,23 @@ public void onAddToDownloadList() {
...
+	@Override
+	public DatasetEditorProps getItemsEditorProps() {
+		DatasetEditorProps props =
+				DatasetEditorProps.create(
+						entityBundle.getEntity().getId(),
+						() -> closeItemsEditor(),
+						() -> closeItemsEditor()
+				);
+		return props;
+	}
```
nickgros, line 617: Presenter added so the view can get React component props from the model.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -589,4 +604,23 @@ public void onAddToDownloadList() {
...
+
+	public void closeItemsEditor() {
+		isShowingItemsEditor = false;
+		actionMenu.setActionVisible(Action.EDIT_DATASET_ITEMS, true);
+		view.setItemsEditorVisible(false);
+		view.setQueryResultsVisible(true);
+		reconfigureState();
```
nickgros, line 624: The only thing that *needs* to happen is calling `checkState` again, but I think it's safer to just capture the three methods called in the configure and call all of them.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsWidget.java
```diff
@@ -105,11 +105,11 @@ public void configure(EntityBundle bundle, boolean isEditable) {
...
 		viewer.configure(ViewType.VIEWER, this.isEditable);
-		// Certain rows in datasets may not be editable because they reference locked versions.
-		boolean isEditableView = isEditable && (bundle.getEntity() instanceof EntityView || bundle.getEntity() instanceof SubmissionView);
+		// We can get the default columns/annotations for views and datasets
+		boolean isEditableViewOrDataset = isEditable && (bundle.getEntity() instanceof EntityView || bundle.getEntity() instanceof SubmissionView || bundle.getEntity() instanceof Dataset);
 		tableType = TableType.getTableType(bundle.getEntity());
-		editor.setAddDefaultViewColumnsButtonVisible(isEditableView);
-		editor.setAddAnnotationColumnsButtonVisible(isEditableView);
+		editor.setAddDefaultViewColumnsButtonVisible(isEditableViewOrDataset);
+		editor.setAddAnnotationColumnsButtonVisible(isEditableViewOrDataset);
```
nickgros, line 112: Addresses [SWC-5857](https://sagebionetworks.jira.com/browse/SWC-5857) -- buttons to add default columns or annotation columns were missing for datasets. I think in my first pass at this, I thought this was the query results editor, so the comment I left was irrelevant.

jay-hodgson (approved):

PR #4721, opened 2021-12-06
SWC-5851

nickgros:
https://sagebionetworks.jira.com/browse/SWC-5851

No tests since these are only view changes, should I add any?

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/VersionHistoryWidgetViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/VersionHistoryWidgetViewImpl.java
@@ -1,5 +1,8 @@
 package org.sagebionetworks.web.client.widget.entity;
 
+import static org.sagebionetworks.web.client.DisplayConstants.GO_TO_CURRENT_VERSION;
+import static org.sagebionetworks.web.client.DisplayConstants.GO_TO_DRAFT_VERSION;
+
 import java.util.ArrayList;
 import java.util.List;
 import org.gwtbootstrap3.client.ui.Button;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileTitleBarViewImpl.java
```diff
@@ -128,6 +133,15 @@ public void createTitlebar(Entity entity) {
 		favoriteWidget.configure(currentEntityId);
 		md5Link.clear();
 		md5LinkContainer.setWidget(md5Link);
+
+    String currentVersionLinkText;
```
nickgros:
```suggestion
   	String currentVersionLinkText;
```

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileTitleBarViewImpl.java
```diff
@@ -128,6 +133,15 @@ public void createTitlebar(Entity entity) {
 		favoriteWidget.configure(currentEntityId);
 		md5Link.clear();
 		md5LinkContainer.setWidget(md5Link);
+
+   	String currentVersionLinkText;
```
nickgros:
```suggestion
		String currentVersionLinkText;
```

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/VersionHistoryWidgetViewImpl.java
```diff
@@ -94,6 +98,13 @@ public void setEntityBundle(Entity entity, boolean isShowingOlderVersion) {
 		md5TableHeader.setVisible(!isTable);
 		currentVersionLink.setTargetHistoryToken(DisplayUtils.getSynapseHistoryTokenNoHash(entity.getId()));
 		currentVersionLink.setVisible(isShowingOlderVersion);
+		String currentVersionLinkText;
```
jay-hodgson, line 101: May want a shared currentVersionLinkText(Entity entity) method instead of the copy

PR #4724, opened 2021-12-08
SWC-5750

nickgros:


src/main/webapp/Portal.html
```diff
@@ -114,26 +128,19 @@
...
-				head.load(cdnEndpoint + "generated/react-dom.development.js");
-				head.load(cdnEndpoint + "generated/react-query.development.js");
-			}
+			head.load(reactPath);
+			head.load(reactDomPath);
+			head.load(reactQueryPath);
 			head.load(cdnEndpoint + "generated/react-transition-group.min.js");
 			head.load(cdnEndpoint + "generated/prop-types.min.js");
 			head.load(cdnEndpoint + "generated/react-measure.js");
-			head.load(cdnEndpoint + "generated/react-tooltip.min.js");
```
nickgros, line 129: In SRC we're using react tooltip 4 and react tooltip 4 doesn't seem to distribute a UMD build, so we're bundling it with SRC.

package.json
```diff
@@ -12,10 +12,9 @@
     "react-dom": "16.13.1",
     "react-measure": "2.2.2",
     "react-query": "^3.17.0",
-    "react-tooltip": "3.11.2",
```
nickgros, line 15: In SRC we're using React Tooltip 4, which doesn't seem to have a distributed UMD build (unlike this version). It's now bundled with SRC, so removing it here.

PR #4731, opened 2021-12-13
SWC-5897: Check the storage size every 5 minutes (lots of boilerplate…

jay-hodgson:
… code)


src/main/java/org/sagebionetworks/web/client/cache/StorageImpl.java
```diff
@@ -45,4 +45,30 @@ public void setItem(String key, String data) {
...
 		return storage != null;
 	}
+	
+	@Override
+	public double getBytesUsed() {
+		return _getBytesUsed();
+	}
+	
+	private static native double _getBytesUsed() /*-{
+		// Based on https://stackoverflow.com/questions/4391575/how-to-find-the-size-of-localstorage
```
jay-hodgson, line 55: Here's the interesting code, but it's in js.

xschildw (approved):

PR #4746, opened 2022-01-03
SWC-5903

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/SWC-5903)

A successful update to a Dataset would reveal the edit button, which should never be visible for Datasets.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWidget.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWidget.java
@@ -306,6 +306,11 @@ public class TableQueryResultWidget implements TableQueryResultView.Presenter, I
 	 * @return
 	 */
 	public boolean isQueryResultEditable() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWidget.java
```diff
@@ -306,6 +306,11 @@ private void setQueryResultsAndSort(QueryResultBundle bundle, List<SortItem> sor
 	 * @return
 	 */
 	public boolean isQueryResultEditable() {
```
nickgros, line 308: The result of this method is passed into TableEntityWidget.queryExecutionFinished. Line 461 triggers visibility of the edit button:

 https://github.com/Sage-Bionetworks/SynapseWebClient/blob/a8d2a3386ddc871e6440975c711d14eb6d823211/src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java#L457-L483

jay-hodgson (approved):

PR #4761, opened 2022-01-18
SWC-5928 - If only one ActionMenu Action is visible, replace the dropdown with a button

nickgros:
[Jira](https://sagebionetworks.jira.com/browse/SWC-5928)

Loading Skeleton:

https://user-images.githubusercontent.com/17580037/150004701-c45a3f3d-284b-4e89-9128-ec8715ee3e6e.mov

Ignore that the font doesn't load first; I had to throttle my connection in the browser to simulate the loading

Single action:
image

* Add two new states to the action menu
	* Loading, where a skeleton button is shown
	* One visible action, where the dropdown is replaced with a single button with the sole visible action
* Requires exposing the React SkeletonButton from SRC, which has not been done

This is in Draft mode because an upgrade to SRC where SkeletonButton is exposed in the UMD build is required.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
+++ b/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
@@ -901,6 +901,11 @@ public class SynapseJavascriptClient {
 		doGet(url, OBJECT_TYPE.ChallengePagedResults, callback);
 	}
 
+	public FluentFuture<Challenge> getChallengeForProject(String projectId) {
+		String url = getRepoServiceUrl() + ENTITY + "/" + projectId + "/" + CHALLENGE;
+		return getFuture(cb -> doGet(url, OBJECT_TYPE.Challenge, cb));
+	}
+
 	public void getUserBundle(Long principalId, int mask, AsyncCallback<UserBundle> callback) {
 		String url = getRepoServiceUrl() + USER + "/" + principalId + BUNDLE_MASK_PATH + mask;
 		doGet(url, OBJECT_TYPE.UserBundle, callback);
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
@@ -25,6 +25,7 @@ public class SRC {
 		public static ReactFunctionComponent<IconSvgProps> IconSvg;
 		public static ReactFunctionComponent<EntityTypeIconProps> EntityTypeIcon;
 		public static ReactFunctionComponent<UserProfileLinksProps> UserProfileLinks;
+		public static ReactFunctionComponent<SkeletonButtonProps> SkeletonButton;
 		public static ReactFunctionComponent SynapseToastContainer;
 
 		/**
--- a/src/main/java/org/sagebionetworks/web/client/widget/asynch/IsACTMemberAsyncHandlerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/asynch/IsACTMemberAsyncHandlerImpl.java
@@ -1,13 +1,21 @@
 package org.sagebionetworks.web.client.widget.asynch;
 
+import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static org.sagebionetworks.web.client.ServiceEntryPointUtils.fixServiceEntryPoint;
 import static org.sagebionetworks.web.client.presenter.ProfilePresenter.IS_ACT_MEMBER;
+import static org.sagebionetworks.web.client.utils.FutureUtils.getDoneFuture;
+import static org.sagebionetworks.web.client.utils.FutureUtils.getFuture;
+
+import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 import org.sagebionetworks.repo.model.UserBundle;
 import org.sagebionetworks.web.client.SynapseJSNIUtils;
 import org.sagebionetworks.web.client.UserProfileClientAsync;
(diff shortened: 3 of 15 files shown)


src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -901,6 +901,11 @@ public void getChallenges(String userId, Integer limit, Integer offset, AsyncCal
 		doGet(url, OBJECT_TYPE.ChallengePagedResults, callback);
 	}
 
+	public FluentFuture<Challenge> getChallengeForProject(String projectId) {
+		String url = getRepoServiceUrl() + ENTITY + "/" + projectId + "/" + CHALLENGE;
+		return getFuture(cb -> doGet(url, OBJECT_TYPE.Challenge, cb));
+	}
```
nickgros, line 907: I was going to try put this method in the ChallengeClientAsync, but I think that implementation is generated.

src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
```diff
@@ -25,6 +25,7 @@
 		public static ReactFunctionComponent<IconSvgProps> IconSvg;
 		public static ReactFunctionComponent<EntityTypeIconProps> EntityTypeIcon;
 		public static ReactFunctionComponent<UserProfileLinksProps> UserProfileLinks;
+		public static ReactFunctionComponent<SkeletonButtonProps> SkeletonButton;
```
nickgros, line 28: Should we expose this in SRC?

src/main/java/org/sagebionetworks/web/client/widget/asynch/IsACTMemberAsyncHandlerImpl.java
```diff
@@ -46,35 +54,57 @@ public void invoke(Boolean isACTMember) {
...
+			}
+
+			@Override
+			public void onFailure(Throwable caught) {
+				// log the error, and tell client that user is not part of the ACT (give the developer something to
+				// look at if something goes wrong for an ACT member)
+				jsniUtils.consoleError(caught.getMessage());
+				callback.invoke(false);
+
+			}
```
nickgros, line 77: Same failure scenario as below. As of now, this should never happen because it calls a Future that can never fail.

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -438,9 +440,8 @@ public void configure(ActionMenuWidget actionMenu, EntityBundle entityBundle, bo
 		this.entityTypeDisplay = getFriendlyEntityTypeName(entityBundle.getEntity());
 		this.currentArea = currentArea;
 
-		// hide all commands by default
-		actionMenu.hideAllActions();
-		gwt.scheduleExecution(reconfigureActionsCallback, 2000);
```
nickgros, line 443: @jay-hodgson Do you know the original purpose of this delay?

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/controller/EntityActionControllerImplTest.java
```diff
@@ -1980,32 +1978,34 @@ public void testNotFolderDeletionPrompt() {
 	}
 
 	@Test
-	public void testConfigureManageAccessRequirements() {
```
nickgros, line 1983: The original test was unclear to me, so I split this into two new tests where I assume we were testing the behavior for an ACT member and a non-ACT member

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
jay-hodgson: ChallengeClientAsync is part of one of the older GWT RPC servlet endpoints (see http://www.gwtproject.org/doc/latest/DevGuideServerCommunication.html for more info).  Perhaps surprisingly, this was not generated.
I prefer to now go directly to the backend via this SynapseJavascriptClient.

src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
jay-hodgson: You mean it's available in the UMD build but not to portals?  If that's the case, I wouldn't worry about it (we can export once we have a case that needs it)

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
jay-hodgson: No idea.  It was a change I made 4 years ago alongside a number of performance improvements:
https://github.com/Sage-Bionetworks/SynapseWebClient/commit/6fe87966e115028de586eb9002f96327f51191b8

I'm very please if it can be deleted!

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/controller/EntityActionControllerImplTest.java
jay-hodgson: I think the test names need to be revisited.  It looks like the first test is when isACTActionAvailable is false (currently named testConfigureManageAccessRequirementsAsACTMember), and the second is when the ACT Actions are available (currently named testConfigureManageAccessRequirementsAsACTMemberAsNonACTMember)

jay-hodgson (approved):
Changes look great, assuming SRC dependency is updated

PR #4762, opened 2022-01-25
SWC-5929, SWC-5930, SWC-5931 - Toast messages

nickgros:
- SWC-5929: Show toast when saving a dataset
- SWC-5930: Show toast when updating a table schema
- SWC-5931: Show toast when creating/updating a DOI

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
@@ -256,6 +256,10 @@ public class DisplayUtils {
 		notify(message, variant, null);
 	}
 
+	public static void notify(String title, String message, NotificationVariant variant) {
+		notify(message, variant, new ToastMessageOptions.Builder().setTitle(title).build());
+	}
+
 	/**
 	 * Pushes a toast by calling SRC's displayToast method.
 	 *
--- a/src/main/java/org/sagebionetworks/web/client/PopupUtilsView.java
+++ b/src/main/java/org/sagebionetworks/web/client/PopupUtilsView.java
@@ -10,6 +10,8 @@ import org.sagebionetworks.web.client.utils.Callback;
 public interface PopupUtilsView {
 	void notify(String message, DisplayUtils.NotificationVariant variant, ToastMessageOptions options);
 
+	void notify(String title, String message, DisplayUtils.NotificationVariant notificationVariant);
+
 	void showInfo(String message);
 
 	void showInfo(String message, Integer timeout);
--- a/src/main/java/org/sagebionetworks/web/client/widget/doi/CreateOrUpdateDoiModal.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/doi/CreateOrUpdateDoiModal.java
@@ -23,6 +23,7 @@ import org.sagebionetworks.repo.model.doi.v2.DoiResourceType;
 import org.sagebionetworks.repo.model.doi.v2.DoiResourceTypeGeneral;
 import org.sagebionetworks.repo.model.doi.v2.DoiTitle;
 import org.sagebionetworks.web.client.DateTimeUtils;
+import org.sagebionetworks.web.client.DisplayUtils;
 import org.sagebionetworks.web.client.EntityTypeUtils;
 import org.sagebionetworks.web.client.PopupUtilsView;
 import org.sagebionetworks.web.client.SynapseJavascriptClient;
@@ -55,6 +56,8 @@ public class CreateOrUpdateDoiModal implements CreateOrUpdateDoiModalView.Presen
 	private PopupUtilsView popupUtilsView;
 	private DateTimeUtils dateTimeUtils;
 
+	private boolean doiExists = false;
+
 	@Inject
 	public CreateOrUpdateDoiModal(CreateOrUpdateDoiModalView view, JobTrackingWidget jobTrackingWidget, SynapseJavascriptClient javascriptClient, SynapseAlert synapseAlert, PopupUtilsView popupUtilsView, EventBus eventBus, DateTimeUtils dateTimeUtils) {
 		this.view = view;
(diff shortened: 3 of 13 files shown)


src/test/java/org/sagebionetworks/web/unitclient/widget/doi/CreateOrUpdateDoiModalTest.java
```diff
@@ -104,9 +101,20 @@
 	@Captor
 	ArgumentCaptor<DoiRequest> doiRequestCaptor;
 
+	Project mockProjectEntity;
+	FileEntity mockFileEntity;
```
nickgros: Use real objects instead of mocks since `getClass` is called on these, which has unexpected behavior with mocks (`getClass` yields a proxy instead of the mocked class).

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -610,7 +612,14 @@ public DatasetEditorProps getItemsEditorProps() {
...
 						entityBundle.getEntity().getId(),
-						() -> closeItemsEditor(),
+						() -> {
+							ToastMessageOptions toastOptions = new ToastMessageOptions.Builder()
+									.setTitle("Dataset Saved")
+									.setPrimaryButton("Show Schema", () -> this.actionMenu.onAction(Action.SHOW_TABLE_SCHEMA))
+									.build();
+							DisplayUtils.notify("Edit the Dataset Schema to add additional annotation columns to this dataset", DisplayUtils.NotificationVariant.SUCCESS, toastOptions);
+							closeItemsEditor();
+						},
```
nickgros, line 622: Toast to display when saving a Dataset

src/main/java/org/sagebionetworks/web/client/widget/doi/CreateOrUpdateDoiModal.java
```diff
@@ -55,6 +56,8 @@
 	private PopupUtilsView popupUtilsView;
 	private DateTimeUtils dateTimeUtils;
 
+	private boolean doiExists = false;
```
nickgros, line 59: Moved this to a class instance variable so we can access it outside of the configure method

src/main/java/org/sagebionetworks/web/client/widget/doi/CreateOrUpdateDoiModalView.java
```diff
@@ -32,7 +32,7 @@
 
 	void setPublicationYear(Long publicationYear);
 
-	void showOverwriteWarning(boolean showWarning);
+	void setOverwriteWarningVisible(boolean showWarning);
```
nickgros: No change, just clarity/consistency with other view visibility methods

src/test/java/org/sagebionetworks/web/unitclient/widget/doi/CreateOrUpdateDoiModalTest.java
```diff
@@ -86,27 +87,32 @@
...
-	@Mock
-	FileEntity mockFileEntity;
-	@Mock
 	UserProfile mockUserProfile;
 	@Mock
 	PopupUtilsView mockPopupUtilsView;
 	@Mock
 	DateTimeUtils mockDateTimeUtils;
-	@Mock
-	FluentFuture<Doi> mockDoiFuture;
```
nickgros, line 99: unused

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -256,6 +256,10 @@ public static void notify(String message, NotificationVariant variant) {
 		notify(message, variant, null);
 	}
 
+	public static void notify(String title, String message, NotificationVariant variant) {
+		notify(message, variant, new ToastMessageOptions.Builder().setTitle(title).build());
+	}
```
nickgros, line 261: It's common to have a toast with just a title and a message. Handy to have this so we don't have to pull out `ToastMessageOptions.Builder` for what should be a more simple configuration

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsWidget.java
```diff
@@ -229,6 +233,7 @@ public void onSuccess(TableUpdateTransactionRequest request) {
 	}
 
 	public void finished() {
+		popupUtilsView.notify("Schema Updated", "You made changes to the columns in this " + tableType.getDisplayName(), DisplayUtils.NotificationVariant.INFO);
```
nickgros, line 236: Toast to display when updating a table schema

src/main/java/org/sagebionetworks/web/client/PopupUtilsView.java
```diff
@@ -10,6 +10,8 @@
 public interface PopupUtilsView {
 	void notify(String message, DisplayUtils.NotificationVariant variant, ToastMessageOptions options);
```
jay-hodgson, line 11: Would love to get rid of some of these signatures, but if there are many references that have varying requirements then I agree that it should be another project.

src/main/java/org/sagebionetworks/web/client/widget/doi/CreateOrUpdateDoiModalView.java
jay-hodgson: (could rename showWarning to visible)

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
jay-hodgson: This is why I love user research, leading the user to a likely next logical step in the workflow.

src/test/java/org/sagebionetworks/web/unitclient/widget/doi/CreateOrUpdateDoiModalTest.java
jay-hodgson: Should probably rename.  Maybe testProjectEntity and testFileEntity (since they are not mocked)?

jay-hodgson (approved):

PR #4763, opened 2022-01-26
SWC-5935

nickgros:
Fixes [SWC-5935](https://sagebionetworks.jira.com/browse/SWC-5935)

Note that the target branch is release-391


src/main/java/org/sagebionetworks/web/client/widget/FullWidthAlert.java
```diff
@@ -31,8 +31,15 @@ public FullWidthAlert() {
 	
 	private void rerender() {
 		Double autoCloseAfterDelayInSeconds = null;
-		AlertButtonConfig primaryButtonConfig = AlertButtonConfig.create(primaryButtonText, onPrimaryClick);
```
nickgros, line 34: Leave the button configurations null if we don't specify text + an action (or disabled, for the secondary button)

jay-hodgson (approved):

PR #4771, opened 2022-01-27
SWC-5934: do not output string Version with label

jay-hodgson:


jay-hodgson:
Verified that this was the cause: Screen Shot 2022-01-27 at 3 01 25 PM

nickgros (approved):

PR #4774, opened 2022-01-28
SWC-5940: tell the browser to expire js assets after 8 hours

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/servlet/filter/GWTCacheControlFilter.java
```diff
@@ -49,26 +50,24 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 		if (lastDotIndex > -1) {
 			extension = requestURI.substring(lastDotIndex).toLowerCase();
 		}
-		if (requestURI.contains(".cache.")) {
```
jay-hodgson, line 52: Diff is a little misleading.  I moved the .nocache. block before the .cache. block (since .nocache files are also js)

xschildw (approved):

xschildw (approved):

PR #4777, opened 2022-01-31
SWC-5943: work towards standardizing the base font size as 16px inste…

jay-hodgson:
…ad of 14px


jay-hodgson:
Also needs to update to the latest SRC (once it has the corresponding base font-size changes)

nickgros (approved):

PR #4778, opened 2022-02-01
SWC-5921

nickgros:
[SWC-5921](https://sagebionetworks.jira.com/browse/SWC-5921), I think I also ended up addressing [SWC-5852](https://sagebionetworks.jira.com/browse/SWC-5852)

- After creating a Dataset/Table/View, automatically navigate to the new Entity
- As an editor, when opening the page of a Dataset with no items, automatically show Edit mode (including when the Dataset is first created)
- Show a banner if the View/Dataset has no scope (this will have priority over the existing banner for when there are no columns)
- Don't automatically query Views/Datasets with an empty/null scope (we already do this for columns)
- Disable Download Options, Edit Table Data if we don't query Views/Datasets (because of no scope/columns)

Clips + Screenshots:
Auto show navigate and show dataset editor after creating:

https://user-images.githubusercontent.com/17580037/152058730-98bdb067-f55c-4c56-95d5-c620ae3e35f1.mov

Banner for no items:

image

Same banner, logged out:
image

Disabled download options for a Dataset with no items: 

image

File View with no scope has disabled "Edit" and "Download" options:

https://user-images.githubusercontent.com/17580037/152059190-d35a39b3-35cd-4fc9-9f7a-4f88e401a7ec.mov

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/ToastMessageOptions.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/ToastMessageOptions.java
@@ -141,8 +141,14 @@ public class ToastMessageOptions extends ReactComponentProps {
 		}
 
 		public ToastMessageOptions build() {
-			AlertButtonConfig primaryButtonConfig = AlertButtonConfig.create(primaryButtonText, onPrimaryButtonClick);
-			AlertButtonConfig secondaryButtonConfig = AlertButtonConfig.create(secondaryButtonText, onSecondaryButtonClickOrHref);
+			AlertButtonConfig primaryButtonConfig = null;
+			AlertButtonConfig secondaryButtonConfig = null;
+			if (primaryButtonText != null && onPrimaryButtonClick != null) {
+				primaryButtonConfig = AlertButtonConfig.create(primaryButtonText, onPrimaryButtonClick);
+			}
+			if (secondaryButtonText != null && onSecondaryButtonClickOrHref != null) {
+				secondaryButtonConfig = AlertButtonConfig.create(secondaryButtonText, onSecondaryButtonClickOrHref);
+			}
             return ToastMessageOptions.create(title, autoCloseInMs, primaryButtonConfig, dismissOnPrimaryButtonClick, secondaryButtonConfig, dismissOnSecondaryButtonClick);
         }
 
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -759,19 +759,19 @@ public class EntityActionControllerImpl implements EntityActionController, Actio
 			boolean canEditResults = entityBundle.getPermissions().getCanCertifiedUserEdit() && !isDataset;
 			actionMenu.setActionVisible(Action.UPLOAD_TABLE_DATA, canEditResults);
 			actionMenu.setActionText(Action.UPLOAD_TABLE_DATA, "Upload Data to " + entityTypeDisplay);
-			actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, canEditResults);
 			actionMenu.setActionVisible(Action.DOWNLOAD_TABLE_QUERY_RESULTS, true);
 			actionMenu.setActionVisible(Action.SHOW_TABLE_SCHEMA, true);
 			actionMenu.setActionVisible(Action.SHOW_VIEW_SCOPE, !(entityBundle.getEntity() instanceof TableEntity) && !isDataset);
 			actionMenu.setActionVisible(Action.EDIT_DATASET_ITEMS, isDataset);
 		} else {
 			actionMenu.setActionVisible(Action.UPLOAD_TABLE_DATA, false);
-			actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, false);
 			actionMenu.setActionVisible(Action.DOWNLOAD_TABLE_QUERY_RESULTS, false);
 			actionMenu.setActionVisible(Action.SHOW_TABLE_SCHEMA, false);
 			actionMenu.setActionVisible(Action.SHOW_VIEW_SCOPE, false);
 			actionMenu.setActionVisible(Action.EDIT_DATASET_ITEMS, false);
 		}
+		// The EDIT_TABLE_DATA action will be shown by other widgets when data is editable.
+		actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, false);
 	}
 
 	private void configureFileUpload() {
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -759,19 +759,19 @@ public class EntityActionControllerImpl implements EntityActionController, Actio
 			boolean canEditResults = entityBundle.getPermissions().getCanCertifiedUserEdit() && !isDataset;
 			actionMenu.setActionVisible(Action.UPLOAD_TABLE_DATA, canEditResults);
 			actionMenu.setActionText(Action.UPLOAD_TABLE_DATA, "Upload Data to " + entityTypeDisplay);
+			actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, canEditResults);
 			actionMenu.setActionVisible(Action.DOWNLOAD_TABLE_QUERY_RESULTS, true);
 			actionMenu.setActionVisible(Action.SHOW_TABLE_SCHEMA, true);
 			actionMenu.setActionVisible(Action.SHOW_VIEW_SCOPE, !(entityBundle.getEntity() instanceof TableEntity) && !isDataset);
 			actionMenu.setActionVisible(Action.EDIT_DATASET_ITEMS, isDataset);
 		} else {
 			actionMenu.setActionVisible(Action.UPLOAD_TABLE_DATA, false);
+			actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, false);
 			actionMenu.setActionVisible(Action.DOWNLOAD_TABLE_QUERY_RESULTS, false);
 			actionMenu.setActionVisible(Action.SHOW_TABLE_SCHEMA, false);
 			actionMenu.setActionVisible(Action.SHOW_VIEW_SCOPE, false);
 			actionMenu.setActionVisible(Action.EDIT_DATASET_ITEMS, false);
 		}
-		// The EDIT_TABLE_DATA action will be shown by other widgets when data is editable.
-		actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, false);
 	}
 
 	private void configureFileUpload() {
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/menu/v2/ActionMenuWidgetViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/menu/v2/ActionMenuWidgetViewImpl.java
@@ -63,9 +63,11 @@ public class ActionMenuWidgetViewImpl implements ActionMenuWidgetView {
 	@UiField
 	ActionMenuItem programmaticOptionsMenuItem;
 	@UiField
-	Tooltip addToDownloadListMenuItemTooltip;
+	ActionMenuItem exportTableMenuItem;
 	@UiField
-	Tooltip programmaticOptionsMenuItemTooltip;
+	Tooltip downloadOptionsTooltip;
+	@UiField
+	Tooltip editTableDataTooltip;
 	@UiField
 	ReactComponentDiv skeletonButton;
 
@@ -174,23 +176,30 @@ public class ActionMenuWidgetViewImpl implements ActionMenuWidgetView {
 	}
 
 	@Override
+	public void setDownloadActionsDisabledTooltipText(String tooltipText) {
+		downloadOptionsTooltip.setTitle(tooltipText);
+		downloadOptionsTooltip.recreate();
+	}
+
+	@Override
+	public void setEditTableDataTooltipText(String tooltipText) {
+		editTableDataTooltip.setTitle(tooltipText);
+		editTableDataTooltip.recreate();
+	}
+
+	@Override
 	public void setDownloadActionsEnabled(boolean enabled) {
-		addToDownloadListMenuItem.setEnabled(enabled);
-		programmaticOptionsMenuItem.setEnabled(enabled);
+		tableDownloadOptions.setEnabled(enabled);
 		if (enabled) {
-			// The tooltips only show information about why it's disabled, so hide the tooltips if enabled
-			addToDownloadListMenuItemTooltip.setTrigger(Trigger.MANUAL);
-			programmaticOptionsMenuItemTooltip.setTrigger(Trigger.MANUAL);
-			addToDownloadListMenuItemTooltip.hide();
-			programmaticOptionsMenuItemTooltip.hide();
+			// The tooltip only shows information about why it's disabled, so hide the tooltips if enabled
+			downloadOptionsTooltip.setTrigger(Trigger.MANUAL);
+			downloadOptionsTooltip.hide();
 		} else {
-			addToDownloadListMenuItemTooltip.setTrigger(Trigger.HOVER);
-			programmaticOptionsMenuItemTooltip.setTrigger(Trigger.HOVER);
+			downloadOptionsTooltip.setTrigger(Trigger.HOVER);
 		}
 
-		// Commit the changes to the tooltips
-		addToDownloadListMenuItemTooltip.recreate();
-		programmaticOptionsMenuItemTooltip.recreate();
+		// Commit the changes to the tooltip
+		downloadOptionsTooltip.recreate();
 	}
 
 	@Override
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
@@ -348,13 +348,17 @@ public abstract class AbstractTablesTab implements TablesTabView.Presenter, Quer
 	private void configureVersionAlert() {
 		boolean versionHistoryIsVisible = this.metadata.getVersionHistoryWidget().isVisible();
 		this.view.setVersionAlertPrimaryAction((versionHistoryIsVisible ? "Hide" : "Show") + " Version History",
-				e -> {
+				event -> {
 					this.metadata.getVersionHistoryWidget().setVisible(!versionHistoryIsVisible);
 					// Reconfigure the version alert to update the button text.
 					configureVersionAlert();
 				});
-		if (entityBundle.getEntity() instanceof Dataset && ((Dataset) entityBundle.getEntity()).getIsLatestVersion()) {
+		if (!(entityBundle.getEntity() instanceof Dataset)) {
+			// Don't show a version alert if this isn't a dataset
+			this.view.setVersionAlertVisible(false);
+		} else if (entityBundle.getEntity() instanceof Dataset && ((Dataset) entityBundle.getEntity()).getIsLatestVersion()) {
 			// This is the 'draft' version of the dataset
+			// Show an alert pointing to the most recent snapshot (if it exists)
 			this.view.setVersionAlertVisible(true);
 			this.view.setVersionAlertCopy(VERSION_ALERT_DRAFT_DATASET_TITLE, VERSION_ALERT_DRAFT_DATASET_MESSAGE);
 			ClickHandler goToLatestSnapshot = e -> ginInjector.getGlobalApplicationState()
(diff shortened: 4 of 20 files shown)


src/main/java/org/sagebionetworks/web/client/jsinterop/ToastMessageOptions.java
```diff
@@ -141,8 +141,14 @@ public Builder setSecondaryButton(String text, AlertButtonConfig.Callback onClic
...
-			AlertButtonConfig primaryButtonConfig = AlertButtonConfig.create(primaryButtonText, onPrimaryButtonClick);
-			AlertButtonConfig secondaryButtonConfig = AlertButtonConfig.create(secondaryButtonText, onSecondaryButtonClickOrHref);
+			AlertButtonConfig primaryButtonConfig = null;
+			AlertButtonConfig secondaryButtonConfig = null;
+			if (primaryButtonText != null && onPrimaryButtonClick != null) {
+				primaryButtonConfig = AlertButtonConfig.create(primaryButtonText, onPrimaryButtonClick);
+			}
+			if (secondaryButtonText != null && onSecondaryButtonClickOrHref != null) {
+				secondaryButtonConfig = AlertButtonConfig.create(secondaryButtonText, onSecondaryButtonClickOrHref);
+			}
```
nickgros, line 151: We were still getting some empty buttons in some alerts, but hopefully this change fixes the last of the issues. I think all of the alerts affected by this code change are rare (in experimental mode), so IMO this doesn't need to go into staging.

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -759,19 +759,19 @@ private void configureTableCommands() {
...
 		} else {
 			actionMenu.setActionVisible(Action.UPLOAD_TABLE_DATA, false);
-			actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, false);
 			actionMenu.setActionVisible(Action.DOWNLOAD_TABLE_QUERY_RESULTS, false);
 			actionMenu.setActionVisible(Action.SHOW_TABLE_SCHEMA, false);
 			actionMenu.setActionVisible(Action.SHOW_VIEW_SCOPE, false);
 			actionMenu.setActionVisible(Action.EDIT_DATASET_ITEMS, false);
 		}
+		// The EDIT_TABLE_DATA action will be shown by other widgets when data is editable.
+		actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, false);
```
nickgros, line 774: Whether data is editable depends on the query state. If a View query has no results (or we don't even query it because it has no scope/columns), then we need to alter the state of this button, which we cannot do here.

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
```diff
@@ -284,15 +283,10 @@ public void setTargetBundle(EntityBundle bundle, Long versionNumber) {
 		modifiedCreatedBy.setVisible(false);
 		view.setTableUIVisible(isShownInTab);
 		view.setActionMenu(tab.getEntityActionMenu());
-		tab.getEntityActionMenu().setTableDownloadOptionsVisible(isShownInTab);
-		boolean isCurrentVersion = version == null;
-
-		if (isDataset && isCurrentVersion) {
-			// SWC-5878 - On the current (non-snapshot) version of a dataset, only editors should be able to download
-			tab.getEntityActionMenu().setTableDownloadOptionsEnabled(bundle.getPermissions().getCanCertifiedUserEdit());
```
nickgros, line 292: Moved this logic to TableEntityWidget. We also don't need to check if this is a dataset in this class anymore, which is nice since that's the point of the `abstract`!

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
```diff
@@ -284,15 +283,10 @@ public void setTargetBundle(EntityBundle bundle, Long versionNumber) {
...
-		} else {
-			tab.getEntityActionMenu().setTableDownloadOptionsEnabled(true);
+		if (!isShownInTab) {
+			tab.getEntityActionMenu().setTableDownloadOptionsVisible(false);
 		}
```
nickgros, line 289: Hide the download options if we're no longer looking at a specific Table. Previously, we were doing this:

```java
tab.getEntityActionMenu().setTableDownloadOptionsVisible(isShownInTab);
```

but we no longer want to unconditionally show the download options if we're looking at a Table. If we don't query the table (e.g. because there are no columns), then we shouldn't allow Downloads.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep1.java
```diff
@@ -169,6 +175,8 @@ private void createEntity(final Entity entity) {
 			public void onSuccess(Entity table) {
 				// For Datasets, this is the only step.
 				if (TableType.dataset.equals(tableType)) {
+					// Go to the dataset page
+					globalAppState.getPlaceChanger().goTo(new Synapse(table.getId()));
```
nickgros, line 179: Immediately navigate to the new Dataset page after creating the Dataset.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep2.java
```diff
@@ -213,6 +217,7 @@ public void onFailure(Throwable caught) {
 			@Override
 			public void onSuccess(TableUpdateTransactionRequest request) {
 				if (request.getChanges().isEmpty()) {
+					globalAppState.getPlaceChanger().goTo(new Synapse(entity.getId()));
```
nickgros, line 220: Immediately navigate to the new Table/View page after creating it (if we don't have to create an asynch job).

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep2.java
```diff
@@ -241,6 +246,7 @@ public void onFailure(Throwable failure) {
 			@Override
 			public void onComplete(AsynchronousResponseBody response) {
 				view.setJobTrackerVisible(false);
+				globalAppState.getPlaceChanger().goTo(new Synapse(entity.getId()));
```
nickgros, line 249: Immediately navigate to the new Table/View page after the tracked asynch job ends.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -172,21 +194,38 @@ public void configure(EntityBundle bundle, Long versionNumber, boolean canEdit,
 		this.actionMenu = actionMenu;
 		this.entityTypeDisplay = EntityTypeUtils.getFriendlyEntityTypeName(bundle.getEntity());
 		reconfigureState();
+		showEditorIfEditableAndEmpty();
```
nickgros, line 191: When the Table widget is first configured, show the editor if the View is "empty" (no scope/items) and the user has edit permissions. This will only work for datasets for now, since that's the only "inline" editor we have.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -172,21 +194,38 @@ public void configure(EntityBundle bundle, Long versionNumber, boolean canEdit,
 		this.actionMenu = actionMenu;
 		this.entityTypeDisplay = EntityTypeUtils.getFriendlyEntityTypeName(bundle.getEntity());
 		reconfigureState();
+		showEditorIfEditableAndEmpty();
 	}
 
 	private void reconfigureState() {
+		initializeQuery(() -> initSimpleAdvancedQueryState());
```
nickgros, line 201: Initialize the query first so we can reconfigure actions based on the query state.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -172,21 +194,38 @@ public void configure(EntityBundle bundle, Long versionNumber, boolean canEdit,
...
+		}
 	}
 
 	/**
 	 * Setup the actions for this widget.
 	 */
 	private void configureActions() {
 		// Listen to action events.
-		isShowingScope = false;
-		isShowingSchema = false;
```
nickgros, line 189: Replaced with getters on the View where we can just get the visibility state.

src/main/java/org/sagebionetworks/web/client/widget/entity/menu/v2/ActionMenuWidgetViewImpl.java
```diff
@@ -173,24 +175,31 @@ public void setTableDownloadOptionsVisible(boolean visible) {
...
+	public void setEditTableDataTooltipText(String tooltipText) {
+		editTableDataTooltip.setTitle(tooltipText);
+		editTableDataTooltip.recreate();
+	}
+
 	@Override
 	public void setDownloadActionsEnabled(boolean enabled) {
-		addToDownloadListMenuItem.setEnabled(enabled);
-		programmaticOptionsMenuItem.setEnabled(enabled);
+		tableDownloadOptions.setEnabled(enabled);
```
nickgros, line 192: Before, we were disabling the individual download menu items. Now, just disable the whole dropdown.

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
```diff
@@ -385,7 +380,7 @@ private void configureVersionAlert() {
 					goToLatestSnapshot,
 					isLinkToCurrentSnapshotEnabled,
 					linkToCurrentSnapshotTooltipText);
-		} else if (entityBundle.getEntity() instanceof Dataset && !version.equals(latestSnapshotVersionNumber)) {
+		} else if (entityBundle.getEntity() instanceof Dataset && !((Dataset) entityBundle.getEntity()).getIsLatestVersion() && version != latestSnapshotVersionNumber) {
```
nickgros: This additional condition was always true, but this makes the check more clear.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -94,16 +120,17 @@ public static final String getNoColumnsNotEditableMessage(TableType tableType) {
...
 	FileViewClientsHelp fileViewClientsHelp;
-	boolean isShowingSchema, isShowingScope, isShowingItemsEditor;
 	public static final String SHOW = "Show ";
 	public static final String HIDE = "Hide ";
 	public static final String SCOPE = "Scope of ";
 	public static final String SCHEMA = " Schema";
 	String entityTypeDisplay;
 	PortalGinInjector ginInjector;
 
+	boolean hasQueryableData; // if `false`, then a query will never yield data.
```
nickgros, line 118: e.g. if an EntityView/SubmissionView has no scope or Dataset has no items

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -172,21 +200,41 @@ public void configure(EntityBundle bundle, Long versionNumber, boolean canEdit,
 		this.actionMenu = actionMenu;
 		this.entityTypeDisplay = EntityTypeUtils.getFriendlyEntityTypeName(bundle.getEntity());
 		reconfigureState();
+		showEditorIfEditableAndEmpty();
 	}
 
 	private void reconfigureState() {
+		initializeQuery();
```
nickgros, line 207: `initializeQuery` will set `hasQueryableData`

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -172,21 +200,41 @@ public void configure(EntityBundle bundle, Long versionNumber, boolean canEdit,
...
 	}
 
 	private void reconfigureState() {
+		initializeQuery();
+		if (this.hasQueryableData) {
+			initSimpleAdvancedQueryState();
+		}
 		configureActions();
-		checkState();
-		initSimpleAdvancedQueryState();
```
nickgros, line 180: - Renamed `checkState` -> `initializeQuery`.
- `initializeQuery` will set `this.hasQueryableData`.
- We call `configureActions` after `initializeQuery` because it now depends on `this.hasQueryableData`

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -195,37 +243,63 @@ private void configureActions() {
...
+			showDatasetItemsEditor();
 		});
 
+		// Edit data
+		this.actionMenu.setActionEnabled(Action.EDIT_TABLE_DATA, hasQueryableData);
+		if (hasQueryableData) {
+			this.actionMenu.setEditTableDataTooltipText("Bulk edit cell values");
+		} else {
+			this.actionMenu.setEditTableDataTooltipText("There is no data to edit");
+		}
```
nickgros, line 273: I think this addresses [SWC-5852](https://sagebionetworks.jira.com/browse/SWC-5852)

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -195,37 +243,63 @@ private void configureActions() {
...
-				actionMenu.setActionVisible(Action.EDIT_DATASET_ITEMS, false);
-				view.setItemsEditorVisible(true);
-				isShowingItemsEditor = true;
-				view.setQueryResultsVisible(false);
-			};
+			showDatasetItemsEditor();
 		});
 
+		// Edit data
+		this.actionMenu.setActionEnabled(Action.EDIT_TABLE_DATA, hasQueryableData);
```
nickgros, line 256: Disable the button if the table can't have any editable data.

EntityActionController will continue to show/hide this button when appropriate.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -195,37 +243,63 @@ private void configureActions() {
...
+		if (hasQueryableData) {
+			this.actionMenu.setEditTableDataTooltipText("Bulk edit cell values");
```
nickgros, line 258: I think this addresses [SWC-5852](https://sagebionetworks.jira.com/browse/SWC-5852)

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -195,37 +243,63 @@ private void configureActions() {
...
+		} else {
+			this.actionMenu.setEditTableDataTooltipText("There is no data to edit");
+		}
+
+		// Download options
+		this.actionMenu.setTableDownloadOptionsVisible(true);
```
nickgros, line 264: If this Widget is being configured, the download options should always be visible.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -195,37 +243,63 @@ private void configureActions() {
...
+			onDownloadResults();
+		});
 		this.actionMenu.setActionVisible(Action.ADD_TABLE_RESULTS_TO_DOWNLOAD_LIST, tableType.isIncludeFiles());
 		this.actionMenu.setActionVisible(Action.TABLE_DOWNLOAD_PROGRAMMATIC_OPTIONS, tableType.isIncludeFiles());
+		if (this.entityBundle.getEntity() instanceof Dataset && isCurrentVersion && hasQueryableData) {
+			// SWC-5878 - On the current (non-snapshot) version of a dataset, only editors should be able to download
+			this.actionMenu.setTableDownloadOptionsEnabled(canEdit);
+			if (!canEdit) {
+				this.actionMenu.setDownloadActionsDisabledTooltipText("A draft version of a Dataset cannot be downloaded");
+			}
```
nickgros, line 275: This behavior was moved here from AbstractTablesTab

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -195,37 +243,63 @@ private void configureActions() {
...
+		} else {
+			this.actionMenu.setTableDownloadOptionsEnabled(hasQueryableData);
+			if (!hasQueryableData) {
+				this.actionMenu.setDownloadActionsDisabledTooltipText("There is no data to download");
+			}
```
nickgros, line 280: If a View has no data, it's not downloadable.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -235,24 +309,26 @@ private void configureActions() {
 		this.actionMenu.setActionListener(Action.SHOW_QUERY, action -> {
 			onShowQuery();
 		});
-		this.actionMenu.setActionListener(Action.TABLE_DOWNLOAD_PROGRAMMATIC_OPTIONS, action -> {
```
nickgros, line 238: These were just moved above

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -195,37 +243,63 @@ private void configureActions() {
 		this.actionMenu.setActionListener(Action.UPLOAD_TABLE_DATA, action -> {
 			onUploadTableData();
 		});
-		this.actionMenu.setActionListener(Action.DOWNLOAD_TABLE_QUERY_RESULTS, action -> {
```
nickgros, line 198: moved below

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/tabs/DatasetsTabTest.java
```diff
@@ -463,49 +463,6 @@ public void testShowProjectLoadError() {
 		verify(mockSynapseAlert).handleException(projectLoadError);
 	}
 
-	@Test
-	public void testEditorCanDownloadLatestVersion() {
```
nickgros, line 467: This behavior is now in the TableEntityWidget so tests were removed here and new tests were added there.

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
jay-hodgson: @nickgros , could you clarify the cases (in a code comment) to clarify this logic?
I think the cases are:  latest/draft version, most recent snapshot, and older snapshot.
To confirm, is getIsLatestVersion() referring to the "draft" version, and latestSnapshotVersionNumber is the most recent snapshot?

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -66,6 +74,23 @@ public static final String getNoColumnsNotEditableMessage(TableType tableType) {
 		return "This " + tableType.getDisplayName() + " does not have any columns.";
 	}
 
+	public static final String noScopeEditableMessage(TableType tableType) {
```
jay-hodgson: Append to noScopeNotEditableMessage()? Benefit is that it would remove duplicate copy.

src/main/resources/org/sagebionetworks/web/client/widget/entity/menu/v2/ActionMenuWidgetViewImpl.ui.xml
```diff
@@ -17,24 +17,22 @@
 		<b:Tooltip title="Show query">
 			<a:ActionButton visible="false" icon="DATABASE" addStyleNames="margin-right-5" action="SHOW_QUERY" />
 		</b:Tooltip>
-		<b:Tooltip title="Edit query results">
-			<a:ActionButton visible="false" icon="EDIT" addStyleNames="margin-right-5" action="EDIT_TABLE_DATA" />
+		<b:Tooltip ui:field="editTableDataTooltip" title="Edit query results">
+            <bh:Span> <!-- Have to wrap the button in a span we get a mouseevent to show the tooltip if the button is disabled -->
```
jay-hodgson: > Have to wrap the button in a span we get a mouseevent to show the tooltip if the button is disabled

Interesting!
Seems to be a minor spacing issue here which you could fix (not mandatory).

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
nickgros: I commented + refactored the logic to try to clarify here: https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4778/files/f6fc15f51af523c7bfc6a89d6020e37bb6f41887..0be3a2a7b0f04672354e792bd1d41913dfc1c7ac

The cases you listed are correct, and we also have to check if the Table is a Dataset

PR #4779, opened 2022-02-01
SWC-5946 and SWC-5945

jay-hodgson:


src/main/resources/org/sagebionetworks/web/client/widget/footer/FooterViewImpl.ui.xml
```diff
@@ -39,7 +39,7 @@
 							<bh:Div width="100%" addStyleNames="flexcontainer-row">
 								<g:Image ui:field="alphaModeStatusIcon" addStyleNames="flexcontainer-column margin-right-5 margin-top-3" width="18px" height="18px" />
 								<b:Anchor ui:field="debugLink"
-									addStyleNames="flexcontainer-column flexcontainer-justify-flex-end color-white">
+									addStyleNames="flexcontainer-column flexcontainer-justify-flex-end color-white font-size-14">
```
jay-hodgson, line 42: Fix for SWC-5945: revert size of Experimental Mode ON/OFF text (back to font-size 14px)

jay-hodgson:
Updates the facet values UI a bit.  Replaces the GWT HorizontalPanel (which outputs a html table element) with a css grid.  Put the facet value in the auto width cell, and hide any x overflow (with text ellipsis support).

jay-hodgson:
Screen Shot 2022-02-01 at 11 57 13 AM

Also updated t
Screen Shot 2022-02-01 at 11 57 38 AM
he styling of the facet values (the layout and the count):

src/main/webapp/sass/_core.scss
```diff
@@ -3637,4 +3637,27 @@ div.border {
...
+	margin-right: 5px;
+	> * {
+		align-self: center;
+	}
+	.checkbox {
+		grid-column: 1 / span 1;
+	}
+	.value {
+		grid-column: 2 / span 1;
+		max-width: 355px;
```
nickgros, line 3654: Does this need to be specified? The width is already constrained by the grid column, but maybe it's a grid/overflow quirk

nickgros (approved):
Changes look great!

src/main/webapp/sass/_core.scss
jay-hodgson: The grid column 2 is set to auto, so without this the grid will extend outside of the container boundary to render the facet value content.

PR #4789, opened 2022-02-08
SWC-5950

nickgros:
We'll now show the total number of results above the Query results for all Tables.

Like the facets and select columns, we'll cache the query count so the backend doesn't need to recompute it unless the Table or View changed.

image


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWidget.java
```diff
@@ -272,8 +273,10 @@ public void onCancel() {
 	 */
 	private void setQueryResults(final QueryResultBundle bundle) {
 		if (cachedFullQueryResultBundle != null) {
+			// Add the cached parts that we didn't request to the new result bundle.
+			// See CACHED_PARTS_MASK for which parts are cached
+			bundle.setQueryCount(cachedFullQueryResultBundle.getQueryCount());
 			bundle.setColumnModels(cachedFullQueryResultBundle.getColumnModels());
-			bundle.setFacets(cachedFullQueryResultBundle.getFacets());
```
nickgros, line 276: We didn't previously retrieve the Facets in the query bundle (it wasn't present in the above parts mask). The FacetsWidget handles getting this info in its own request.

jay-hodgson (approved):

PR #4790, opened 2022-02-09
SWC-5949, SWC-5951

nickgros:
image


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/PlotlyWidget.java
```diff
@@ -1,10 +1,10 @@
 package org.sagebionetworks.web.client.widget.entity.renderer;
 
+import static org.sagebionetworks.repo.model.table.QueryOptions.BUNDLE_MASK_QUERY_MAX_ROWS_PER_PAGE;
+import static org.sagebionetworks.repo.model.table.QueryOptions.BUNDLE_MASK_QUERY_RESULTS;
+import static org.sagebionetworks.repo.model.table.QueryOptions.BUNDLE_MASK_QUERY_SELECT_COLUMNS;
 import static org.sagebionetworks.web.client.place.Synapse.EntityArea.TABLES;
 import static org.sagebionetworks.web.client.widget.entity.tabs.AbstractTablesTab.TABLE_QUERY_PREFIX;
-import static org.sagebionetworks.web.client.widget.table.v2.results.TableQueryResultWidget.BUNDLE_MASK_QUERY_MAX_ROWS_PER_PAGE;
-import static org.sagebionetworks.web.client.widget.table.v2.results.TableQueryResultWidget.BUNDLE_MASK_QUERY_RESULTS;
-import static org.sagebionetworks.web.client.widget.table.v2.results.TableQueryResultWidget.BUNDLE_MASK_QUERY_SELECT_COLUMNS;
```
nickgros, line 7: Throughout this PR, I changed these references to use the bitmask value that we get directly from the backend models

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -65,8 +69,6 @@
 	public static final String RESET_SEARCH_QUERY_MESSAGE = "The search query will be reset. Are you sure that you would like to switch to simple search mode?";
 
 	public static final String RESET_SEARCH_QUERY = "Reset search query?";
-	public static final long DEFAULT_OFFSET = 0L;
-	public static final String SELECT_FROM = "SELECT * FROM ";
```
nickgros, line 69: Moved to QueryBundleUtils

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -79,7 +81,6 @@ public static final String noScopeMessage(TableType tableType, boolean editable)
 		}
 	}
 
-	public static final long DEFAULT_LIMIT = 25;
```
nickgros, line 82: Moved to QueryBundleUtils

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -193,6 +194,9 @@ public void configure(EntityBundle bundle, Long versionNumber, boolean canEdit,
 
 	private void reconfigureState() {
 		initializeQuery();
+		if (entityBundle.getEntity() instanceof View) {
+			queryResultsWidget.getTotalVisibleResultsWidget().configure((View) entityBundle.getEntity());
+		}
```
nickgros, line 199: The TotalVisibleResultsWidget lives in the TableQueryResultsWidget because we want to display the widget within the context of the results/facets. However, TableQueryResultsWidget knows nothing about the actual entity. We get the widget from the queryResultsWidget here because we have the entity info, rather than passing along a ton of data that the TableQueryResultsWidget probably shouldn't care about.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -512,18 +516,7 @@ private void setNoScopeState() {
 	 * @return
 	 */
 	public Query getDefaultQuery() {
```
nickgros, line 518: Moved the logic to a static utility, but didn't delete this because there is one place where another widget calls this method and relies on the state of this widget.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TotalVisibleResultsWidget.java
```diff
@@ -0,0 +1,91 @@
...
+				@Override
+				public void onFailure(Throwable failure) {
+					popupUtils.showErrorMessage(failure.getMessage());
+				}
+
+				@Override
+				public void onComplete(AsynchronousResponseBody response) {
+					totalNumberOfVisibleResults = ((QueryResultBundle) response).getQueryCount().intValue();
+					view.setNumberOfHiddenResults(totalNumberOfResults - totalNumberOfVisibleResults);
+					view.setNumberOfHiddenResultsVisible(!totalNumberOfResults.equals(totalNumberOfVisibleResults));
```
nickgros, line 80: Only show the warning if something is hidden

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TotalVisibleResultsWidget.java
```diff
@@ -0,0 +1,91 @@
...
+		this.popupUtils = popupUtils;
+		this.asyncJobTracker = asyncJobTracker;
+	}
+
+	@Override
+	public Widget asWidget() {
+		return view.asWidget();
+	}
+
+	public void configure(View viewEntity) {
```
nickgros, line 47: Widget should handle all views if we can ever get the info from the backend

jay-hodgson, line 47: If this component is reused (reconfigured using different views), you probably want to initialize
setNumberOfHiddenResultsVisible(false) here (since it may show this while the async call is being run to answer this question).

src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultViewImpl.java
```diff
@@ -103,7 +106,7 @@ public void setResultCount(Long resultCount) {
 		} else if (new Long(1).equals(resultCount)) {
 			resultCountHeader.setText("1 Result");
 		} else {
-			resultCountHeader.setText(resultCount + " Results");
+			resultCountHeader.setText(NumberFormat.getDecimalFormat().format(resultCount) + " Results");
```
jay-hodgson, line 109: This is a nice touch (locale specific and friendly rendering of the number)

src/main/resources/org/sagebionetworks/web/client/widget/table/v2/TotalVisibleResultsWidgetViewImpl.ui.xml
```diff
@@ -0,0 +1,30 @@
...
+			 xmlns:bt="urn:import:org.sagebionetworks.web.client.view.bootstrap.table"
+			 xmlns:a="urn:import:org.sagebionetworks.web.client.widget.table.v2"
+			 xmlns:bg="urn:import:org.gwtbootstrap3.client.ui.gwt"
+			 xmlns:bh="urn:import:org.gwtbootstrap3.client.ui.html"
+			 xmlns:w="urn:import:org.sagebionetworks.web.client.widget">
+<bh:Div visible="false">
+		<b:Heading size="H3" ui:field="totalItemCountContainer">
+			<bh:Span ui:field="totalItemCountItemName">Files</bh:Span>
+			<bh:Span>&nbsp;(</bh:Span>
+			<bh:Span ui:field="totalItemCount">10000</bh:Span>
```
jay-hodgson: Doesn't matter in this case (since this widget is not shown before it's initialized with data), but I'd prefer this to default to an unknown value (like blank, or '-') rather than a random number.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TotalVisibleResultsWidget.java
```diff
@@ -0,0 +1,91 @@
...
+		// Currently, the only case where we can reliably get this info is for the current version of a Dataset.
+		// Other cases will need a new service, tracked by PLFM-7046
+		if (viewEntity instanceof Dataset && viewEntity.getIsLatestVersion()) {
+			this.totalNumberOfResults = ((Dataset) viewEntity).getItems().size();
+			view.setTotalNumberOfResults(totalNumberOfResults);
+			view.setHelpMarkdown(DATASETS_HELP);
```
jay-hodgson, line 53: Is the idea that this help text will change in the future, depending on the View type?  It looks like this is unnecessary today (since it's currently initialized to this help markdown text)

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/TotalVisibleResultsWidgetTest.java
```diff
@@ -0,0 +1,149 @@
...
+		datasetLatestVersion.setId(TABLE_ID);
+		datasetLatestVersion.setItems(datasetItems);
+		datasetLatestVersion.setIsLatestVersion(true);
+
+		datasetSnapshot = new Dataset();
+		datasetSnapshot.setId(TABLE_ID);
+		datasetLatestVersion.setItems(datasetItems);
+		datasetSnapshot.setIsLatestVersion(false);
+
+		entityView = new EntityView();
```
jay-hodgson, line 76: Reasoning for concrete objects rather than mocking these (DatasetItems, Datasets, EntityView, QueryResultBundle)?

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/TotalVisibleResultsWidgetTest.java
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	@Test
+	public void testEntityView() {
+		// We have no way to determine the maximum number of items that should be available.
+		widget.configure(entityView);
+
+		verify(mockView).setVisible(false);
+		verifyZeroInteractions(mockAsyncJobTracker);
+	}
```
jay-hodgson, line 148: Test logic looks great.

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TotalVisibleResultsWidget.java
nickgros: Correct

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/TotalVisibleResultsWidgetTest.java
nickgros: IMO it's easier to follow and debug real POJOs with data than mocks. This is also what the backend team has been doing. But if you prefer mocks, I can change these, not a big deal to me

PR #4791, opened 2022-02-10
Add .gitattributes and renormalize line endings

nickgros:
I used the generator based on https://github.com/alexkaratarakis/gitattributes to generate a [.gitattributes file](https://git-scm.com/docs/gitattributes).

The .gitattributes file will specify the line endings used when committing and checking out the file regardless of the local user's git settings. In general. The line endings specified may differ based on the file extension. For example, shell scripts will always be LF and batch scripts will always be CRLF.

Then I ran `git add --renormalize .` to fix all existing line endings to adhere to the .gitattributes conventions.

If you look at the [diff without whitespace changes](https://github.com/Sage-Bionetworks/SynapseWebClient/pull/4791/files?w=1), you should see only the new .gitattributes


.gitattributes
```diff
@@ -0,0 +1,325 @@
+## Generated with https://www.richie-bendall.ml/gitattributes-generator/, which uses https://github.com/alexkaratarakis/gitattributes
```
nickgros, line 1: For the generator options, I picked "Common" + "Web" + "Java"

PR #4803, opened 2022-02-17
SWC-5965

nickgros:
Before this fix, if you were previously in the dataset editor, creating a dataset will not show the editor, certain GWT widgets, or any other React components on the EntityView widget. Reproduction video below:

https://user-images.githubusercontent.com/17580037/154567087-73765c73-1da3-41fb-946e-801d9519d637.mov

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -1257,7 +1257,16 @@ public class EntityActionControllerImpl implements EntityActionController, Actio
 
 	private void postCheckCreateTableOrView(TableType type) {
 		getCreateTableViewWizard().configure(entityBundle.getEntity().getId(), type);
-		getCreateTableViewWizard().showModal(entityUpdatedWizardCallback);
+		getCreateTableViewWizard().showModal(new WizardCallback() {
+			@Override
+			public void onFinished() {
+				// Intentionally empty
+				// The modal will redirect us to the appropriate Place, which will trigger a full page update
+			}
+
+			@Override
+			public void onCanceled() {}
+		});
 	}
 
 	public void onAddSubmissionView() {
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1257,7 +1257,16 @@ public void onAddTable() {
 
 	private void postCheckCreateTableOrView(TableType type) {
 		getCreateTableViewWizard().configure(entityBundle.getEntity().getId(), type);
-		getCreateTableViewWizard().showModal(entityUpdatedWizardCallback);
+		getCreateTableViewWizard().showModal(new WizardCallback() {
+			@Override
+			public void onFinished() {
+				// Intentionally empty
+				// The modal will redirect us to the appropriate Place, which will trigger a full page update
```
nickgros, line 1264: The previous value of `onFinished` invoked `fireEntityUpdatedEvent`. The wizard calls `PlaceChanger.goTo` right before invoking `onFinished`, so we triggered two updates. I still don't understand why calling `fireEntityUpdatedEvent` immediately after changing places to the new entity would cause the Entity page configuration to completely break.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TotalVisibleResultsWidget.java
```diff
@@ -48,7 +51,8 @@ public void configure(View viewEntity) {
 		// Currently, the only case where we can reliably get this info is for the current version of a Dataset.
 		// Other cases will need a new service, tracked by PLFM-7046
 		if (viewEntity instanceof Dataset && viewEntity.getIsLatestVersion()) {
-			this.totalNumberOfResults = ((Dataset) viewEntity).getItems().size();
+			List<DatasetItem> datasetItems = ((Dataset) viewEntity).getItems();
+			this.totalNumberOfResults = datasetItems != null ? datasetItems.size() : 0;
```
nickgros, line 55: Unrelated fix, just need a null check here (null value is possible but unlikely).

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
jay-hodgson: Does not make sense to me either, but since the Placechanger.goTo will trigger a page update we don't need the EntityUpdatedEvent on the gwt event bus, so I think this is a safe change to make.

jay-hodgson (approved):

PR #4804, opened 2022-02-17
SWC-5805 - Add autocomplete values for password forms

nickgros:


src/main/java/org/sagebionetworks/web/client/view/users/PasswordResetViewImpl.java
```diff
@@ -182,6 +187,11 @@ public void onClick(ClickEvent event) {
 		password1Field.getElement().setAttribute("placeholder", "Enter password");
 		password2Field.getElement().setAttribute("placeholder", "Confirm password");
 
+		emailAddressField.getElement().setAttribute(AUTOCOMPLETE_ATTRIBUTE, AUTOCOMPLETE_VALUE_USERNAME);
+		currentPasswordField.getElement().setAttribute(AUTOCOMPLETE_ATTRIBUTE, AUTOCOMPLETE_VALUE_CURRENT_PASSWORD);
```
nickgros, line 191: I couldn't figure out how to get the email address or current password fields to show in the PasswordReset place. Not sure if it's still possible to do so

src/main/java/org/sagebionetworks/web/client/DisplayConstants.java
```diff
@@ -259,5 +259,9 @@
 	public static final String GO_TO_CURRENT_VERSION = "Go to the current version";
 	public static final String GO_TO_DRAFT_VERSION = "Go to the draft version";
 
+	public static final String AUTOCOMPLETE_ATTRIBUTE = "autocomplete";
+	public static final String AUTOCOMPLETE_VALUE_USERNAME = "username";
+	public static final String AUTOCOMPLETE_VALUE_CURRENT_PASSWORD = "current-password";
+	public static final String AUTOCOMPLETE_VALUE_NEW_PASSWORD = "new-password";
```
jay-hodgson, line 266: Looks good.  Maybe add a link to the possible values for context? https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/DisplayConstants.java
```diff
@@ -259,5 +259,9 @@
 	public static final String GO_TO_CURRENT_VERSION = "Go to the current version";
 	public static final String GO_TO_DRAFT_VERSION = "Go to the draft version";
 
+	public static final String AUTOCOMPLETE_ATTRIBUTE = "autocomplete";
+	public static final String AUTOCOMPLETE_VALUE_USERNAME = "username";
```
nickgros, line 264:
```suggestion
	// See possible values: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete
	public static final String AUTOCOMPLETE_VALUE_USERNAME = "username";
```

PR #4815, opened 2022-02-28
SWC-5837

nickgros:
* Cleaner "Empty" state for tables and datasets tabs
* View now recieves a state enum so we exclusively show loading vs empty vs populated views. Before we were showing the empty text, the table, and the loading spinner all at the same time. Now, you see the loading spinner only until we finish fetching items, and then only show the empty state or the table depending on the results.

image


src/main/java/org/sagebionetworks/web/client/widget/table/TableListWidget.java
```diff
@@ -72,6 +74,11 @@ public void configure(EntityBundle parentBundle, List<EntityType> typesToShow) {
 		isInitializing = true;
 		this.parentBundle = parentBundle;
 		this.typesToShow = typesToShow;
+		if (typesToShow.size() == 1 && typesToShow.contains(EntityType.dataset)) {
+			this.view.setTableType(TableType.dataset);
+		} else {
+			this.view.setTableType(TableType.table);
+		}
```
nickgros, line 80: The only two cases are the datasets tab and tables tab, so this is super specific and not very generalizable. I didn't want to spend more time handling copy for cases we don't have, and might not ever have.

jay-hodgson (approved):

PR #4817, opened 2022-03-01
SWC-5986

yjyoo773:


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/FilesBrowserViewImpl.java
```diff
@@ -60,13 +63,16 @@ public FilesBrowserViewImpl(FilesBrowserViewImplUiBinder binder, EntityTreeBrows
 			if (isEmpty) {
 				downloadHelp.setHelpMarkdown("There are no downloadable items in this folder.");
 				downloadHelp.setVisible(true);
```
nickgros: Is the the tooltip handling all of the cases that that the `downloadHelp` widget handled? If so, let's remove it.

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileTitleBarViewImpl.java
```diff
@@ -1,10 +1,6 @@
 package org.sagebionetworks.web.client.widget.entity.file;
 
-import org.gwtbootstrap3.client.ui.Anchor;
-import org.gwtbootstrap3.client.ui.AnchorListItem;
-import org.gwtbootstrap3.client.ui.Button;
-import org.gwtbootstrap3.client.ui.DropDownMenu;
-import org.gwtbootstrap3.client.ui.Heading;
+import org.gwtbootstrap3.client.ui.*;
```
nickgros, line 3: Could you configure your IDE in the following way? https://stackoverflow.com/questions/14716283/is-it-possible-for-intellij-to-organize-imports-the-same-way-as-in-eclipse

Sorry that this was missing from the bootstrap doc/not enforced by a linter built-in to the project!

No need to change it for this PR but just so it doesn't keep changing imports going forward.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -277,6 +277,8 @@ private void configureActions() {
 			if (!canEdit) {
 				this.actionMenu.setDownloadActionsDisabledTooltipText("A draft version of a Dataset cannot be downloaded");
 			}
+		} else if(!entityBundle.getPermissions().getCanDownload()) {
+			this.actionMenu.setTableDownloadOptionsEnabled(false);
```
nickgros: Do we need to set tooltip text for this case?

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
```diff
@@ -408,7 +410,7 @@ private void showSimpleSearchUI() {
 		actionMenu.setActionVisible(Action.SHOW_SIMPLE_SEARCH, false);
 		queryInputWidget.setShowSimpleSearchButtonVisible(false);
 		queryResultsWidget.setFacetsVisible(true);
-		actionMenu.setActionVisible(Action.SHOW_QUERY, true);
+		actionMenu.setActionVisible(Action.SHOW_QUERY, false);
```
nickgros: Is there a case where we're ever showing this button, or is it always supposed to be hidden now? If we never show it, let's remove the Action.SHOW_QUERY

src/main/resources/org/sagebionetworks/web/client/widget/entity/menu/v2/ActionMenuWidgetViewImpl.ui.xml
```diff
@@ -8,7 +8,7 @@
 	xmlns:w="urn:import:org.sagebionetworks.web.client.widget">
 
 	<g:FlowPanel>
-		<b:Tooltip title="Show advanced search">
+		<b:Tooltip title="Edit view query">
```
nickgros:
```suggestion
		<b:Tooltip title="Edit query">
```

This wasn't obvious from the design, but this tooltip needs to be dynamic or generic enough for all tables (not just Views, which are a type of table). Since this piece is built into the Explore component that's being ported, I'd say it's not worth the effort to make it dynamic

nickgros (changes requested):
Some questions and changes requested but looking good!

src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidget.java
yjyoo773: The tooltip text here is set in the setDownloadAcrionsEnabled function so this should be ok

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/FilesBrowserViewImpl.java
```diff
@@ -58,15 +61,13 @@ public FilesBrowserViewImpl(FilesBrowserViewImplUiBinder binder, EntityTreeBrows
 		});
 		entityTreeBrowser.setIsEmptyCallback(isEmpty -> {
 			if (isEmpty) {
-				downloadHelp.setHelpMarkdown("There are no downloadable items in this folder.");
-				downloadHelp.setVisible(true);
```
nickgros, line 62: Let's totally remove the downloadHelp widget from the Java (line 41) and the corresponding XML

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileTitleBarViewImpl.java
```diff
@@ -224,7 +218,12 @@ public void setVersion(Long versionNumber) {
 	@Override
 	public void setCanDownload(boolean canDownload) {
 		downloadOptionsButton.setEnabled(canDownload);
-		viewOnlyHelp.setVisible(!canDownload);
+		if(!canDownload){
+			String viewOnlyHelpText = authController.isLoggedIn() ? "You do not have download access for this item." : "You need to log in to download this file.";
+			downloadTooltip.setTitle(viewOnlyHelpText);
+		} else{
+			downloadTooltip.setTitle("Direct and programmatic download options");
+		}
```
nickgros: Just recommend flipping the cases for readability
```suggestion
		if (canDownload) {
			downloadTooltip.setTitle("Direct and programmatic download options");
		} else{
			String viewOnlyHelpText = authController.isLoggedIn() ? "You do not have download access for this item." : "You need to log in to download this file.";
			downloadTooltip.setTitle(viewOnlyHelpText);
		}
```

nickgros (approved):
Looks great! Noted one optional change

PR #4819, opened 2022-03-03
SWC-5991: basic support for materialized views

jay-hodgson:


jay-hodgson:
Need to update to the latest SRC for better support (entity icon support)

src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
```diff
@@ -123,6 +130,9 @@ public static IconType getIconTypeForEntityClassName(String className) {
 		} else if (EntityView.class.getName().equals(className)) {
 			// FileView
 			icon = IconType.TH_LIST;
+		} else if (MaterializedView.class.getName().equals(className)) {
+			// FileView
+			icon = IconType.TH_LIST;
```
jay-hodgson, line 135: Is this dead code?

jay-hodgson:
Added an editor, that is used during the initial creation of a MaterializedView from the web UI:
Screen Shot 2022-03-01 at 12 54 57 PM

And a special Edit Defining SQL action after creation:
Screen Shot 2022-03-03 at 12 39 01 PM

Removed version-related actions and UI since creating a snapshot of a MaterializedView results in a backend error.

src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
nickgros: Looks like it's still used by the Provenance graph but that's it

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -979,7 +998,8 @@ private void configureShareAction() {
 	}
 
 	public static boolean isVersionSupported(Entity entity, CookieProvider cookies) {
-		return entity instanceof Versionable;
+		boolean isMaterializedView = entity instanceof MaterializedView;
+		return entity instanceof Versionable && !isMaterializedView;
```
nickgros, line 1002: MaterializedViews implement Versionable but we don't support them in the UI?

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1254,6 +1280,13 @@ public void onAddTable() {
 			postCheckCreateTableOrView(TableType.table);
 		});
 	}
+	
+	public void onAddMaterializedView() {
+		preflightController.checkCreateEntity(entityBundle, MaterializedView.class.getName(), () -> {
+			// to create a MaterializedView, we need to know the definingSQL
+			getMaterializedViewEditor().configure(entityBundle.getEntity().getId()).show();
+		});
```
nickgros, line 1288: A user could create a materialized view with null definingSql if they close the editor, so the UI needs to be able to handle that case.

nickgros (approved):
Looks great!

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
jay-hodgson: Fails on attempted creation of the entity.
Screen Shot 2022-03-03 at 2 15 14 PM

PR #4821, opened 2022-03-09
SWC-6024

yjyoo773:


src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadMenuItem.java
```diff
@@ -112,9 +112,9 @@ public void configure(EntityBundle bundle, RestrictionInformationResponse restri
 					if (directDownloadUrl.startsWith(sftpProxy)) {
 						view.setIsAuthorizedDirectDownloadLink();
 						loginModalWidget.configure(directDownloadUrl, FormPanel.METHOD_POST, FormPanel.ENCODING_MULTIPART);
```
jay-hodgson: You could delete LoginModalWidget (it was only used for this sftp authenticated download).  Recommend you change this to simply popupUtilsView.showInfoDialog(...)

jay-hodgson (changes requested):
Let's get rid of LoginModalWidget

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadMenuItemViewImpl.java
```diff
@@ -56,9 +53,8 @@ private void clearClickHandlers() {
 	}
 
 	@Override
-	public void setIsAuthorizedDirectDownloadLink() {
-		clearClickHandlers();
-		downloadLinkHandlerRegistration = downloadLink.addClickHandler(authorizedDirectDownloadClickHandler);
+	public void showDownloadError(){
+		downloadLinkHandlerRegistration = downloadLink.addClickHandler(errorDownloadClickHandler);
```
jay-hodgson: I think you still need to call clearClickHandlers() before addClickHandler, since you can add multiple click handlers to the link (which is why we hold onto the downloadLinkHandlerRegistration).

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadMenuItemView.java
```diff
@@ -10,9 +10,9 @@
 
 	void setIsDirectDownloadLink(String href);
 
-	void setIsUnauthenticatedS3DirectDownload();
+	void showDownloadError();
```
jay-hodgson: Rename (I'd expect this to show a download error by calling).
Perhaps to something like
setIsSFTPDownload() (to be aligned with the other method names in this view)

jay-hodgson (approved):

PR #4824, opened 2022-03-14
SWC-6043

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWikiWidget.java
```diff
@@ -124,10 +124,11 @@ public void onSuccess(EntityBundle bundle) {
 				boolean isCurrentVersion = true;
 				entityActionController.configure(actionMenu, bundle, isCurrentVersion, bundle.getRootWikiId(), EntityArea.TABLES);
 				boolean canEdit = false;
-				hideEditActions();
 				if (!DisplayUtils.isInTestWebsite(ginInjector.getCookieProvider())) {
```
jay-hodgson, line 127: Going to delete this block when we bring Explore 2.0 out of alpha mode.  But the order apparently matters, since hideEditActions asks the TableEntityWidget to hide filtering, but we must set the actionMenu before calling this.

xschildw (approved):

PR #4829, opened 2022-03-18
SWC-6004 - Datasets out of Experimental Mode

nickgros:


src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep1ViewImpl.java
```diff
@@ -42,7 +42,6 @@
 	public CreateTableViewWizardStep1ViewImpl(Binder binder, FileViewOptions viewOptions, CookieProvider cookies) {
 		widget = binder.createAndBindUi(this);
 		this.viewOptions = viewOptions;
-		viewOptions.configure();
 		viewOptionsContainer.add(viewOptions);
```
jay-hodgson, line 45: viewOptions.configure() unnecessary?

nickgros: Yes, the only thing it did was check experimental mode.

PR #4832, opened 2022-03-22
SWC-6055

nickgros:
Don't fire EntityUpdatedEvent and use placeChanger.goTo at the same time, it causes the page to break.

Merging this into release-399 because we may decide to release Datasets from experimental mode in release-399


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1642,8 +1642,6 @@ public void onComplete(TableUpdateTransactionResponse response) {
 						if (!errors.isEmpty()) {
 							view.showErrorMessage(errors);
 						} else {
-							fireEntityUpdatedEvent();
```
nickgros, line 1645: We're using the PlaceChanger to go to the new version at line 1662, so this is unnecessary (and actually causes the page to break)

jay-hodgson (approved):

PR #4839, opened 2022-03-28
PLFM-6067: Team names are shared with usernames

marcomarasca:


nickgros (approved):
Thanks, @marcomarasca !

PR #4841, opened 2022-03-31
PLFM-7233: Update spring version (Fix for CVE-2022-22965)

marcomarasca:


marcomarasca:
Note that this is a patch for prod

nickgros (approved):

PR #4847, opened 2022-04-11
Bump Synapse version to 402.0, drop unused dependencies

nickgros:


pom.xml
```diff
@@ -426,27 +426,6 @@
...
-		<dependency>
-			<groupId>org.sagebionetworks</groupId>
-			<artifactId>lib-logging</artifactId>
-			<version>${synapse.version}</version>
-		</dependency>
-		<dependency>
-			<groupId>org.sagebionetworks</groupId>
-			<artifactId>lib-communicationUtilities</artifactId>
-			<version>${synapse.version}</version>
-		</dependency>
```
nickgros, line 449: I don't think we're using any of these

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/QueryInputWidgetTest.java
```diff
@@ -1,22 +1,19 @@
...
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
+
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
-import org.sagebionetworks.table.query.ParseException;
```
nickgros, line 13: Unnecessary reference to otherwise unused lib-table-query dependency

jay-hodgson (approved):

PR #4848, opened 2022-04-11
Bump SRC to 3.0.2, adjust scss

nickgros:
Hotfix for https://sagebionetworks.jira.com/browse/SWC-6086

### DIFF
--- a/src/main/webapp/sass/_core.scss
+++ b/src/main/webapp/sass/_core.scss
@@ -3595,7 +3595,7 @@ div.border {
 }
 
 #body .EntityFinder .EntityFinderDetailsView .checkbox {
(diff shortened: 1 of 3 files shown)


src/main/webapp/sass/_core.scss
```diff
@@ -3595,7 +3595,7 @@ div.border {
 }
 
 #body .EntityFinder .EntityFinderDetailsView .checkbox {
-	margin-top: 0px;
+	margin-top: -12px;
```
nickgros, line 3598: These CSS changes are related to a change in synapse-react-client 3.0.1. Those changes were going to be in 402, but need to come out now

PR #4852, opened 2022-04-13
SWC-6070 - Clear react-query entity cache on updated event

nickgros:
When we fire an EntityUpdatedEvent, we should clear the entity-related data cached by react-query.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/events/EntityUpdatedEvent.java
+++ b/src/main/java/org/sagebionetworks/web/client/events/EntityUpdatedEvent.java
@@ -2,5 +2,26 @@ package org.sagebionetworks.web.client.events;
 
 import com.google.web.bindery.event.shared.binder.GenericEvent;
 
+/**
+ * When fired, refreshes the EntityPresenter and clears the react-query cache for the specified entity, or all entities
+ * if not specified.
+ */
 public class EntityUpdatedEvent extends GenericEvent {
+	private String entityId = null;
+
+	public EntityUpdatedEvent() {
+	}
+
+
+	public EntityUpdatedEvent(String entityId) {
+		this.entityId = entityId;
+	}
+
+	/**
+	 * Gets the ID of the updated entity. May be null.
+	 * @return
+	 */
+	public String getEntityId() {
+		return entityId;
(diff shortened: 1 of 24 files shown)


src/main/java/org/sagebionetworks/web/client/events/EntityUpdatedEvent.java
```diff
@@ -2,5 +2,26 @@
...
 public class EntityUpdatedEvent extends GenericEvent {
+	private String entityId = null;
+
+	public EntityUpdatedEvent() {
+	}
+
+
+	public EntityUpdatedEvent(String entityId) {
+		this.entityId = entityId;
+	}
```
nickgros: Optionally supply an entity ID when creating this event

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -268,6 +273,9 @@ public Widget asWidget() {
 
 	@EventHandler
 	public void onEntityUpdatedEvent(EntityUpdatedEvent event) {
+		SynapseReactClientQueryKey queryKey = SynapseReactClientQueryKey.create("entity", event.getEntityId());
+		// The query key is an array where the first element is the object we constructed
+		queryClient.invalidateQueries(new Object[]{queryKey});
```
nickgros, line 278: Invalidate the react-query client for the updated entity, or all entities if the event's entity ID is null

src/main/java/org/sagebionetworks/web/client/events/EntityUpdatedEvent.java
```diff
@@ -2,5 +2,28 @@
...
+
+	public EntityUpdatedEvent() {
+		super();
+	}
+
+
+	public EntityUpdatedEvent(String entityId) {
+		super();
+		this.entityId = entityId;
+	}
```
nickgros, line 20: Optionally supply an entity ID when creating this event.

jay-hodgson (approved):

jay-hodgson:
Merged to include in Synapse stack 403

PR #4859, opened 2022-04-26
SWC-6070 - Reset queries to undefined rather than invalidating

nickgros:
[invalidateQueries](https://react-query.tanstack.com/reference/QueryClient#queryclientinvalidatequeries), isn't quite what we want--it tells all matching queries to re-fetch, but won't clear the data.

Instead we should use [resetQueries](https://react-query.tanstack.com/reference/QueryClient#queryclientresetqueries), which resets the data to the initial value. This way we avoid showing stale data if the refetch takes a while (e.g. because a table needs to be re-built).

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
@@ -2,6 +2,8 @@ package org.sagebionetworks.web.client;
 
 import org.sagebionetworks.repo.model.file.FileHandleAssociateType;
 import org.sagebionetworks.web.client.callback.MD5Callback;
+import org.sagebionetworks.web.client.jsinterop.reactquery.QueryClient;
+import org.sagebionetworks.web.client.jsinterop.reactquery.SynapseReactClientQueryKey;
 import org.sagebionetworks.web.client.widget.provenance.nchart.LayoutResult;
 import org.sagebionetworks.web.client.widget.provenance.nchart.NChartCharacters;
 import org.sagebionetworks.web.client.widget.provenance.nchart.NChartLayersArray;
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
```diff
@@ -5,6 +5,7 @@
 import org.sagebionetworks.web.client.widget.provenance.nchart.LayoutResult;
 import org.sagebionetworks.web.client.widget.provenance.nchart.NChartCharacters;
 import org.sagebionetworks.web.client.widget.provenance.nchart.NChartLayersArray;
+
```
nickgros:
```suggestion
```

src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -778,6 +778,7 @@ public void setIsInnerProgrammaticHistoryChange() {
 		_setIsInnerProgrammaticHistoryChange();
 	}
 
+
```
nickgros:
```suggestion
```

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -273,9 +273,8 @@ public Widget asWidget() {
 
 	@EventHandler
 	public void onEntityUpdatedEvent(EntityUpdatedEvent event) {
-		SynapseReactClientQueryKey queryKey = SynapseReactClientQueryKey.create("entity", event.getEntityId());
-		// The query key is an array where the first element is the object we constructed
-		queryClient.invalidateQueries(new Object[]{queryKey});
+		List<SynapseReactClientQueryKey> queryKey = SynapseReactClientQueryKey.create("entity", event.getEntityId());
+		queryClient.resetQueries(queryKey);
```
nickgros, line 277: JsInterop works intuitively with ArrayList, doesn't work intuitively with arrays.

jay-hodgson (approved):
Interesting distinction

PR #4863, opened 2022-05-02
SWC-6070 - Pass version to EntityModal

nickgros:
SWC-side fix for [SWC-6070](https://sagebionetworks.jira.com/browse/SWC-6070)


src/main/java/org/sagebionetworks/web/client/jsinterop/EntityModalProps.java
```diff
@@ -22,11 +22,16 @@ public interface Callback {
 	String initialTab; // "METADATA" | "ANNOTATIONS"
 	@JsNullable
 	boolean showTabs;
+	@JsNullable
+	Double versionNumber;
```
nickgros, line 26: has to be Double for JS compatibility

jay-hodgson (approved):

PR #4868, opened 2022-05-05
SWC-6032 Pull Explore 2.0 out of alpha mode

nickgros:
* Delete TableEntityWidget, and remove alpha mode checks to use TableEntityWidgetV2
* Remove Tables download dropdown menu because QueryWrapperPlotNav includes a download menu 
* Add `showLastUpdatedOn` prop to always show the last updated datetime


src/main/java/org/sagebionetworks/web/client/widget/table/v2/results/TableQueryResultWikiWidget.java
```diff
@@ -128,13 +118,8 @@ public void onSuccess(EntityBundle bundle) {
...
 				boolean canEdit = false;
-				if (!DisplayUtils.isInTestWebsite(ginInjector.getCookieProvider())) {
-					getTableEntityWidget().configure(bundle, tableVersionNumber, canEdit, TableQueryResultWikiWidget.this, actionMenu);
-					hideEditActions();
-				} else {
-					hideEditActions();
-					getTableEntityWidgetV2().configure(bundle, tableVersionNumber, canEdit, isShowTableOnly, TableQueryResultWikiWidget.this, actionMenu);
-				}
+				hideEditActions();
+				getTableEntityWidgetV2().configure(bundle, tableVersionNumber, canEdit, isShowTableOnly, TableQueryResultWikiWidget.this, actionMenu);
```
nickgros, line 122: @jay-hodgson Looks like we're still properly handling the special case for wiki

jay-hodgson: Could you also take a look at TableQueryResultWikiViewImpl.  The isShowTableOnly option should be available on release of Explore 2.0

src/main/java/org/sagebionetworks/web/client/widget/entity/editor/TableQueryResultWikiViewImpl.java
```diff
@@ -23,10 +23,9 @@ public interface TableQueryResultViewUiBinder extends UiBinder<Widget, TableQuer
 	CheckBox isShowTableOnly;
 
 	@Inject
-	public TableQueryResultWikiViewImpl(TableQueryResultViewUiBinder binder, CookieProvider cookies) {
+	public TableQueryResultWikiViewImpl(TableQueryResultViewUiBinder binder) {
 		widget = binder.createAndBindUi(this);
-		boolean isInTestWebsiteMode = DisplayUtils.isInTestWebsite(cookies);
-		isShowTableOnly.setVisible(isInTestWebsiteMode);
+		isShowTableOnly.setVisible(true);
```
jay-hodgson: This works - alternatively it could be set to visible in the ui.xml instead (and this line removed).

jay-hodgson (approved):

PR #4871, opened 2022-05-12
SWC-6021

yjyoo773:


src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/TermsOfUseAccessRequirementWidgetViewImpl.java
```diff
@@ -95,6 +100,16 @@ public void setTerms(String arText) {
...
 
+	@Override
+	public void setAccessRequirementID(String arID) {
+		accessRequirementIDField.setText(arID);
+	}
+
+	@Override
+	public void setAccessRequirementIDVisible() {
+		accessRequirementIDUI.setVisible(true);
+	}
```
nickgros: Could you just make default visibility true and remove this method?

nickgros: Actually maybe not, I see that a non ACT user may see this view and shouldn't see the ID

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/TermsOfUseAccessRequirementWidget.java
```diff
@@ -99,6 +100,7 @@ public void setDataAccessSubmissionStatus(BasicAccessRequirementStatus status) {
 		// set up view based on DataAccessSubmission state
 		if (status.getIsApproved()) {
 			view.showApprovedHeading();
+			view.setAccessRequirementIDVisible();
```
nickgros: Visibility should be determined based on ACT member status, not approval status. See [ManagedACTAccessRequirementWidget](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/5351f7427b88b8752f54ae598a530ec6a18f1d4e/src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidget.java#L123)

src/main/resources/org/sagebionetworks/web/client/widget/accessrequirements/TermsOfUseAccessRequirementWidgetViewImpl.ui.xml
```diff
@@ -7,6 +7,11 @@
 		type='org.sagebionetworks.web.client.IconsImageBundle' />
 	<b:Panel marginBottom="0">
 		<b:PanelBody>
+			<bh:Div ui:field="accessRequirementIDUI" visible="false">
+				<g:InlineLabel addStyleNames="boldText margin-right-5"
+							   text="Click Wrap ID:" />
```
nickgros: Click-wrap ARs share the same ID space as other AR types, so let's label the ID so as not to mislead ACT members into thinking that they are separate.
```suggestion
							   text="Access Requirement ID:" />
```

nickgros (changes requested):
Please change the visibility to show when the caller is an ACT member

src/main/resources/org/sagebionetworks/web/client/widget/accessrequirements/TermsOfUseAccessRequirementWidgetViewImpl.ui.xml
yjyoo773: I asked Ann about this and she wanted it labeled as Click Wrap but if you think Access Requirement ID is better I can go with it.

nickgros: If that's just to make it more clear to ACT that this it's a click wrap, maybe add a line above the ID that says "Click Wrap". It's not a big deal, so I'm ok with Click Wrap ID or whatever makes the most sense

nickgros (approved):
LGTM!

PR #4873, opened 2022-05-17
SWC-6129: If SynapseForum project, fix direct reply link

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/utils/TopicUtils.java
```diff
@@ -4,6 +4,7 @@
 import static org.sagebionetworks.web.client.widget.discussion.ForumWidget.REPLY_ID_KEY;
 import static org.sagebionetworks.web.client.widget.discussion.ForumWidget.THREAD_ID_KEY;
 import org.sagebionetworks.web.client.place.Synapse;
+import org.sagebionetworks.web.client.place.SynapseForumPlace;
```
yjyoo773: Can we remove this?

yjyoo773 (approved):

PR #4874, opened 2022-05-18
SWC-6102

yjyoo773:


src/main/java/org/sagebionetworks/web/client/widget/discussion/SingleDiscussionThreadWidget.java
```diff
@@ -447,6 +447,7 @@ public void onFailure(Throwable caught) {
 			public void onSuccess(DiscussionReplyBundle bundle) {
 				ReplyWidget replyWidget = ginInjector.createReplyWidget();
 				replyWidget.configure(bundle, isCurrentUserModerator, moderatorIds, refreshCallback, isThreadDeleted);
+				replyWidget.asWidget().addStyleName("highlight-reply");
```
jay-hodgson: asWidget() is going to return null in the junit tests.  One solution would be to add a method addStyleName() to the replyWidget, which would call it's view (and in the view implementation it would actually add the style name to the widget)

jay-hodgson (approved):

yjyoo773:
Screen Shot 2022-05-27 at 12 40 48 PM

PR #4875, opened 2022-05-18
SWC-6108: show error if html file is too large to preview

jay-hodgson:


jay-hodgson:
The HtmlPreviewWidget already handles this case more gracefully (shows a friendly error).  Do not try to handle the same case in the PreviewWidget as well.
https://github.com/Sage-Bionetworks/SynapseWebClient/blob/73291d386679670ad4fb6960afb209e7a2cf632d/src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/HtmlPreviewWidget.java#L110

jay-hodgson:
Screen Shot 2022-05-18 at 4 09 45 PM

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/HtmlPreviewWidget.java
```diff
@@ -43,7 +43,7 @@ public class HtmlPreviewWidget implements IsWidget, HtmlPreviewView.Presenter {
 	protected PopupUtilsView popupUtils;
 	protected GWTWrapper gwt;
 
-	public static final double MAX_HTML_FILE_SIZE = 40 * MB;
+	public static final double MAX_HTML_FILE_SIZE = 10 * MB;
```
jay-hodgson, line 46: The max user-provided html file size was set to 5MB in PreviewWidget, and 40MB here.  What do you think about increasing the limit from 5MB to 10MB?  10MB is a pretty large html file (so it may take some time to download and parse), but I don't think this would make Synapse.org look terrible.

xschildw: Sounds good.

xschildw (approved):

PR #4913, opened 2022-06-22
PORTALS-2223

jay-hodgson:


src/main/webapp/sass/_core.scss
```diff
@@ -3321,9 +3321,11 @@ div.border {
 	.SRC-primary-text-color {
 		color: $synapse-blue;
 	}
-	.SRC-primary-background-color-hover:hover {
-		background: $synapse-blue !important;
-		color: white;
+	.SRC-primary-background-color-hover {
+		&:hover, &[aria-expanded="true"] {
```
jay-hodgson, line 3325: Apply style if we're hovering over the item, or (if it's a dropdown button) if it has the aria-expanded attribute set to true.

nickgros (approved):

PR #4916, opened 2022-06-30
SWC-6192

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
```diff
@@ -137,8 +133,6 @@ public void showView(final LoginPlace place) {
 			showTermsOfUse(false);
 		} else if (SHOW_SIGNED_TOU.equals(token) && authenticationController.getCurrentUserAccessToken() != null) {
 			showTermsOfUse(true);
-		} else if (!recognizedTokens.contains(token) && !"".equals(token) && token != null) {
```
jay-hodgson, line 140: Previously assumed that if we did not recognize the token then we should attempt to use it as an access token.  Removed

src/main/java/org/sagebionetworks/web/server/servlet/InitSessionServlet.java
```diff
@@ -51,6 +51,27 @@ public void service(ServletRequest arg0, ServletResponse arg1) throws ServletExc
 		super.service(arg0, arg1);
 	}
 
+	public static Cookie getNewAccessTokenCookie(String accessToken, String requestScheme, String requestServerName) {
```
jay-hodgson, line 54: Refactored code so the same thing can be used during SSO login to set the user_login_token cookie

xschildw (approved):

xschildw (approved):

PR #4932, opened 2022-07-15
SWC-6239: if override marker if MUI pagination component

jay-hodgson:


jay-hodgson:
Also noticed that the base font size is slightly modified in a few cases (to 1.1em).  This is in the project list, the wiki table of contents, and challenge names.  Remove 1.1em specification so the base size is used ([16px for body](https://www.figma.com/file/0oPm5lLSUva8kyfVNMS6FA/Sage-Style-%26-Component-Library?node-id=76%3A1231))

jay-hodgson:
After the fix, verified that unordered lists still have the special marker but the pagination items do not:
Screen Shot 2022-07-15 at 3 44 45 PM

yjyoo773 (approved):

PR #4937, opened 2022-07-18
SWC-6191: Do not enable Add to Download Cart if the current container contains no files

jay-hodgson:


nickgros (approved):
LGTM assuming the tooltip works when you try it

jay-hodgson:
During testing the tooltip worked as expected.  In the case when the menu item was not enabled it showed up, and was not present when the menu item was enabled.

PR #4939, opened 2022-07-19
SWC-6210

nickgros:
I don't think we need experimental mode only for any of this--users can't create Dataset Collections in the UI, even with these changes.

image

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
@@ -10,6 +10,7 @@ import org.sagebionetworks.repo.model.Link;
 import org.sagebionetworks.repo.model.Project;
 import org.sagebionetworks.repo.model.docker.DockerRepository;
 import org.sagebionetworks.repo.model.table.Dataset;
+import org.sagebionetworks.repo.model.table.DatasetCollection;
 import org.sagebionetworks.repo.model.table.EntityView;
 import org.sagebionetworks.repo.model.table.MaterializedView;
 import org.sagebionetworks.repo.model.table.SubmissionView;
@@ -24,6 +25,8 @@ public class EntityTypeUtils {
 	public static final String ENTITY_VIEW_DISPLAY_NAME = "View";
 	public static final String MATERIALIZED_VIEW_DISPLAY_NAME = "Materialized View";
 	public static final String DATASET_DISPLAY_NAME = "Dataset";
+
+	public static final String DATASET_COLLECTION_DISPLAY_NAME = "Dataset Collection";
 	public static final String TABLE_ENTITY_DISPLAY_NAME = "Table";
 	public static final String UNKNOWN_TABLE_TYPE = "Unknown Table Type";
 
@@ -50,6 +53,8 @@ public class EntityTypeUtils {
 				className = MaterializedView.class.getName();
 			} else if (entityType.equalsIgnoreCase(EntityType.dataset.name())) {
 				className = Dataset.class.getName();
+			} else if (entityType.equalsIgnoreCase(EntityType.datasetcollection.name())) {
+				className = DatasetCollection.class.getName();
 			}
 		}
 		return className;
@@ -91,6 +96,8 @@ public class EntityTypeUtils {
 			type = EntityType.materializedview;
 		} else if (Dataset.class.getName().equals(className)) {
 			type = EntityType.dataset;
+		} else if (DatasetCollection.class.getName().equals(className)) {
+			type = EntityType.datasetcollection;
 		}
 		return type;
 	}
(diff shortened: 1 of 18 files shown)


src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
```diff
@@ -127,6 +134,9 @@ public static IconType getIconTypeForEntityClassName(String className) {
 		} else if (Dataset.class.getName().equals(className)) {
 			// Dataset
 			icon = IconType.TABLE;
+		} else if (DatasetCollection.class.getName().equals(className)) {
+			// Dataset
```
nickgros:
```suggestion
```

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
```diff
@@ -380,7 +380,7 @@ private void configureVersionAlert() {
 					goToLatestSnapshot,
 					isLinkToCurrentSnapshotEnabled,
 					linkToCurrentSnapshotTooltipText);
-		} else if (!version.equals(latestSnapshotVersionNumber)) {
+		} else if (version != null && !version.equals(latestSnapshotVersionNumber)) {
```
nickgros, line 383: I don't think this can happen in the real world, but it was causing some non-failing error messages in tests.

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/DatasetsTab.java
```diff
@@ -18,7 +18,7 @@
  */
 public class DatasetsTab extends AbstractTablesTab {
 
-	public static final String DATASETS_HELP = "Create a Draft Dataset and add File versions. Annotate, Mint DOI’s, and Publish your Dataset to share it with others.";
+	public static final String DATASETS_HELP = "Use Datasets to produce and distribute an immutable set of files found across one or more Projects or Folders. You can also create Dataset Collections which contain multiple Datasets .";
```
nickgros:
```suggestion
 	public static final String DATASETS_HELP = "Create a Draft Dataset and add File versions. Annotate, Mint DOI’s, and Publish your Dataset to share it with others.";
```

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/DatasetsTab.java
```diff
@@ -18,7 +18,7 @@
  */
 public class DatasetsTab extends AbstractTablesTab {
 
-	public static final String DATASETS_HELP = "Create a Draft Dataset and add File versions. Annotate, Mint DOI’s, and Publish your Dataset to share it with others.";
+ 	public static final String DATASETS_HELP = "Create a Draft Dataset and add File versions. Annotate, Mint DOI’s, and Publish your Dataset to share it with others.";
```
nickgros:
```suggestion
	public static final String DATASETS_HELP = "Create a Draft Dataset and add File versions. Annotate, Mint DOI’s, and Publish your Dataset to share it with others.";
```

src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
```diff
@@ -127,6 +134,8 @@ public static IconType getIconTypeForEntityClassName(String className) {
 		} else if (Dataset.class.getName().equals(className)) {
 			// Dataset
 			icon = IconType.TABLE;
+		} else if (DatasetCollection.class.getName().equals(className)) {
```
jay-hodgson, line 137: Surprised that this still exists.  Could you open up an SWC ticket to remove references to getIconTypeForEntityClassName().  I think Provenance is the only place that still uses this, and it should just use EntityIcon from SRC

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFilter.java
```diff
@@ -21,15 +23,15 @@
 
 public enum EntityFilter {
 	ALL(EntityType.values()),
-	ALL_BUT_LINK(project, folder, file, table, entityview, dockerrepo, submissionview, dataset),
+	ALL_BUT_LINK(Arrays.stream(EntityType.values()).filter(entityType -> link.equals(entityType)).toArray(EntityType[]::new)),
```
jay-hodgson: This will future-proof ALL_BUT_LINK, well done

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/EntityTypeUtils.java
nickgros: It is indeed only used by Provenance, I checked before adding this. Will do

PR #4959, opened 2022-08-15
SWC-6133

yjyoo773:


yjyoo773:
Screen Shot 2022-08-15 at 9 26 29 AM
Tooltip has the same text as below the title which is mentioned in the ticket. However, I believe it is redundant and should be removed?

yjyoo773:
> Screen Shot 2022-08-15 at 9 26 29 AM

Tooltip has the same text as below the title which is mentioned in the ticket. However, I believe it is redundant and should be removed?

jay-hodgson:
> Screen Shot 2022-08-15 at 9 26 29 AM
> 
> Tooltip has the same text as below the title which is mentioned in the ticket. However, I believe it is redundant and should be removed?

I'd be ok if you removed, but I think the important part is adding this link to the docs.  Approving the current changeset.

jay-hodgson (approved):

PR #4966, opened 2022-08-24
SWC-6267

nickgros:
- Disable the inline scope editor
- All scope edits go through the modal
- Hide the mask options if the mask is not editable.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityViewScopeWidget.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityViewScopeWidget.java
@@ -6,7 +6,6 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.sagebionetworks.repo.model.Entity;
-import org.sagebionetworks.repo.model.EntityRef;
 import org.sagebionetworks.repo.model.Reference;
 import org.sagebionetworks.repo.model.entitybundle.v2.EntityBundle;
 import org.sagebionetworks.repo.model.table.EntityView;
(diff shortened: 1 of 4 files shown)


src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityViewScopeWidget.java
```diff
@@ -66,20 +65,6 @@ public EntityViewScopeWidget(EntityViewScopeWidgetView view, SynapseJavascriptCl
 		view.setSynAlert(synAlert.asWidget());
 	}
 
-	private List<Reference> getReferencesFromDatasetItems(List<EntityRef> datasetItems) {
```
nickgros, line 69: Method was unused prior to these changes.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityViewScopeWidget.java
```diff
@@ -104,18 +89,8 @@ public void configure(EntityBundle bundle, boolean isEditable) {
 			tableType = TableType.getTableType(currentView);
 			List<Reference> references = getReferencesFromIdList(currentView.getScopeIds());
```
nickgros, line 91: We were trying to do something funky with two editors--a modal where you could edit the mask and scope, or an inline editor if you could just edit the scope. The inline editor never wrote updates to the backend, so it was impossible to update a project view (or any view which has a mask that you cannot update in the UI).

Instead, always use the modal, and never show the inline edit options.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/EntityViewScopeWidget.java
```diff
@@ -172,12 +147,20 @@ public void onEditScopeAndMask() {
...
-		view.setIsFolderSelected(tableType.isIncludeFolders());
-		view.setIsTableSelected(tableType.isIncludeTables());
-		view.setIsDatasetSelected(tableType.isIncludeDatasets());
+		// The mask may not be editable since not all masks are supported in the web client
+		boolean isMaskEditable = isEditable && isEditMaskSupportedInWebClient(tableType);
+		if (!isMaskEditable) {
+			synAlert.consoleError("View type mask is not supported by web client, blocking edit. Mask value:" + ((EntityView) currentView).getViewTypeMask());
+		}
 
+		view.setEditMaskVisible(isMaskEditable);
```
nickgros, line 156: Hide the mask editor in the modal if the mask cannot be edited via UI.

jay-hodgson (approved):

PR #4969, opened 2022-08-26
SWC-6181

nickgros:
- Rename ReactFunctionComponent to ReactComponentType to match @types/react
- Rename ReactElement to ReactNode to match @types/react
- Replace many instances of JSNI React render code with JSInterop
- Remove all calls to ReactDOM.render
- Change ReactComponentDiv/Span to utilize React 18 API
- Exclusively manage React component lifecycle in ReactComponentDiv/Span


src/main/java/org/sagebionetworks/web/client/widget/ReactComponentDiv.java
```diff
@@ -1,20 +1,56 @@
...
+import org.sagebionetworks.web.client.jsinterop.ReactDOMRoot;
+import org.sagebionetworks.web.client.jsinterop.ReactNode;
+
+import com.google.gwt.user.client.ui.HTMLPanel;
 
 /**
  * Automatically unmounts the ReactComponent (if any) inside this div when this container is detached/unloaded.
  */
-public class ReactComponentDiv extends Div {
+public class ReactComponentDiv extends HTMLPanel {
```
nickgros: Use HTMLPanel instead of Div (no need to use gwtbootstrap3's div)

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentSpan.java
```diff
@@ -1,20 +1,50 @@
...
+import org.sagebionetworks.web.client.jsinterop.ReactDOMRoot;
+import org.sagebionetworks.web.client.jsinterop.ReactNode;
+
+import com.google.gwt.user.client.ui.InlineHTML;
 
 /**
  * Automatically unmounts the ReactComponent (if any) inside this div when this container is detached/unloaded.
  */
-public class ReactComponentSpan extends Span {
+public class ReactComponentSpan extends InlineHTML {
```
nickgros: Use InlineHTML insead of gwtboostrap3's Span

src/main/java/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.java
```diff
@@ -102,22 +102,23 @@ public void setPresenter(Presenter presenter) {
 		refresh();
 	}
 
-	private void detachNavBar() {
```
nickgros, line 105: Since the ReactComponentDiv can now hold onto the `root` created by ReactDOM, attaching and detaching caused undesired flashing. Let's just show/hide it and rerender with new props instead of completely detaching and re-attaching.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentDiv.java
```diff
@@ -1,20 +1,56 @@
...
+
+	private ReactDOMRoot root;
+	private ReactNode component;
```
nickgros, line 14: Need to store the component because sometimes this Widget could unload and load without being re-configured

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentDiv.java
jay-hodgson: GWT FlowPanel is even lighter weight (if you want to move away from gwtbootstrap3 Div and don't require the HTMLPanel html parsing functionality)

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentSpan.java
jay-hodgson: Would the GWT InlineLabel (basic span) work here?

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/PlotlyWidgetViewImpl.java
```diff
@@ -126,7 +126,9 @@ private static native void _showChart(Element el, JavaScriptObject[] xyData, Str
 					height : "100%"
 				}
 			};
-			$wnd.ReactDOM.render($wnd.React.createElement(plot, props), el);
+
+            var component = $wnd.React.createElement(plot, props)
+            reactComponentDiv.@org.sagebionetworks.web.client.widget.ReactComponentDiv::render(Lorg/sagebionetworks/web/client/jsinterop/ReactNode;)(component);
```
jay-hodgson: I was surprised to find new ugly jsni in this PR (where so much code has been converted to nice jsinterop), but this should work

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/SRCDemoWidgetViewImpl.java
```diff
@@ -55,10 +55,10 @@ public void setSynAlertWidget(IsWidget w) {
 	@Override
 	public void setDemoVisible(boolean visible) {
```
jay-hodgson: Is this still in here?!  You could delete SRCDemoWidget (and view and view impl) and remove the reference from the ComingSoon page in this PR if you'd rather

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/TabViewImpl.java
```diff
@@ -34,6 +34,8 @@ public class TabViewImpl implements TabView {
 	Div contentDiv;
 	boolean isActive = false;
 
+
+
```
jay-hodgson: Extra whitespace could be removed

src/main/java/org/sagebionetworks/web/client/widget/header/HeaderViewImpl.java
```diff
@@ -102,22 +102,23 @@ public void setPresenter(Presenter presenter) {
...
-		synapseNavDrawerContainer.removeFromParent();
+	private void hideNavBar() {
+		synapseNavDrawerContainer.addStyleName("hidden");
```
jay-hodgson: I think .setVisible() is preferable, since it will set the display property and the aria-hidden element attribute.

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentSpan.java
nickgros: Yes! I'm not very familiar with picking widgets that let me do things as simply as possible 🙂

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/PlotlyWidgetViewImpl.java
nickgros: I got lazy and didn't feel like writing JsInterop code for nested objects--this was much easier

PR #4970, opened 2022-08-29
SWC-6249

nickgros:
Force paragraph above entity finder to wrap instead of stretching modal.

The modal needs to stretch because the entity finder can expand when "search" is clicked.

Confirmed that this "hack" works in Chrome, Firefox, and Safari

image

### DIFF
--- a/src/main/webapp/sass/_core.scss
+++ b/src/main/webapp/sass/_core.scss
@@ -3658,3 +3658,9 @@ div.border {
 #body .markdown .SynapseTable {
 	min-height: unset;
 }
+
+// https://stackoverflow.com/a/57599409
+.fit-within-parent {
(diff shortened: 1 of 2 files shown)


src/main/webapp/sass/_core.scss
```diff
@@ -3657,4 +3657,10 @@ div.border {
 // override SynapseTable style when embedded in a wiki
 #body .markdown .SynapseTable {
 	min-height: unset;
+}
+
+// https://stackoverflow.com/a/57599409
+.fit-within-parent {
+  width: 0;
```
jay-hodgson, line 3664: Got it, css sorcery

jay-hodgson (approved):

PR #4972, opened 2022-08-31
SWC-6077 - Handle breaking change in EntityFinder

nickgros:
Account for EntityFinder versionSelection breaking change in synapse-react-client. Needs new version of SRC


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderWidgetViewImpl.java
```diff
@@ -117,16 +117,15 @@ public void renderComponent(EntityFinderScope initialScope, EntityFinderWidget.I
 				EntityFinderProps.create(
 						onSelected,
 						multiSelect,
-						showVersions,
+						versionSelection.name(),
```
jay-hodgson, line 120: Ah, there's the enum to string value that I was looking for :)

jay-hodgson (approved):

PR #4975, opened 2022-09-02
SWC-6272

yjyoo773:


jay-hodgson:
The problem with creating a new place is that all links must be updated.  That includes from [the docs site](https://help.synapse.org/docs/Synapse-User-Account-Types.2007072795.html) and all references in the codebase (like [this one](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/9e9a07c11282c6fc0f115d6a0c2ee90f89b6f1b4/src/main/java/org/sagebionetworks/web/client/widget/entity/download/QuizInfoDialog.java#L46)).
Couple of possible solutions:
1.  Add redirect to your new place in the existing QuizPresenter (if in alpha mode until release), and open up tickets to change each reference so this redirect is being used less and less over time.
2. Change QuizPresenter to use your your new view if in alpha, but the old view if not.

jay-hodgson (changes requested):
Changes requested (see notes)

src/main/java/org/sagebionetworks/web/client/presenter/QuizPresenter.java
```diff
@@ -227,8 +237,12 @@ public String mayStop() {
 
 	@Override
 	public void setPlace(org.sagebionetworks.web.client.place.Quiz place) {
-		view.setPresenter(this);
-		view.clear();
+		if(DisplayUtils.isInTestWebsite(ginInjector.getCookieProvider())){
+			this.place = place;
```
jay-hodgson: Is storing the place necessary?  If so, safe to always store it (not just in alpha mode)

src/test/java/org/sagebionetworks/web/unitclient/presenter/QuizPresenterTest.java
```diff
@@ -63,15 +68,18 @@ public class QuizPresenterTest {
...
+		mockSRCView = mock(CertificationQuizView.class);
 		mockSynapseClient = mock(SynapseClientAsync.class);
 		mockAuthenticationController = mock(AuthenticationController.class);
 		mockGlobalApplicationState = mock(GlobalApplicationState.class);
 		mockQuestionContainer = mock(QuestionContainerWidget.class);
 		mockInjector = mock(PortalGinInjector.class);
 		mockPassingRecord = mock(PassingRecord.class);
 		mockQuestionResponse = mock(MultichoiceResponse.class);
+		mockCookieProvider = mock(CookieProvider.class);
+		when(mockInjector.getCookieProvider()).thenReturn(mockCookieProvider);
```
jay-hodgson, line 87: Expect for all existing tests, so I'd expect this to be in the before: 
 when(mockCookieProvider.getCookie(DisplayUtils.SYNAPSE_TEST_WEBSITE_COOKIE_KEY)).thenReturn(null)

There should be a new test though, where this cookie is set and you verify your new SRC view is used.

jay-hodgson (approved):

PR #4985, opened 2022-10-06
SWC-6288

nickgros:
Clear the QueryClient cache when the authentication token changes, fixing SWC-6288.

SWC-6288 is caused by the following sequence:

1. A non-ACT user opens the new Governance Dashboard. The authentication token is stored as a cookie, but has not been loaded by the AuthenticationControllerImpl
2. The Governance Dashboard React component is loaded. Since the AuthenticationControllerImpl has not stored the auth token yet, the component does not use an auth token
3. An unauthenticated request is made to get open access submissions. The response is empty, and is stored in the cache
4. The authentication token is loaded by the AuthenticationControllerImpl
5. The authentication token is passed to the React component
6. The access submission hook re-runs and gets a cache hit on the empty response

Now, between (4) and (5), the cache is cleared.

Potential follow up work:
* Figure out why the React component is loaded before the AuthenticationController receives the token--not sure if this is working as intended or how this process is supposed to work.
* In synapse-react-client, make the QueryClient cache sensitive to a changing authentication token

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/context/QueryClientProvider.java
+++ b/src/main/java/org/sagebionetworks/web/client/context/QueryClientProvider.java
@@ -1,0 +1,13 @@
+package org.sagebionetworks.web.client.context;
+
+import org.sagebionetworks.web.client.jsinterop.reactquery.QueryClient;
+
+/**
+ * Provides a singleton react-query QueryClient.
+ */
+public interface QueryClientProvider {
+	/**
+	 * Returns the global QueryClient
+	 */
+	QueryClient getQueryClient();
+}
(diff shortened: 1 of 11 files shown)


src/main/java/org/sagebionetworks/web/client/context/QueryClientProvider.java
```diff
@@ -0,0 +1,13 @@
+package org.sagebionetworks.web.client.context;
+
+import org.sagebionetworks.web.client.jsinterop.reactquery.QueryClient;
+
+/**
+ * Provides a singleton react-query QueryClient.
+ */
+public interface QueryClientProvider {
```
nickgros, line 8: Pulled out of SynapseContextPropsProvider to break cycle

would have been
```
SynapseContextPropsProvider -> AuthenticationController -> SynapseContextPropsProvider
```
and is now
```
SynapseContextPropsProvider -> AuthenticationController -> QueryClientProvider
SynapseContextPropsProvider -> QueryClientProvider
```

jay-hodgson (approved):

PR #4990, opened 2022-10-13
PORTALS-2377

nickgros:
Fix bug where EntityFinderBuilder would reuse dependencies passed to each built EntityFinderWidget

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderWidgetImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderWidgetImpl.java
@@ -14,6 +14,7 @@ import org.sagebionetworks.repo.model.entitybundle.v2.EntityBundleRequest;
 import org.sagebionetworks.web.client.DisplayConstants;
 import org.sagebionetworks.web.client.GlobalApplicationState;
 import org.sagebionetworks.web.client.PopupUtilsView;
+import org.sagebionetworks.web.client.PortalGinInjector;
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderWidgetImpl.java
```diff
@@ -63,15 +64,15 @@ public EntityFinderWidgetImpl(final EntityFinderWidgetView view, final GlobalApp
         view.setPresenter(this);
     }
 
-    private EntityFinderWidgetImpl(Builder builder) {
+    private EntityFinderWidgetImpl(final Builder builder, final PortalGinInjector ginInjector) {
         this.selectedEntities = new ArrayList<>();
 
-        // Pass along dependencies injected into the builder because they are not injected in this constructor.
-        this.view = builder.view;
```
nickgros, line 70: Bug was caused by the builder re-using the same view instance for each built EntityFinderWidget

jay-hodgson:
What do you think about patching 426?

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderWidgetImpl.java
jay-hodgson: This is a nice simplification of the code, let the GIN (Guice) binding determine which objects should be singletons

PR #4993, opened 2022-10-17
SWC-6300

nickgros:


src/main/java/org/sagebionetworks/web/client/widget/entity/browse/EntityFinderWidgetViewImpl.java
```diff
@@ -170,7 +170,6 @@ public void clearError() {
 
 	@Override
 	public void hide() {
-		this.jsniUtils.unmountComponentAtNode(entityFinderContainer.getElement());
```
nickgros, line 173: Should not be used with React 18

src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
```diff
@@ -108,8 +108,6 @@ public interface SynapseJSNIUtils {
 
 	void scrollIntoView(Element el);
 
-	void unmountComponentAtNode(Element el);
-
```
nickgros, line 112: Unused after changes

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentDiv.java
```diff
@@ -39,6 +40,13 @@ protected void onUnload() {
 		super.onUnload();
 	}
 
+	@Override
+	public void clear() {
```
nickgros, line 44: This will fix it. In the relevant component, we call `clear` expecting it to unmount the component tree, but that never happens before this change.

jay-hodgson (approved):

PR #5001, opened 2022-10-26
SWC-6314: Make "Anyone on the web" (public group) permission set not …

jay-hodgson:
…editable in the UI, and render Can Download if OPEN_DATA

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/sharing/SharingPermissionsGridViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/sharing/SharingPermissionsGridViewImpl.java
@@ -99,37 +99,44 @@ public class SharingPermissionsGridViewImpl extends Composite implements Sharing
 		permListBox.addStyleName("input-xs");
 		data = new TableData();
 		row.add(data);
-		if (!deleteButtonVisible || deleteButtonCallback == null) {
-			// Don't allow editing the permissions and don't add delete button.
-			String selectedItemText = permListBox.getSelectedItemText();
-			// SWC-6083: If this is the ACL entry of the public group
-			// and this is not editable (deleteButtonCallback == null),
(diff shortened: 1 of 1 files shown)


src/main/java/org/sagebionetworks/web/client/widget/sharing/SharingPermissionsGridViewImpl.java
```diff
@@ -99,37 +99,44 @@ private TableRow createAclEntryTableRow(final AclEntry aclEntry, PermissionLevel
...
+		boolean isDeleteButtonAvailable = deleteButtonVisible && deleteButtonCallback != null;
+		Widget permissionWidget;
+		
+		if (aclEntry.getOwnerId().equals(publicAclPrincipalId)) {
+			// SWC-6314 (and SWC-6083): If this is the public group, render Can Download if OPEN_DATA, otherwise Can View.
+			if (isOpenData) {
+				permissionWidget = new Text(permissionDisplay.get(PermissionLevel.CAN_DOWNLOAD));
 			} else {
-				data.add(new Text(selectedItemText));	
+				permissionWidget = new Text(permListBox.getSelectedItemText());
```
xschildw, line 110: Maybe missing something... Should you not be able to edit if you're admin?

jay-hodgson: Usually yes.  However, this is the "Anyone on the web" entry in the ACL, and the backend only allows the READ permission (rendered as "Can view").

xschildw: Got it, thought it'd be needed to fix a bad permission given to public...

xschildw (approved):

PR #5005, opened 2022-11-02
SWC-6323

nickgros:
- QueryClient cache was being improperly cleared every 10s.


jay-hodgson (approved):

jay-hodgson:
As noted (by Nick) this is still correctly called when a user change is detected (in checkForUserChange())

PR #5009, opened 2022-11-04
SWC-6326

nickgros:
- Check if user is logged in before sending them to LoginPlace


jay-hodgson (approved):

jay-hodgson:
This seems like a reasonable thing to do.

PR #5011, opened 2022-11-07
SWC-6180

nickgros:
- Account for breaking change in Sage-Bionetworks/Synapse-React-Client#1880

Requires new release of SRC


jay-hodgson:
Does staging (stack 430) need to be patched?

jay-hodgson (approved):

PR #5012, opened 2022-11-07
SWC-6312: Integrate EntityPageBreadcrumbs

nickgros:
Depends on #5011, Sage-Bionetworks/Synapse-React-Client#1887, and a new version of SRC

- Use new EntityPageBreadcrumbs
- Ensure components without Synapse authentication context still receive theme context
- Adjust tab content layouts to ensure breadcrumbs are flush with the tabs

https://user-images.githubusercontent.com/17580037/200374627-aff15244-9bff-4a99-8dfe-a464e16382f9.mov

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
@@ -41,6 +41,8 @@ import org.sagebionetworks.repo.model.entitybundle.v2.EntityBundle;
 import org.sagebionetworks.repo.model.file.CloudProviderFileHandleInterface;
 import org.sagebionetworks.repo.model.file.FileHandle;
 import org.sagebionetworks.web.client.cookie.CookieProvider;
+import org.sagebionetworks.web.client.jsinterop.React;
+import org.sagebionetworks.web.client.jsinterop.ReactMouseEvent;
 import org.sagebionetworks.web.client.jsinterop.SRC;
 import org.sagebionetworks.web.client.jsinterop.ToastMessageOptions;
 import org.sagebionetworks.web.client.place.PeopleSearch;
@@ -1079,6 +1081,13 @@ public class DisplayUtils {
 			return false;
 		}
 		return event.isAltKeyDown() || event.isControlKeyDown() || event.isMetaKeyDown() || event.isShiftKeyDown();
+	}
+
+	public static boolean isAnyModifierKeyDown(ReactMouseEvent event) {
+		if (event == null) {
+			return false;
+		}
+		return event.altKey || event.ctrlKey || event.metaKey || event.shiftKey;
 	}
 
 	public static String capitalize(String s) {
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
@@ -12,8 +12,6 @@ public class React {
 
 	public static native <P extends ReactComponentProps> ReactNode createElement(ReactComponentType<P> component, P props, ReactNode[] children);
 
-	public static native <P extends ReactComponentProps> ReactNode createElement(ReactComponentType<P> component, P props, ReactNode child1, ReactNode child2);
-
 	/**
 	 * Similar to {@link #createElementWithSynapseContext} but only includes the theme. Any components rendered will NOT get the full Synapse context, including
 	 * access token + auth state, experimental mode status, and time display settings.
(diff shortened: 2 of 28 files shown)


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -1081,6 +1083,13 @@ public static boolean isAnyModifierKeyDown(ClickEvent event) {
 		return event.isAltKeyDown() || event.isControlKeyDown() || event.isMetaKeyDown() || event.isShiftKeyDown();
 	}
 
+	public static boolean isAnyModifierKeyDown(ReactMouseEvent event) {
```
nickgros, line 1086: Same as method right above this one, but works with the React synthetic event

src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
```diff
@@ -12,6 +12,19 @@ public class React {
...
+
+	/**
+	 * Similar to {@link #createElementWithSynapseContext} but only includes the theme. Any components rendered will NOT get the full Synapse context, including
+	 * access token + auth state, experimental mode status, and time display settings.
+	 */
+	@JsOverlay
+	public static <P extends ReactComponentProps> ReactNode createElementWithThemeContext(ReactComponentType<P> component, P props) {
+		SynapseContextProviderProps emptyContext = SynapseContextProviderProps.create(SynapseContextJsObject.create(null, false, false), null);
+		return createElementWithSynapseContext(component, props, emptyContext);
+	}
```
nickgros, line 23: Since our MUI components now use the theme context instead of inline styles, components that didn't get context lost their styles.

One example of this is FullWidthAlert. The FullWidthAlert widget cannot have the context props injected by Gin without a large refactor because we use it in UI XML.

To ensure it gets wrapped in theme context, we use this overlay method instead of plain React.createElement. All usage of React.createElement was replaced with React.createElementWithThemeContext.

src/main/resources/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTabViewImpl.ui.xml
```diff
@@ -4,15 +4,15 @@
 	xmlns:b="urn:import:org.gwtbootstrap3.client.ui"
 	xmlns:bh="urn:import:org.gwtbootstrap3.client.ui.html">
 	<bh:Div>
-		<b:Row ui:field="adminTabContainer" addStyleNames="challengeTabContent" >
-			<b:Column size="XS_12">
+        <b:Column size="XS_12" addStyleNames="margin-top-15">
+		    <b:Row ui:field="adminTabContainer">
```
nickgros: BS3 rows should be wrapped in columns (they have a negative margin that the column offsets)

src/main/resources/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTabViewImpl.ui.xml
```diff
@@ -3,16 +3,14 @@
 	xmlns:g="urn:import:com.google.gwt.user.client.ui"
 	xmlns:b="urn:import:org.gwtbootstrap3.client.ui"
 	xmlns:bh="urn:import:org.gwtbootstrap3.client.ui.html">
-	<bh:Div>
-		<b:Row ui:field="adminTabContainer" addStyleNames="challengeTabContent" >
```
nickgros, line 7: In general swap Bootstrap3 rows/columns for FlowPanel because we aren't using the grid/responsive features for these tab panes.

I did this primarily to simplify dealing with side margins

src/main/java/org/sagebionetworks/web/client/widget/breadcrumb/BreadcrumbViewImpl.java
```diff
@@ -46,41 +46,33 @@ public void setLinksList(List<LinkData> breadcrumbs) {
...
+			return BreadcrumbItem.create(data.getText(), false, href, clickHandler);
+		}).collect(Collectors.toList());
+		// If there's a "current" item, add it to the end of the list
 		if (current != null) {
-			current = stubString(current);
-			SafeHtmlBuilder shb = new SafeHtmlBuilder();
-			shb.appendHtmlConstant(ClientProperties.BREADCRUMB_SEP);
-			shb.appendEscaped(current);
-			panel.add(new InlineHTML(shb.toSafeHtml()));
+			items.add(BreadcrumbItem.create(current, true, null, null));
```
jay-hodgson, line 68: Right, need to add the last item (that does not link).

jay-hodgson (approved):

PR #5022, opened 2022-11-30
SWC-6334: add social media meta tags to synapse.org

jay-hodgson:


jay-hodgson:
Note that the new svg is a copy of logo.svg with a background style applied.  I tested logo.svg on https://www.opengraph.xyz/, and the previews looked great (but I thought a blue background might be preferable).

nickgros (approved):

tschaffter (approved):

PR #5023, opened 2022-12-01
SWC-6335

jay-hodgson:


jay-hodgson:
Screen Shot 2022-11-30 at 5 40 29 PM

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -445,6 +479,72 @@ private String getEntityHtml(String entityId)
...
+
+      object = new JSONObjectAdapterImpl();
+      object.put("@type", "Person");
+      UserProfile profile = synapseClient.getUserProfile(ds.getCreatedBy());
+      object.put("name", getDisplayName(profile));
+      object.put(
+        "url",
+        "https://www.synapse.org/#!Profile:" + ds.getCreatedBy()
+      );
+      json.put("creator", object);
```
nickgros: The Dataset owners can change the name and the description, but they have no control over this field.

Is there risk in surfacing datasets with the user who created it as an author when semantically it should be a consortia or different individual (e.g. the PI)? Proper attribution is critical to our mission, so I'd rather not have this than get this wrong, unless we either provide a way to override it, or to opt out of surfacing this document.

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -445,6 +479,72 @@ private String getEntityHtml(String entityId)
...
+        array.put(index++, String.join(", ", keyValuePairList));
+      }
+      json.put("keywords", array);
+
+      // include identifier if there is a DOI association?
+
+      JSONObjectAdapter object = new JSONObjectAdapterImpl();
+      object.put("@type", "DataCatalog");
+      object.put("name", "Synapse");
+      object.put("url", "https://www.synapse.org");
```
nickgros, line 531: Would be cool if we had a specific instance of this Synapse DataCatalog object in the semantic web, and then maybe Google has a feature where you could reverse index and find all public datasets in Synapse!

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -332,6 +349,20 @@ private String getEntityHtml(String entityId)
       bundleRequest
     );
     Entity entity = bundle.getEntity();
+    if (entity instanceof Dataset) {
+      // attempt to get the latest stable version (instead of the draft)
+      try {
+        bundle =
+          synapseClient.getEntityBundleForVersion(
+            entityId,
+            ((Dataset) entity).getVersionNumber() - 1,
```
nickgros: May fail if the version was deleted. Up to you if you think it’s a big deal enough to fix; otherwise, would suggest putting a comment

nickgros (approved):

PR #5024, opened 2022-12-01
SWC-6334

jay-hodgson:
Likely lesson learned from portals, use bitmap image instead of svg


jay-hodgson:
I'm having fun using Inkscape these days!

src/main/webapp/Portal.html
```diff
@@ -25,7 +25,7 @@
     />
     <meta
       property="og:image"
```
tschaffter, line 28:
```suggestion
      name="image"
      property="og:image"
```

tschaffter (approved):
Only one comment, otherwise LGTM.

src/main/webapp/Portal.html
jay-hodgson: You're too fast at reviewing :)
Change made.

tschaffter (approved):

PR #5039, opened 2023-01-06
SWC-6317

jay-hodgson:
- Use HtmlPreview from SRC
- Move logic from HtmlPreview to NbConvertPreviewWidget (since it previously extended HtmlPreviewWidget)
- Clean-up after move.  For example, the NbConvertPreviewWidget always needs to show the "limited rendering only" UI because it is non-interactive.   User must download the file and run the server locally to interact with it.  But it may still be sanitized, depending on the createdBy user who uploaded the file.


jay-hodgson:
Note that I am using a local SRC umd build based on code from the SRC develop branch.
This is using the HtmlPreview component from SRC:

Screen Shot 2023-01-06 at 12 22 06 PM

Screen Shot 2023-01-06 at 12 21 52 PM
Screen Shot 2023-01-06 at 12 09 54 PM
Screen Shot 2023-01-06 at 12 11 02 PM

nickgros (approved):

PR #5041, opened 2023-01-06
SWC-6360: use a plain text version of the wiki for the dataset descri…

jay-hodgson:
…ption

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
@@ -372,8 +372,9 @@ public class CrawlFilter extends OncePerRequestFilter {
     Annotations annotations = bundle.getAnnotations();
     String name = escapeHtml(entity.getName());
     String description = escapeHtml(entity.getDescription());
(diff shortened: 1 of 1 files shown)


jay-hodgson:
With the change:
Screen Shot 2023-01-06 at 3 19 35 PM

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -495,7 +500,9 @@ public String getDatasetScriptElement(EntityBundle bundle)
       json.put("@context", "http://schema.org/");
       json.put("@type", "Dataset");
       json.put("name", ds.getName());
-      json.put("description", ds.getDescription());
```
nickgros, line 498: Consider keeping as a fallback if there's no wiki

nickgros (approved):

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
jay-hodgson: Done!

PR #5042, opened 2023-01-11
SWC-6359: add support for the new MEDIUMTEXT column type, and remove …

jay-hodgson:
…some dead code now that we rely on SRC to render tables/views


jay-hodgson:
Need to support cell renderers for default entity view columns, shown here:
Screen Shot 2023-01-11 at 2 58 52 PM

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelTableRowEditorWidgetImpl.java
```diff
@@ -260,12 +269,25 @@ private String getMaxSizeForType(ColumnTypeViewEnum type) {
...
   ) {
     view.setSelectionPresenter(selectionPresenter);
+
+    // add column types to the dropdown:
+    Arrays
+      .asList(ColumnTypeViewEnum.values())
+      .stream()
+      .filter((val -> (!isView || !unsupportedTypesForViews.contains(val)))) // keep all values for tables, or keep value if not found in unsupportedTypesForViews.
+      .forEach(val -> {
+        view.addColumnType(val);
```
jay-hodgson, line 289: https://sagebionetworks.jira.com/browse/SWC-6333

jay-hodgson: Removed the list from the ui.xml, and add dynamically.  If View, do not include MediumText or LargeText

src/test/java/org/sagebionetworks/web/unitclient/widget/table/v2/results/facets/FacetsWidgetTest.java
```diff
@@ -177,7 +176,7 @@ public void testConfigureFromQuery() {
     verify(mockView).add(mockJobTrackingWidget);
     verify(mockJobTrackingWidget)
       .startAndTrackJob(
-        eq(TableQueryResultWidget.RUNNING_QUERY_MESSAGE),
+        eq(""),
```
nickgros: I think FacetsWidget can also be deleted

nickgros (approved):
Core change looks good, there might be more dead code to clean up

nickgros (approved):
That was a lot more dead code than I was expecting...awesome!

PR #5043, opened 2023-01-13
SWC-6356

nickgros:
- Remove ReviewerDashboardProps, no longer needed.
- Refactor to remove unnecessary code
- Backfill missing tests for DataAccessManagementPresenter


src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
```diff
@@ -34,7 +34,7 @@ public static class SynapseComponents {
     public static ReactComponentType<QueryWrapperPlotNavProps> QueryWrapperPlotNav;
     public static ReactComponentType<StandaloneQueryWrapperProps> StandaloneQueryWrapper;
     public static ReactComponentType<ForumSearchProps> ForumSearch;
-    public static ReactComponentType<ReviewerDashboardProps> ReviewerDashboard;
```
nickgros, line 37: ReviewerDashboardProps is no longer needed in SWC. The only configured prop enabled us to use the old submission rejection UI.

src/test/java/org/sagebionetworks/web/unitclient/presenter/DataAccessManagementPresenterTest.java
```diff
@@ -1,7 +1,14 @@
 package org.sagebionetworks.web.unitclient.presenter;
```
nickgros, line 1: Noticed this test suite was missing tests, so I backfilled a few

src/main/java/org/sagebionetworks/web/client/view/DataAccessManagementView.java
```diff
@@ -1,15 +1,7 @@
...
 public interface DataAccessManagementView extends IsWidget {
-  /**
-   * Set this view's presenter
-   *
-   * @param presenter
-   */
-  public void setPresenter(DataAccessManagementPresenter presenter);
-
-  public void render();
+  void render();
```
nickgros, line 6: The view never used the presenter. It can be simplified to just have a `render` method.

jay-hodgson (approved):

PR #5046, opened 2023-01-20
SWC-6349: Use new CreatedByModifiedBy component

nickgros:
- Use new CreatedByModifiedBy component from SRC for ModifiedByCreatedByWidget
- Change `configure` method on ModifiedCreatedByWidget to accomodate props
- SCSS: Use variables and account for changed footer height in determining rootPanel height
- Move ModifiedByCreatedByWidget location to be flush with footer
- Remove unused `addToDownloadListContainer` field

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/QueryWrapperPlotNavProps.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/QueryWrapperPlotNavProps.java
@@ -66,7 +66,7 @@ public class QueryWrapperPlotNavProps extends ReactComponentProps {
     props.shouldDeepLink = false;
     props.name = "Items";
     props.downloadCartPageUrl = "#!DownloadCart:0";
-    props.showLastUpdatedOn = true;
+    props.showLastUpdatedOn = false;
     // SWC-6138 - hide charts by default
     props.defaultShowFacetVisualization = false;
     return props;
--- a/src/main/resources/org/sagebionetworks/web/client/widget/docker/DockerRepoWidgetViewImpl.ui.xml
+++ b/src/main/resources/org/sagebionetworks/web/client/widget/docker/DockerRepoWidgetViewImpl.ui.xml
@@ -67,13 +67,8 @@
             </g:FlowPanel>
           </b:Column>
         </b:Row>
-        <b:Row>
-          <g:SimplePanel
-            addStyleNames="margin-bottom-15 margin-left-15"
-            ui:field="dockerModifiedAndCreatedContainer"
-          />
-        </b:Row>
       </g:FlowPanel>
+      <g:SimplePanel ui:field="dockerModifiedAndCreatedContainer" />
     </b:Column>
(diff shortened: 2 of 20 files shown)


src/main/java/org/sagebionetworks/web/client/jsinterop/QueryWrapperPlotNavProps.java
```diff
@@ -66,7 +66,7 @@ public static QueryWrapperPlotNavProps create(
     props.shouldDeepLink = false;
     props.name = "Items";
     props.downloadCartPageUrl = "#!DownloadCart:0";
-    props.showLastUpdatedOn = true;
+    props.showLastUpdatedOn = false;
```
nickgros, line 69: Hide the `lastUpdatedOn` in the QueryWrapperPlotNav because the new component now shows it. Behavior unchanged for StandaloneQueryWrapper, which is not shown on an entity page, so the information is still visible.

src/main/resources/org/sagebionetworks/web/client/widget/docker/DockerRepoWidgetViewImpl.ui.xml
```diff
@@ -67,13 +67,8 @@
...
           </b:Column>
         </b:Row>
-        <b:Row>
-          <g:SimplePanel
-            addStyleNames="margin-bottom-15 margin-left-15"
-            ui:field="dockerModifiedAndCreatedContainer"
-          />
-        </b:Row>
       </g:FlowPanel>
+      <g:SimplePanel ui:field="dockerModifiedAndCreatedContainer" />
```
nickgros, line 71: Move so it's flush with the footer and doesn't have undesired margin. Same change for files and tables pages.

src/main/resources/org/sagebionetworks/web/client/widget/entity/EntityPageTopViewImpl.ui.xml
```diff
@@ -27,9 +27,5 @@
       </g:FlowPanel>
     </bh:Div>
     <bh:Div ui:field="tabsUI" />
-    <bh:Div
-      addStyleNames="margin-top-10"
-      ui:field="addToDownloadListContainer"
-    />
```
nickgros, line 33: This was unused/unreferenced, and created an undesirable margin between the rootPanel and the footer

src/main/webapp/sass/_core.scss
```diff
@@ -52,7 +52,12 @@ $synapse-tertiary-color-palette: (
   900: #2f2814,
 );
 
-$full-root-panel-height: calc(100vh - 93px - 130px);
+$page-header-min-height: 93px;
+$footer-height: 160px;
```
nickgros, line 56: This is not used anywhere (the footer height is dynamic). It's changed since it's last been set, so while this doesn't permanently fix the problem, it at least looks better now. With the underestimated value, there could be whitespace at the bottom of the page if the combined heights of header + rootPanel + footer was less than the window height.

jay-hodgson (approved):

PR #5048, opened 2023-01-23
SWC-6377: Fix case where the profile is not loaded when initially signing terms of use

nickgros:
- Call clearQueryClientCache when the accessToken changes
- Remove redundant calls

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
@@ -103,7 +103,6 @@ public class AuthenticationControllerImpl implements AuthenticationController {
           public void onSuccess(LoginResponse response) {
             storeAuthenticationReceipt(response.getAuthenticationReceipt());
             setNewAccessToken(response.getAccessToken(), callback);
-            clearQueryClientCache();
           }
 
           @Override
@@ -198,6 +197,7 @@ public class AuthenticationControllerImpl implements AuthenticationController {
             DateTimeUtilsImpl.getWeekFromNow()
           );
           currentUserAccessToken = accessToken;
+          clearQueryClientCache();
           userAccountService.getMyProfile(
             new AsyncCallback<UserProfile>() {
               @Override
@@ -236,6 +236,7 @@ public class AuthenticationControllerImpl implements AuthenticationController {
         public void onFailure(Throwable t) {
           currentUserAccessToken = null;
           currentUserProfile = null;
+          clearQueryClientCache();
           ginInjector.getSessionDetector().initializeAccessTokenState();
           callback.onFailure(t);
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -198,6 +197,7 @@ public void onSuccess(String accessToken) {
             DateTimeUtilsImpl.getWeekFromNow()
           );
           currentUserAccessToken = accessToken;
+          clearQueryClientCache();
```
nickgros, line 200: Regardless of success or failure, when we set the accessToken, clear the cache. This fixes the issue because this method is called directly when the terms of use are accepted. Before this change, the cache would contain a stale empty profile.

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -236,6 +236,7 @@ public void onFailure(Throwable t) {
         public void onFailure(Throwable t) {
           currentUserAccessToken = null;
           currentUserProfile = null;
+          clearQueryClientCache();
```
nickgros, line 239: Regardless of success or failure, when we set the accessToken, clear the cache. This fixes the issue because this method is called directly when the terms of use are accepted. Before this change, the cache would contain a stale empty profile.

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -103,7 +103,6 @@ public void loginUser(
           public void onSuccess(LoginResponse response) {
             storeAuthenticationReceipt(response.getAuthenticationReceipt());
             setNewAccessToken(response.getAccessToken(), callback);
-            clearQueryClientCache();
```
nickgros, line 106: `setNewAccessToken` calls `initializeFromExistingAccessTokenCookie`, which will clear the cache, making this call redundant.

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -388,7 +389,6 @@ public void onSuccess(UserProfile result) {
               ginInjector.getGlobalApplicationState().refreshPage();
             }
             checkForQuarantinedEmail();
-            clearQueryClientCache();
```
nickgros, line 391: This callback is within a call to `initializeFromExistingAccessTokenCookie`, which unconditionally clears the cache, so this call is redundant.

src/test/java/org/sagebionetworks/web/client/security/AuthenticationControllerImplTest.java
```diff
@@ -283,6 +283,7 @@ public void testLoginUserNotAcceptedTermsOfUse() {
 
     assertNull(authenticationController.getCurrentUserProfile());
     verify(mockSessionDetector).initializeAccessTokenState();
+    verify(mockQueryClient).clear();
```
nickgros, line 286: This test represents the specific scenario captured in the Jira issue, so this specific verification shows that the bug is fixed.

jay-hodgson (approved):

PR #5050, opened 2023-01-26
SWC-6364: Fix case where default columns were not fetched for dataset collections

nickgros:
- Handle fetching default columns for Datasets, Dataset Collections

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/ViewDefaultColumns.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/ViewDefaultColumns.java
@@ -99,6 +99,11 @@ public class ViewDefaultColumns {
   }
 
   public List<ColumnModel> getDefaultViewColumns(TableType tableType) {
+    /*
+     * TODO: We should instead asynchronously fetch the default columns using the viewEntityType and viewTypeMask of the specific table.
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/ViewDefaultColumns.java
```diff
@@ -46,12 +46,20 @@ public void init() {
     FluentFuture<List<ColumnModel>> submissionViewColumnsFuture = jsClient.getDefaultColumnsForView(
       ViewEntityType.submissionview
     );
+    FluentFuture<List<ColumnModel>> datasetColumnsFuture = jsClient.getDefaultColumnsForView(
+      ViewEntityType.dataset
+    );
+    FluentFuture<List<ColumnModel>> datasetCollectionColumnsFuture = jsClient.getDefaultColumnsForView(
+      ViewEntityType.datasetcollection
+    );
```
nickgros, line 54: We only fetch default columns for certain view types, and treat this class as a sort of cache for these default column sets.

This solution might not be perfect for views with unusual masks, and could the same issue could recur if we added a new view type. The "proper" solution would probably require making this fetch asynchronous based on the current table. Then we would need to add loading + failure cases for the places where we need this data, and do some refactoring to prevent fetching the default columns multiple times--ultimately, I decided to stick with the current pattern instead.

jay-hodgson (approved):

PR #5051, opened 2023-01-26
SWC-6381

jay-hodgson:
In the header (so on every page).  Dismissible, which stores a value in local storage (which persists on logout).

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/cookie/CookieKeys.java
+++ b/src/main/java/org/sagebionetworks/web/client/cookie/CookieKeys.java
@@ -29,6 +29,9 @@ public class CookieKeys {
     "org.sagebionetworks.synapse.datetime.utc";
   public static String COOKIES_ACCEPTED =
     "org.sagebionetworks.security.cookies.notification.okclicked";
+  public static String NIH_SHARING_ALERT_DISMISSED =
+    "org.sagebionetworks.nih.sharing.notification.clicked";
+
   public static String PORTAL_CONFIG =
     "org.sagebionetworks.security.cookies.portal.config";
 }
--- a/src/main/java/org/sagebionetworks/web/client/cookie/CookieKeys.java
+++ b/src/main/java/org/sagebionetworks/web/client/cookie/CookieKeys.java
@@ -8,9 +8,6 @@ package org.sagebionetworks.web.client.cookie;
  */
 public class CookieKeys {
 
-  public static String USER_LOGGED_IN_RECENTLY =
-    "org.sagebionetworks.security.user.login.recently";
-
   /**
    * Login token
    */
@@ -27,10 +24,6 @@ public class CookieKeys {
    */
   public static String SHOW_DATETIME_IN_UTC =
     "org.sagebionetworks.synapse.datetime.utc";
-  public static String COOKIES_ACCEPTED =
-    "org.sagebionetworks.security.cookies.notification.okclicked";
-  public static String NIH_SHARING_ALERT_DISMISSED =
-    "org.sagebionetworks.nih.sharing.notification.clicked";
 
   public static String PORTAL_CONFIG =
     "org.sagebionetworks.security.cookies.portal.config";
--- a/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
@@ -9,6 +9,8 @@ import com.google.gwt.place.shared.Place;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.rpc.StatusCodeException;
 import com.google.inject.Inject;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 import org.sagebionetworks.repo.model.UserProfile;
 import org.sagebionetworks.repo.model.auth.LoginRequest;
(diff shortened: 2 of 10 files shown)


src/main/java/org/sagebionetworks/web/client/cookie/CookieKeys.java
```diff
@@ -8,9 +8,6 @@
  */
 public class CookieKeys {
 
-  public static String USER_LOGGED_IN_RECENTLY =
```
jay-hodgson, line 11: Deleted (not used after setting)

src/main/java/org/sagebionetworks/web/client/cookie/CookieKeys.java
```diff
@@ -27,8 +24,7 @@ public class CookieKeys {
    */
   public static String SHOW_DATETIME_IN_UTC =
     "org.sagebionetworks.synapse.datetime.utc";
-  public static String COOKIES_ACCEPTED =
```
jay-hodgson, line 30: Moved to local storage

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -192,11 +204,6 @@ public void initializeFromExistingAccessTokenCookie(
       new FutureCallback<String>() {
         @Override
         public void onSuccess(String accessToken) {
-          cookies.setCookie(
-            CookieKeys.USER_LOGGED_IN_RECENTLY,
```
jay-hodgson, line 196: Not referenced

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -274,13 +281,29 @@ public static boolean isQuarantined(EmailQuarantineStatus status) {
     );
   }
 
+  public void clearLocalStorage() {
+    Map<String, String> storedKeyValues = new HashMap<String, String>();
```
jay-hodgson, line 284: Store local storage keys that should persist in this map, clear local storage, and then repopulate

src/main/java/org/sagebionetworks/web/client/widget/FullWidthAlert.java
```diff
@@ -156,6 +156,11 @@ public void setAlertType(AlertType type) {
     rerender();
   }
 
+  public void setOnClose(FullWidthAlertProps.Callback onClose) {
+    this.onClose = onClose;
+    rerender();
```
jay-hodgson, line 161: provide ability to set a custom onClose callback, which will result in setting a local storage key for this case

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -274,13 +280,29 @@ public static boolean isQuarantined(EmailQuarantineStatus status) {
...
+    storedKeyValues.clear();
```
nickgros: New object, doesn't need to be cleared.
```suggestion
```

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -50,6 +51,16 @@ public class AuthenticationControllerImpl implements AuthenticationController {
...
+  public static final String NIH_NOTIFICATION_DISMISSED =
+    "nih_notification_dismissed";
+  public static String COOKIES_ACCEPTED =
+    "org.sagebionetworks.security.cookies.notification.okclicked";
+
+  private String[] persistentLocalStorageKeys = new String[] {
+    USER_AUTHENTICATION_RECEIPT,
+    NIH_NOTIFICATION_DISMISSED,
+    COOKIES_ACCEPTED,
+  };
```
nickgros, line 63: There may be other keys referenced in SRC only that we want to persist

jay-hodgson: That's true.  Today, SRC values in local storage are cleared on logout.

src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
```diff
@@ -27,21 +30,33 @@ public Header(
...
     }
+    if (
+      !localStorage.contains(
+        AuthenticationControllerImpl.NIH_NOTIFICATION_DISMISSED
+      )
+    ) {
+      view.setNIHAlertVisible(true);
+    } else {
+      view.setNIHAlertVisible(false);
+    }
```
nickgros: Opinionated code style, take it or leave it
```suggestion
    view.setCookieNotificationVisible(
      !localStorage.contains(AuthenticationControllerImpl.COOKIES_ACCEPTED)
    )
    view.setNIHAlertVisible(
      !localStorage.contains(AuthenticationControllerImpl.NIH_NOTIFICATION_DISMISSED)
    )
```

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
jay-hodgson: This can be extended to add those SRC keys that we want to protect

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/widget/header/Header.java
jay-hodgson: Copied style from my more verbose days as a java developer :)

PR #5059, opened 2023-02-07
SWC-6383: Restrict subdomains that can use credentials via CORS

nickgros:
- Disallow unknown subdomains from using Access-Control-Allow-Credentials by only specifying the Access-Control-Allow-Origin for known subdomains, instead of any *.synapse.org host.

The list was much larger, I annotated why I removed them in this commit--are there any that should stay in the list?: [`44197ac` (#5059)](https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5059/commits/44197ac72e0893c01ffbbcbcee3b393b2d6be7b3)


src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -29,93 +29,23 @@ public class CORSFilter extends OncePerRequestFilter {
     "www",
     "staging",
```
nickgros, line 30: Still allow staging.synapse.org to access the session cookie

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -29,93 +29,23 @@ public class CORSFilter extends OncePerRequestFilter {
...
     "tst",
-    "synapse-staging",
-    "status",
-    "staging-signin",
```
nickgros: In general I removed the staging/prod subdomains for all portals and portal-like apps, to keep the list maintainable. We can add back some/all of them if desired.

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -29,93 +29,23 @@ public class CORSFilter extends OncePerRequestFilter {
...
     "signin",
-    "prod",
```
nickgros: Many of these also redirect, e.g. prod.synapse.org redirects to www.synapse.org

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -29,93 +29,23 @@ public class CORSFilter extends OncePerRequestFilter {
...
-    "prod.signin",
-    "nbconvert",
-    "docker",
-    "cdn-www",
-    "cdn-staging",
-    "cdn-tst",
```
nickgros: Also excluding CDN subdomains

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -29,93 +29,23 @@ public class CORSFilter extends OncePerRequestFilter {
...
     "htan",
-    "prod.htan",
-    "staging.htan",
-    "minidream",
     "nf",
-    "help.nf",
-    "staging-tools.nf",
-    "staging.nf",
-    "static.nf",
-    "tools.nf",
```
nickgros: redirects to `nf.synapse.org/...`

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -29,93 +29,23 @@ public class CORSFilter extends OncePerRequestFilter {
...
-    "prod.arkportal",
-    "staging.arkportal",
     "bsmn",
-    "prod.bsmn",
-    "staging.bsmn",
     "cancercomplexity",
-    "help.cancercomplexity",
-    "prod.cancercomplexity",
-    "staging.cancercomplexity",
     "www.cancercomplexity",
```
nickgros, line 39: This was the only portal  I saw with an additional subdomain site that 
1. Uses Synapse auth and
2. Doesn't seem to redirect -- keeping it here in case this site is accessed often

jay-hodgson (approved):

PR #5060, opened 2023-02-08
SWC-6389: Automatically configure repo endpoint using incoming request host

nickgros:
- Server-side requests to Synapse backend services are now forwarded to the appropriate stack based on the incoming request URL host. For example, requests to www.synapse.org will be passed to repo-prod.prod.sagebase.org. Requests to staging.synapse.org will be passed to repo-staging.prod.synapse.org.
- Moved StackEndpoints.java to the `server` package, where it was exclusively used.
- Cleanup unused or duplicate variables


src/test/java/org/sagebionetworks/web/unitserver/StackEndpointsTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+import org.sagebionetworks.web.server.StackEndpoints;
+
+public class StackEndpointsTest {
+
+  @Before
+  public void setup() {
+    clearProperties();
```
jay-hodgson, line 26: Currently reverts the change before the test is executed, which is great.  But shouldn't this be after the test is executed so it does not have the potential to pollute any downstream tests?

jay-hodgson (approved):
Approved (on the condition of a successful test suite execution)

PR #5065, opened 2023-02-20
SWC-6398 - fix cache issue leading to unusual bugs and excessive refetching

nickgros:
Targeting release-442, which is currently staging

- Use `resetQueries` instead of `clearQueries`. `clearQueries` removes subscribers, which is what I suspect caused the unusual state bug in SWC-6398
- don't reset the query cache when the access token doesn't change
- change `AuthenticationControllerImpl.currentUserAccessToken` and `AuthenticationControllerImpl.currentUserProfile` to non-static, which was causing issues in tests. AuthenticationControllerImpl is a singleton, so this should not cause new issues.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryClient.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryClient.java
@@ -9,6 +9,11 @@ public class QueryClient {
   public QueryClient(QueryClientOptions config) {}
 
   /**
+   * Removes all cached query data and triggers a refetch.
+   */
+  public native void resetQueries();
+
+  /**
    * Removes the cached query data and triggers a refetch.
    * @param queryKey
    */
@@ -19,6 +24,4 @@ public class QueryClient {
    * @param queryKey
    */
   public native void invalidateQueries(List<?> queryKey);
-
-  public native void clear();
 }
--- a/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
@@ -61,8 +61,8 @@ public class AuthenticationControllerImpl implements AuthenticationController {
     NIH_NOTIFICATION_DISMISSED,
     COOKIES_ACCEPTED,
   };
-  private static String currentUserAccessToken;
-  private static UserProfile currentUserProfile;
+  private String currentUserAccessToken;
+  private UserProfile currentUserProfile;
   private UserAccountServiceAsync userAccountService;
   private ClientCache localStorage;
   private SessionStorage sessionStorage;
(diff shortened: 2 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -61,8 +61,8 @@ public class AuthenticationControllerImpl implements AuthenticationController {
     NIH_NOTIFICATION_DISMISSED,
     COOKIES_ACCEPTED,
   };
-  private static String currentUserAccessToken;
-  private static UserProfile currentUserProfile;
```
nickgros, line 65: Am I correct in assuming that we won't have issues removing `static` because this is a singleton class? We were never accessing these in a static context, so no additional changes were required by this change.

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryClient.java
```diff
@@ -19,6 +24,4 @@ public QueryClient(QueryClientOptions config) {}
    * @param queryKey
    */
   public native void invalidateQueries(List<?> queryKey);
-
-  public native void clear();
```
nickgros, line 23: `clear` removes listeners, which leads to all sorts of bugs where end up in an invalid state, so we pretty much never want to use this.

`resetQueries` removes existing cache data and immediately updates listeners, so you end up seeing things like loaders, which is appropriate when your login state changed.

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryClient.java
```diff
@@ -8,6 +8,11 @@ public class QueryClient {
 
   public QueryClient(QueryClientOptions config) {}
 
+  /**
+   * Removes all cached query data and triggers a refetch.
+   */
+  public native void resetQueries();
```
nickgros, line 14: add no-arg overload, which will reset all queries.

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -202,8 +202,10 @@ public void initializeFromExistingAccessTokenCookie(
       new FutureCallback<String>() {
         @Override
         public void onSuccess(String accessToken) {
-          currentUserAccessToken = accessToken;
-          clearQueryClientCache();
+          if (!Objects.equals(currentUserAccessToken, accessToken)) {
+            resetQueryClientCache();
+            currentUserAccessToken = accessToken;
+          }
```
nickgros, line 208: This runs on a timer, so every minute we were clearing the cache!

jay-hodgson: I understand that the check should be executed every minute, but the access token should not be changing.  Right?

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
nickgros: Correct. Now that we have this condition, the cache will only be reset if the token has changed.

nickgros:
Opted for #5066

PR #5067, opened 2023-02-20
SWC-6399

jay-hodgson:


src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -62,6 +64,10 @@ protected void doFilterInternal(
       String subdomain = url
         .getHost()
         .substring(0, url.getHost().length() - SYNAPSE_ORG_SUFFIX.length());
+      if (subdomain.startsWith(STAGING_PREFIX)) {
+        subdomain =
+          subdomain.substring(STAGING_PREFIX.length(), subdomain.length());
+      }
```
nickgros: Technically this would re-expose us to [NCC-2022-004](https://sagebionetworks.jira.com/browse/SWC-6374) on sites that don't exist e.g. staging.tst.synapse.org

nickgros (approved):

PR #5080, opened 2023-03-06
IT-2094: Add prod.adknowledgeportal to CORS

xschildw:
This PR adds the internal prod.adknowledgeportal endpoint to the CORS list. To be used for testing the deploy before pointing adknowledgeportal.synapse.org to it.


xschildw:
Not needed, already updated the deployment.

PR #5081, opened 2023-03-10
SWC-6393 - Add 2FA Settings

nickgros:
- Add 2FA section to summary page (experimental mode only)
- Add TwoFactorAuth place to handle enrollment + backup code generation
- Tweak styles to match 2FA settings pane

## Demo

https://user-images.githubusercontent.com/17580037/224388484-3a549d49-fa1b-4651-88d7-ff4506f0eb3a.mov

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
@@ -1,11 +1,14 @@
 package org.sagebionetworks.web.client.presenter;
 
+import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static org.sagebionetworks.web.client.ServiceEntryPointUtils.fixServiceEntryPoint;
 import static org.sagebionetworks.web.client.presenter.ProfilePresenter.IS_CERTIFIED;
 import static org.sagebionetworks.web.client.presenter.ProfilePresenter.ORC_ID;
 import static org.sagebionetworks.web.client.presenter.ProfilePresenter.PROFILE;
 import static org.sagebionetworks.web.client.presenter.ProfilePresenter.VERIFICATION_SUBMISSION;
 
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.gwt.core.shared.GWT;
 import com.google.gwt.place.shared.Place;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.Widget;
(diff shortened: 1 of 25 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
```diff
@@ -255,6 +260,32 @@ public void configure() {
     view.setOauthClientSettingsVisible(
       DisplayUtils.isInTestWebsite(ginInjector.getCookieProvider())
     );
+
+    if (DisplayUtils.isInTestWebsite(ginInjector.getCookieProvider())) {
+      view.setTwoFactorAuthSettingsVisible(true);
+    } else {
+      // If 2FA is enabled, show the settings pane, even if experimental mode is disabled
```
nickgros, line 267: Small block to guard against the case where someone adds 2FA but then can't log into apps that haven't been updated.

src/main/java/org/sagebionetworks/web/client/presenter/SettingsPresenter.java
```diff
@@ -255,6 +260,32 @@ public void configure() {
...
+            @Override
+            public void onSuccess(TwoFactorAuthStatus result) {
+              view.setTwoFactorAuthSettingsVisible(
+                TwoFactorState.ENABLED.equals(result.getStatus())
+              );
+            }
+
+            @Override
+            public void onFailure(Throwable t) {
+              GWT.log("Unable to get 2FA status", t);
```
nickgros, line 282: Not worried about the error handling because this is temporary code that will be removed when 2FA comes out of experimental mode

src/main/webapp/sass/_core.scss
```diff
@@ -531,10 +531,9 @@ end sticky footer
...
 
 .light-border {
-  border: solid #ccc;
-  border-width: 0.1em;
-  border-radius: 0.4em;
-  padding: 2px 4px 2px 3px;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+  padding: 16px;
 }
```
nickgros, line 537: .light-border and .highlight-box were both used on the settings page, so change the styles with two goals
- light-border and highlight-box should match visually
- Change both styles so contents (e.g. title font) match the contents of the new 2FA panel. There is a risk that these could drift, but my hope is SageAccountWeb is production-ready before drift happens. IMO this is worth it compared to the complexity of a custom style prop or adding complex CSS to override the 2FA panel styles

jay-hodgson:
Surprised there isn't a confirmation dialog for deactivating 2fa

jay-hodgson (approved):

PR #5083, opened 2023-03-23
SWC-6395 - show actions required for individual file download in tooltip

nickgros:
- Show actions required for download on individual entity page

image

We will also need to update SRC to include https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/91

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -47,7 +47,11 @@ import org.sagebionetworks.repo.model.Versionable;
 import org.sagebionetworks.repo.model.VersionableEntity;
 import org.sagebionetworks.repo.model.auth.UserEntityPermissions;
 import org.sagebionetworks.repo.model.docker.DockerRepository;
+import org.sagebionetworks.repo.model.download.ActionRequiredList;
 import org.sagebionetworks.repo.model.download.AddBatchOfFilesToDownloadListResponse;
+import org.sagebionetworks.repo.model.download.EnableTwoFa;
+import org.sagebionetworks.repo.model.download.MeetAccessRequirement;
+import org.sagebionetworks.repo.model.download.RequestDownload;
 import org.sagebionetworks.repo.model.entitybundle.v2.EntityBundle;
 import org.sagebionetworks.repo.model.file.FileHandle;
 import org.sagebionetworks.repo.model.table.Dataset;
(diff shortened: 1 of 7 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -694,18 +700,66 @@ private void configureFullTextSearch() {
...
+              new FutureCallback<ActionRequiredList>() {
+                @Override
+                public void onSuccess(@NullableDecl ActionRequiredList result) {
+                  if (result != null) {
+                    StringBuilder downloadMenuTooltipText = new StringBuilder(
+                      "You don't have permission to download this file."
+                    );
+                    // There may be multiple actions of the same class, but we only want to show one message for each type
+                    // Get the unique set of action classes.
+                    Set<Class<? extends org.sagebionetworks.repo.model.download.Action>> uniqueClasses = result
```
nickgros, line 740: Unfortunately there's another `Action` class in use here, so we have to use the full classname.

This approach seems unusual at first, but consider the case where a file has multiple Access Requirements. The service will return one action per unmet AR, but we only want to show one generic message for all unmet ARs. Getting the set of unique Action classes handles that case.

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -694,18 +700,66 @@ private void configureFullTextSearch() {
   }
 
   private FluentFuture configureFileDownload() {
-    FluentFuture<RestrictionInformationResponse> future = getDoneFuture(null);
+    FluentFuture<RestrictionInformationResponse> restrictionInformationFuture = getDoneFuture(
+      null
+    );
```
nickgros, line 713: Just renaming this variable since we are now using multiple futures in this method

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2797,4 +2852,17 @@ public Widget asWidget() {
   public void setIsShowingVersion(boolean isShowingVersion) {
     this.isShowingVersion = isShowingVersion;
   }
+
+  private String getTooltipTextForRequiredActionForDownload(
+    Class<? extends org.sagebionetworks.repo.model.download.Action> clazz
+  ) {
+    if (RequestDownload.class.equals(clazz)) {
+      return "Request access from an administrator, shown under File Tools ➔ File Sharing Settings.";
```
jay-hodgson: These could be refactored to be constants (referenced by the tests)

jay-hodgson (approved):

PR #5087, opened 2023-04-10
SWC-6415 - Fix many broken docs links

nickgros:
- Base branch is release-449, which is currently staging

### DIFF
--- a/src/main/resources/org/sagebionetworks/web/client/widget/accessrequirements/ACTAccessRequirementWidgetViewImpl.ui.xml
+++ b/src/main/resources/org/sagebionetworks/web/client/widget/accessrequirements/ACTAccessRequirementWidgetViewImpl.ui.xml
@@ -46,12 +46,12 @@
       </b:BlockQuote>
       <bh:Div>
         <bh:Text>
-          For more information on use conditions, please read the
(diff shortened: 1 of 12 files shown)


hallieswan (approved):
Looks good! 

Not sure if it would make sense in this project, but have we considered managing the URLs in a constants utility, so they only have to be updated in one location, similar to [SynapseConstants](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/lib/utils/SynapseConstants.ts)? Up to you!

nickgros:
> Not sure if it would make sense in this project, but have we considered managing the URLs in a constants utility, so they only have to be updated in one location, similar to [SynapseConstants](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/lib/utils/SynapseConstants.ts?rgh-link-date=2023-04-10T15%3A45%3A43Z)? Up to you!

That would be ideal, I'm not sure if it's possible to use those in the .ui.xml files. Looks like it might be possible, I'll look into it: https://stackoverflow.com/questions/23782547/constant-values-supplied-from-properties-files-in-uibinder-gwt

PR #5090, opened 2023-04-21
SWC-6426: update SRC and host plotly js

jay-hodgson:


src/main/webapp/Portal.html
```diff
@@ -495,6 +495,7 @@
       head.load(cdnEndpoint + 'js/polyfill-2.js')
       head.load(cdnEndpoint + 'js/details-shim.min.js')
       head.load(cdnEndpoint + 'js/xss.min.js')
+      head.load(cdnEndpoint + 'js/plotly-1.54.5.min.js')
```
nickgros: Is it trivial to add plotly.js@1.54.5 (or better yet ^1.58.4) to package.json instead and copy the script from node_modules? That would give us dependabot alerts + minor version bumps, even though I don't think we would see the latter in Plotly 1.x.

src/main/webapp/js/plotly-1.54.5.min.js
nickgros, line 1: Delete this and we're good to go!

nickgros (approved):

src/main/webapp/js/plotly-1.54.5.min.js
jay-hodgson: Done!

PR #5101, opened 2023-05-23
SWC-6459: Add a (bootstrap) tooltip when the entity name is over 40 c…

jay-hodgson:
…haracters

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadgeViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadgeViewImpl.java
@@ -192,7 +192,13 @@ public class EntityBadgeViewImpl extends Composite implements EntityBadgeView {
     entityAnchor.setPaddingBottom(0);
     entityTypeIconContainer.setWidget(entityTypeIcon);
     entityTypeIcon.setType(getEntityTypeForClassName(entityHeader.getType()));
-    entityContainer.add(entityAnchor);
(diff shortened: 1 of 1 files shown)


jay-hodgson:
Here are a couple of examples:
Screen Shot 2023-05-23 at 4 47 19 PM
Screen Shot 2023-05-23 at 4 47 39 PM

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadgeViewImpl.java
```diff
@@ -192,7 +192,13 @@ public void setEntity(final EntityHeader entityHeader) {
     entityAnchor.setPaddingBottom(0);
     entityTypeIconContainer.setWidget(entityTypeIcon);
     entityTypeIcon.setType(getEntityTypeForClassName(entityHeader.getType()));
-    entityContainer.add(entityAnchor);
+    if (entityHeader.getName().length() > 40) {
```
nickgros, line 195: Does this work as expected at a narrow screen size? Wonder if there are downsides to just giving all entities a tooltip.

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityBadgeViewImpl.java
jay-hodgson: I think so based on testing, but this is not a good solution for browsing on mobile

PR #5102, opened 2023-05-24
SWC-6451

jay-hodgson:
- Refactor CrawlFilter to reuse dataset json-ld content
- Expose dataset json-ld content via a new servlet call
- Inject script element into HEAD element if entity is a Dataset
- Remove element on detach (or when it is reconfigured)


src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
nickgros, line 1: Appreciate the cleanup here

nickgros (approved):
🤞that the crawler handles this!

PR #5103, opened 2023-05-25
SWC-6404: Following Page 

nickgros:
Requires a new version of SRC
- Add FollowingPlace which uses new SRC SubscriptionPage
- Remove SubscriptionListWidget and replace on SettingsPage with info about where that was moved.
- Bump node


src/main/resources/org/sagebionetworks/web/client/view/SettingsViewImpl.ui.xml
```diff
@@ -309,13 +309,18 @@
...
+          Your followed forums and threads have moved! To view the topics that
+          you are following, click
+          <a
+            onclick="document.querySelector('[aria-label=\'Your Account\']').click()"
+          >
+            Your Account
+          </a>
+          in the main sidebar menu and
+          then click "Following".
+        </bh:Paragraph>
```
nickgros, line 323: Direct users to the new way to find this info on the old settings page. The `onclick` opens the nav drawer to where the user will see `Following`. It's not implemented in a very reliable way (the label could change), but it's acceptable risk for temporary code IMO

src/main/java/org/sagebionetworks/web/client/presenter/FollowingPagePresenter.java
```diff
@@ -0,0 +1,38 @@
...
+public class FollowingPagePresenter
+  extends AbstractActivity
+  implements Presenter<FollowingPlace> {
+
+  private final FollowingPageView view;
+
+  @Inject
+  public FollowingPagePresenter(FollowingPageView view, SynapseAlert synAlert) {
+    this.view = view;
+    view.setSynAlert(synAlert);
```
jay-hodgson: Can this SynapseAlert be removed (looks like it is not used)?

src/main/java/org/sagebionetworks/web/client/view/FollowingPageViewImpl.java
```diff
@@ -0,0 +1,88 @@
...
+  SynapseReactClientFullContextPropsProvider propsProvider;
+
+  public interface LoginViewImplBinder
+    extends UiBinder<Widget, FollowingPageViewImpl> {}
+
+  @Inject
+  public FollowingPageViewImpl(
+    LoginViewImplBinder uiBinder,
+    Header headerWidget,
+    SynapseJSNIUtils jsniUtils,
```
jay-hodgson: Can SynapseJSNIUtils reference be removed?

src/main/resources/org/sagebionetworks/web/client/view/SettingsViewImpl.ui.xml
jay-hodgson: This is great!

jay-hodgson (approved):

PR #5105, opened 2023-05-25
PORTALS-2660:  Pop up ACL editor on View Sharing Settings

jay-hodgson:
When a user encounters a 403 while logged in viewing a Table/View and is presented the list of Actions required to access


jay-hodgson:
Screen Shot 2023-05-25 at 2 17 23 PM

nickgros (approved):

PR #5106, opened 2023-05-31
PORTALS-2470: return local storage key value if expiration is missing

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/cache/ClientCacheImpl.java
```diff
@@ -18,14 +18,18 @@ public ClientCacheImpl(StorageWrapper storage) {
   @Override
   public String get(String key) {
     String expireTimeString = storage.getItem(key + SUFFIX);
-    if (storage.isStorageSupported() && expireTimeString != null) {
```
jay-hodgson, line 21: Instead of returning null if the expireTimeString is null, return the value from local storage

hallieswan (approved):
Looks good to me!

PR #5107, opened 2023-06-02
PORTALS-2470: add OrientationBanners

hallieswan:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/267

https://github.com/Sage-Bionetworks/SynapseWebClient/assets/26949006/80b76628-630c-4266-8ef6-0e173cb39077

https://github.com/Sage-Bionetworks/SynapseWebClient/assets/26949006/4ee170bd-0b71-4d11-8612-4e65574019e5

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
+++ b/src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
@@ -94,6 +94,8 @@ public interface SynapseJSNIUtils {
 
   public void loadCss(String url);
 
+  public String[] getSrcPersistentLocalStorageKeys();
+
   /**
    * initialize the behavior for on pop state
    */
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
@@ -1,5 +1,6 @@
 package org.sagebionetworks.web.client.jsinterop;
 
+import com.google.gwt.core.client.JsArrayString;
 import jsinterop.annotations.JsNullable;
 import jsinterop.annotations.JsPackage;
 import jsinterop.annotations.JsType;
@@ -23,6 +24,7 @@ public class SRC {
     public static ReactComponentType<DownloadCartPageProps> DownloadCartPage;
     public static ReactComponentType<DownloadConfirmationProps> DownloadConfirmation;
     public static ReactComponentType<FullWidthAlertProps> FullWidthAlert;
+    public static ReactComponentType<OrientationBannerProps> OrientationBanner;
     public static ReactComponentType<SchemaDrivenAnnotationEditorProps> SchemaDrivenAnnotationEditor;
     public static ReactComponentType<SynapseNavDrawerProps> SynapseNavDrawer;
     public static ReactComponentType<EmptyProps> FavoritesPage;
@@ -79,5 +81,11 @@ public class SRC {
     /* We use FullContextProvider because it will provide the SynapseContext, react-query QueryContext, and MUI Theme
      context for all React trees that we render */
     public static ReactComponentType<SynapseReactClientFullContextProviderProps> FullContextProvider;
+  }
+
+  @JsType(isNative = true)
+  public static class SynapseConstants {
+
+    public static JsArrayString PERSISTENT_LOCAL_STORAGE_KEYS;
   }
 }
--- a/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
@@ -9,7 +9,9 @@ import com.google.gwt.place.shared.Place;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.rpc.StatusCodeException;
 import com.google.inject.Inject;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import org.sagebionetworks.repo.model.UserProfile;
@@ -56,7 +58,7 @@ public class AuthenticationControllerImpl implements AuthenticationController {
   public static String COOKIES_ACCEPTED =
     "org.sagebionetworks.security.cookies.notification.okclicked";
 
-  private String[] persistentLocalStorageKeys;
+  private List<String> persistentLocalStorageKeys;
   private String currentUserAccessToken;
   private UserProfile currentUserProfile;
   private UserAccountServiceAsync userAccountService;
@@ -429,18 +431,14 @@ public class AuthenticationControllerImpl implements AuthenticationController {
     };
     String[] srcPersistentLocalStorageKeys = jsniUtils.getSrcPersistentLocalStorageKeys();
 
-    int srcKeysLength = srcPersistentLocalStorageKeys.length;
-    int swcKeysLength = swcPersistentLocalStorageKeys.length;
+    this.persistentLocalStorageKeys = new ArrayList<String>();
 
-    this.persistentLocalStorageKeys = new String[srcKeysLength + swcKeysLength];
-
-    for (int i = 0; i < srcKeysLength; i++) {
-      this.persistentLocalStorageKeys[i] = srcPersistentLocalStorageKeys[i];
+    for (int i = 0; i < srcPersistentLocalStorageKeys.length; i++) {
+      this.persistentLocalStorageKeys.add(srcPersistentLocalStorageKeys[i]);
     }
 
-    for (int i = 0; i < swcKeysLength; i++) {
-      this.persistentLocalStorageKeys[srcKeysLength + i] =
-        swcPersistentLocalStorageKeys[i];
+    for (int i = 0; i < swcPersistentLocalStorageKeys.length; i++) {
+      this.persistentLocalStorageKeys.add(swcPersistentLocalStorageKeys[i]);
     }
   }
 }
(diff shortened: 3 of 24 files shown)


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
```diff
@@ -94,6 +94,8 @@ public void getFilePartMd5(
 
   public void loadCss(String url);
 
+  public String[] getSrcPersistentLocalStorageKeys();
```
hallieswan, line 97: Add a method to convert SRC's persistent local storage keys from a JS array to a Java array in this interface, so that this method can be mocked in AuthenticationControllerImplTest. Happy to adjust if there's a better place or way to handle this case!

src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
```diff
@@ -80,4 +82,10 @@ public static class SynapseContext {
      context for all React trees that we render */
     public static ReactComponentType<SynapseReactClientFullContextProviderProps> FullContextProvider;
   }
+
+  @JsType(isNative = true)
+  public static class SynapseConstants {
+
+    public static JsArrayString PERSISTENT_LOCAL_STORAGE_KEYS;
+  }
```
hallieswan, line 90: Make the persistent local storage keys from SRC available

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -423,4 +420,27 @@ public void onSuccess(UserProfile result) {
...
     );
   }
+
+  private void setPersistentLocalStorageKeys() {
+    String[] swcPersistentLocalStorageKeys = new String[] {
+      USER_AUTHENTICATION_RECEIPT,
+      NIH_NOTIFICATION_DISMISSED,
+      COOKIES_ACCEPTED,
+    };
+    String[] srcPersistentLocalStorageKeys = jsniUtils.getSrcPersistentLocalStorageKeys();
```
hallieswan, line 432: Add keys from both SWC and SRC to the list of persistent local storage keys

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -93,6 +94,9 @@ public class ProfileViewImpl extends Composite implements ProfileView {
   @UiField
   Heading pageHeaderTitle;
 
+  @UiField
+  SimplePanel orientationBannerPanel;
```
hallieswan, line 98: Use a SimplePanel, so that each ProfileView can set its own OrientationBanner

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -516,22 +545,59 @@ public void setTabSelected(Synapse.ProfileArea targetTab) {
...
       pageHeaderTitle.setText("Your Favorites");
-
+      configureOrientationBanner(
+        "Favorites",
+        "Getting Started With Your Favorites",
+        "Throughout Synapse, you can favorite pretty much any item (project, file, folder, table, dataset, etc.) by clicking the star icon next to its name. This will add that item to your favorites list.",
+        null,
+        null,
+        "Learn More About Favorites",
+        "https://help.synapse.org/docs/Navigating-Synapse.2048557182.html#NavigatingSynapse-Favorites"
```
hallieswan, line 604: I found a docs page related to Favorites, so included it here. However, [the design](https://www.figma.com/file/bALXz8AOxdzQyrsSd4RkVm/Governance-UI%3A-Registration-and-Data-Access?type=design&node-id=2557-12762&t=XtpKoMaYnPxvytt6-4) doesn't include a docs link for Favorites, so happy to remove the link if that's preferred!

src/main/java/org/sagebionetworks/web/client/widget/OrientationBanner.java
```diff
@@ -0,0 +1,136 @@
...
+    String secondaryButtonHref
+  ) {
+    this.name = name;
+    this.title = title;
+    this.text = text;
+    this.primaryButtonText = primaryButtonText;
+    this.secondaryButtonText = secondaryButtonText;
+    this.secondaryButtonHref = secondaryButtonHref;
+    setPrimaryCTAHref(primaryButtonHref);
+  }
```
hallieswan: Method to re-configure OrientationBanners in TabViews and ProfileViews without re-rendering the banner after each prop is updated

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/TabViewImpl.java
```diff
@@ -32,6 +34,9 @@ public class TabViewImpl implements TabView {
   @UiField
   ListItem tabListItem;
 
+  @UiField
+  SimplePanel orientationBannerPanel;
```
hallieswan, line 38: Use a SimplePanel, so that each Tab can set its own OrientationBanner

src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
nickgros: It doesn't 100% make sense here since you're using JSInterop, not JSNI. But I think it's fine here compared to other alternatives because it's a set of similar utilities. I'm ok with it.

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -56,11 +56,7 @@ public class AuthenticationControllerImpl implements AuthenticationController {
   public static String COOKIES_ACCEPTED =
     "org.sagebionetworks.security.cookies.notification.okclicked";
 
-  private String[] persistentLocalStorageKeys = new String[] {
-    USER_AUTHENTICATION_RECEIPT,
-    NIH_NOTIFICATION_DISMISSED,
-    COOKIES_ACCEPTED,
-  };
+  private String[] persistentLocalStorageKeys;
```
nickgros: Suggest making this a `List<String>`

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -423,4 +420,27 @@ public void onSuccess(UserProfile result) {
...
+
+    this.persistentLocalStorageKeys = new String[srcKeysLength + swcKeysLength];
+
+    for (int i = 0; i < srcKeysLength; i++) {
+      this.persistentLocalStorageKeys[i] = srcPersistentLocalStorageKeys[i];
+    }
+
+    for (int i = 0; i < swcKeysLength; i++) {
+      this.persistentLocalStorageKeys[srcKeysLength + i] =
+        swcPersistentLocalStorageKeys[i];
```
nickgros: If we used a List<String>, I wouldn't have to think about if `srcKeysLength + i` is the right index

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -516,22 +545,59 @@ public void setTabSelected(Synapse.ProfileArea targetTab) {
...
+      );
       DisplayUtils.show(projectsTabContainer);
     } else if (targetTab == Synapse.ProfileArea.TEAMS) {
       pageHeaderTitle.setText("Your Teams");
+      configureOrientationBanner(
+        "Teams",
+        "Getting Started With Your Teams",
+        "Teams allow you to easily manage groups of users to control access to projects, communicate with colleagues, and participate in challenges.",
+        "Search Teams",
+        "https://www.synapse.org/#!TeamSearch:",
```
nickgros: Link takes you to the wrong stack if you're not on prod--not a huge deal but would like to fix. Is it possible to use a callback here instead of an href and use `goToPlace`?

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
nickgros: Good call!

src/main/resources/org/sagebionetworks/web/client/view/FollowingPageViewImpl.ui.xml
```diff
@@ -16,6 +16,11 @@
           creates a thread.
         </bh:Paragraph>
       </bh:Div>
+      <w:OrientationBanner
+        name="Following"
+        title="All The Discussions You're Following"
+        text="When you click the &ldquo;follow&rdquo; button on any forum or thread in Synapse, a link to that discussion will show up here. From here, it's easy to return to discussions you care about, or stop following the ones you don't."
+      />
```
nickgros, line 23: I appreciate that you created a widget that handles simple XML config cases

src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -566,7 +566,12 @@ public void setTabSelected(Synapse.ProfileArea targetTab) {
         "Getting Started With Your Teams",
         "Teams allow you to easily manage groups of users to control access to projects, communicate with colleagues, and participate in challenges.",
         "Search Teams",
-        "https://www.synapse.org/#!TeamSearch:",
+        new ClickHandler() {
+          @Override
+          public void onClick(ClickEvent event) {
+            presenter.goTo(new TeamSearch(teamSearchTextBox.getValue()));
+          }
+        },
```
hallieswan, line 574: Use a callback so that the link redirects to the correct place if not on prod -- 

https://github.com/Sage-Bionetworks/SynapseWebClient/assets/26949006/bf5464fd-62d9-4cdf-8678-ff1d1796a999

nickgros (approved):

PR #5118, opened 2023-06-16
SWC-6377

nickgros:
Fix SWC-6377, where the NavBar was not visible after registering and signing the terms of use. Root cause was the QueryClient was not reset after signing the ToU.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
@@ -5,6 +5,7 @@ import static org.sagebionetworks.web.client.ServiceEntryPointUtils.fixServiceEn
 
 import com.google.common.util.concurrent.FluentFuture;
 import com.google.common.util.concurrent.FutureCallback;
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.place.shared.Place;
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.rpc.StatusCodeException;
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -204,6 +217,8 @@ public void onSuccess(String accessToken) {
           if (!Objects.equals(currentUserAccessToken, accessToken)) {
             resetQueryClientCache();
             currentUserAccessToken = accessToken;
+          } else if (forceResetQueryClientCache) {
+            resetQueryClientCache();
           }
```
nickgros, line 222: Before you signed the ToU, we couldn't get your user profile. When you signed the ToU, your access token didn't change, so we didn't reset the QueryClient. So the useQuery instance to fetch your profile stayed in the `error` state.

Now, explicity force-reset the cache after signing the ToU.

src/test/java/org/sagebionetworks/web/unitclient/presenter/LoginPresenterTest.java
```diff
@@ -108,10 +115,6 @@ public void setup() {
       .callSuccessWith(mockUserProfile)
       .when(mockAuthenticationController)
       .setNewAccessToken(anyString(), any(AsyncCallback.class));
-    AsyncMockStubber
-      .callSuccessWith(mockUserProfile)
-      .when(mockAuthenticationController)
-      .initializeFromExistingAccessTokenCookie(any(AsyncCallback.class));
```
nickgros, line 114: The `AsyncMockStubber` util doesn't like when a method has extra params other than callbacks, so I had to switch to `verify` + `ArgumentCaptor` below

jay-hodgson (approved):

PR #5125, opened 2023-06-23
SWC-6471: move dataset json ld element injection to entity page load

jay-hodgson:
From TableEntity page load, to inject sooner and possibly be in time for Google crawler to find it

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
@@ -87,6 +87,7 @@ import org.sagebionetworks.repo.model.search.SearchResults;
 import org.sagebionetworks.repo.model.search.query.SearchQuery;
 import org.sagebionetworks.repo.model.table.ColumnChange;
 import org.sagebionetworks.repo.model.table.ColumnModel;
+import org.sagebionetworks.repo.model.table.Dataset;
 import org.sagebionetworks.repo.model.table.FacetColumnRequest;
 import org.sagebionetworks.repo.model.table.TableSchemaChangeRequest;
 import org.sagebionetworks.repo.model.table.TableUpdateRequest;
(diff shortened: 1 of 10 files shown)


src/main/resources/org/sagebionetworks/web/client/view/SettingsViewImpl.ui.xml
```diff
@@ -172,7 +172,7 @@
               </bh:Div>
             </bh:Div>
             <w:HelpWidget
-              helpMarkdown="A Synapse Qualified Researcher is a Certified User with a validated user profile.&#10;Synapse Qualified Researchers are eligible to request access to Bridge data."
+              helpMarkdown="A Synapse user with a validated profile is eligible to request access to specific controlled access tier data, including mHealth data."
```
jay-hodgson, line 175: Also this tooltip text change for https://sagebionetworks.jira.com/browse/SWC-6473

src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -2114,16 +2115,21 @@ public ArrayList<String[]> parseCsv(String csvPreviewText, char delimiter)
...
-    bundleRequest.setIncludeAnnotations(true);
-    EntityBundle entityBundle = getEntityBundle(entityId, bundleRequest);
-    try {
-      return CrawlFilter.getDatasetScriptElementContent(
-        entityBundle,
-        plainTextWiki
-      );
+    try {
+      org.sagebionetworks.client.SynapseClient synapseClient = createSynapseClient();
+      Entity entity = synapseClient.getEntityById(entityId);
```
jay-hodgson, line 2120: Change here is to check to see if the entity ID points to a Dataset entity.  Only return element content if it is a Dataset.

xschildw (approved):

PR #5129, opened 2023-07-03
SWC-6490: Fix AR page update submission action (production patch)

nickgros:
- Update synapse-react-client to capture changes in https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/350
- Handle case where targetSubject is `null`
- Refresh state when dialog is closed (which may indicate that the submission was updated)

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
@@ -9,6 +9,7 @@ import com.google.gwt.user.client.ui.Widget;
 import com.google.inject.Inject;
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.Nullable;
 import org.gwtbootstrap3.client.ui.Alert;
 import org.gwtbootstrap3.client.ui.Button;
 import org.gwtbootstrap3.client.ui.html.Div;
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
```diff
@@ -369,10 +370,14 @@ public void showRequestAccessModal(
...
+        presenter.refreshApprovalState();
         hideRequestAccessModal();
       };
-      String entityId = "";
-      if (targetSubject.getType() == RestrictableObjectType.ENTITY) {
+      String entityId = null;
+      if (
+        targetSubject != null &&
+        targetSubject.getType() == RestrictableObjectType.ENTITY
+      ) {
```
nickgros, line 380: Handle null case

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
```diff
@@ -369,10 +370,14 @@ public void showRequestAccessModal(
       List<JSONObjectAdapter> arList = new ArrayList<JSONObjectAdapter>();
       arList.add(arAsJson);
       AccessRequirementListProps.Callback onHide = () -> {
+        presenter.refreshApprovalState();
```
nickgros, line 373: The modal may hide when the user has submitted their request. It's easy enough to refresh the approval state in this callback, rather than adding a specific `onSuccess` callback.

jay-hodgson (approved):

PR #5132, opened 2023-07-06
Upgrade dependencies

nickgros:
- SWC-6468
- SWC-6422
- SWC-6419
- SWC-6372
- SWC-6429
- SWC-6299

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
   "dependencies": {
-    "@rjsf/core": "^4.0.1",
-    "@sage-bionetworks/react-base-table": "^1.13.2",
+    "@sage-bionetworks/react-base-table": "1.13.4",
     "croppie": "2.6.5",
     "font-awesome": "4.7.0",
     "jquery": "3.5.1",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "jsplumb": "2.13.2",
     "moment": "^2.29.4",
     "pica": "6.0.0",
-    "plotly.js": "^1.58.4",
+    "plotly.js": "^2.24.3",
```
nickgros, line 10: We should be able to bump to v2 safely. Our integration only makes simple plots and the breaking changes in v2 were limited. I manually verified that with v2, we still properly render plots created in wikis.

pom.xml
```diff
@@ -60,15 +60,15 @@
...
+      <url>https://maven-gae-plugin.googlecode.com/svn/repository/</url>
     </repository>
     <repository>
       <id>maven-central-rep-1</id>
       <url>https://repo1.maven.org/maven2/</url>
     </repository>
     <repository>
       <id>sun-maven-repository</id>
-      <url>http://download.java.net/maven/2/</url>
+      <url>https://download.java.net/maven/2/</url>
```
nickgros, line 71: I don't think either of these repositories still exist. Should just remove them?

package.json
```diff
@@ -1,32 +1,31 @@
 {
   "dependencies": {
-    "@rjsf/core": "^4.0.1",
```
nickgros, line 3: RJSF is now bundled in SRC

jay-hodgson (approved):

PR #5138, opened 2023-07-12
SWC-6498

jay-hodgson:
Define Content-Security-Policy


src/main/webapp/Portal.html
```diff
@@ -504,9 +509,8 @@
       head.load(cdnEndpoint + 'generated/prop-types.min.js')
       head.load(cdnEndpoint + 'generated/react-measure.js')
       head.load(cdnEndpoint + 'generated/plotly.min.js')
-      head.load(
-        'https://unpkg.com/react-plotly.js@2.2.0/dist/create-plotly-component.min.js',
-      )
+      head.load(cdnEndpoint + 'js/2.2.0_create-plotly-component.min.js')
+      head.load(cdnEndpoint + 'generated/react-jsonschema-form.js')
```
nickgros: react-jsonschema-form should stay removed from the list of scripts to load, no?

Can we add react-plotly.js to the package.json and add build actions to copy the react-plotly.js dependency like we do for the other scripts (so it is captured when Dependabot scans our package.json/yarn.lock files?)

src/main/webapp/Portal.html
```diff
@@ -8,6 +8,11 @@
     />
     <meta name="author" content="Sage Bionetworks, info@sagebase.org" />
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self' data: 'unsafe-inline' 'unsafe-eval' *.synapse.org *.sagebase.org *.googleapis.com *.gstatic.com *.googletagmanager.com cdn.statuspage.io s3.amazonaws.com www.google-analytics.com kh896k90gyvg.statuspage.io 127.0.0.1:9876"
```
nickgros: This seems like an appropriate list. Just wondering how you came up with it?

jay-hodgson: Reviewing this file, and trial and error (spinning up the client and looking for console messages indicating that the browser blocked a resource due to the csp)

src/main/webapp/Portal.html
jay-hodgson: react-jsonschema-form should certainly stay removed, this must have been a bad complex merge.  I'll delete it

jay-hodgson: And you're absolutely right regarding create-plotly-component.min.js from react-plotly.  This is not a special case

src/main/webapp/Portal.html
nickgros: Is it possible to embed external images in Wiki markdown? Wonder if we need to add `img-src *`. [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src).

nickgros: Would this also affect the contents of an `iframe`d HTML preview? I am not currently sure what functionality remains for that feature

jay-hodgson: Nice catch with the img-src, I'll add it

nickgros (approved):

PR #5139, opened 2023-07-12
SWC-6499

jay-hodgson:
Lock down CORS access a bit more to allowed domains only.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
@@ -16,7 +16,6 @@ import org.springframework.web.filter.OncePerRequestFilter;
 public class CORSFilter extends OncePerRequestFilter {
 
   public static final String ORIGIN_HEADER = "origin";
-  public static final String DEFAULT_ALLOW_ORIGIN = "*";
   public static final String ACCESS_CONTROL_ALLOW_ORIGIN_HEADER =
     "Access-Control-Allow-Origin";
   public static final String ACCESS_CONTROL_ALLOW_CREDENTIALS_HEADER =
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -74,32 +73,30 @@ protected void doFilterInternal(
...
         .getHost()
         .substring(0, url.getHost().length() - SYNAPSE_ORG_SUFFIX.length());
       if (ALLOWED_SYNAPSE_SUBDOMAINS.contains(subdomain)) {
-        allowOrigin = origin;
+        response.addHeader(ACCESS_CONTROL_ALLOW_ORIGIN_HEADER, origin);
         response.addHeader(ACCESS_CONTROL_ALLOW_CREDENTIALS_HEADER, "true");
-      }
-    }
 
-    response.addHeader(ACCESS_CONTROL_ALLOW_ORIGIN_HEADER, allowOrigin);
```
nickgros, line 90: Am I understanding correctly that this change would configure us to block all requests to the server APIs from origins that are not in the allowlist?

jay-hodgson: Yes (for calls to Synapse.org servlets).

nickgros: In my [review](https://sagebionetworks.jira.com/wiki/spaces/PLFM/pages/2799960100/Synapse+CORS+Configuration+-+January+2023) of origins that have queried the portal servlets, `https://accounts.google.com` had made ~4,700 requests in a 30-day span. I am not sure what other non-`*.synapse.org` origins also queried the portal, but we can easily get that data from CloudWatch by re-running some of those queries.

Do we know why these requests are made, and would blocking this break some functionality?

jay-hodgson: That makes sense!  This was before my time, but I think we send Google a portal servlet endpoint during the auth workflow.  Could you re-run those queries and document all of the origins?

PR #5142, opened 2023-07-14
SWC-6471

jay-hodgson:


jay-hodgson:
On html page load, use a standard servlet to get the script application/ld+json response (will only return content if entity is a dataset).  We only attempt the fetch when loading a Synapse place.
This method means that we do not need to wait for the GWT app to load.

src/main/java/org/sagebionetworks/web/client/SynapseClient.java
```diff
@@ -327,7 +327,4 @@ void resendTeamInvitation(
     String membershipInvitationId,
     String hostPageBaseURL
   ) throws RestServiceException;
-
-  String getDatasetScriptElementContent(String entityId)
```
jay-hodgson, line 331: Remove GWT RPC call

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -212,67 +212,50 @@ public void invoke() {
   }
 
   public void getEntityBundleAndLoadPageTop() {
-    synapseClient.getDatasetScriptElementContent(
```
jay-hodgson, line 215: This is a terrible diff.  In reality, I just removed the getDatasetScriptElementContent block (looking at the raw source might be preferable).

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -348,10 +331,11 @@ public Widget asWidget() {
 
   @EventHandler
   public void onEntityUpdatedEvent(EntityUpdatedEvent event) {
-    List<SynapseReactClientEntityQueryKey> queryKey = SynapseReactClientEntityQueryKey.create(
-      QueryKeyConstants.ENTITY,
-      event.getEntityId()
-    );
+    List<SynapseReactClientEntityQueryKey> queryKey =
+      SynapseReactClientEntityQueryKey.create(
```
jay-hodgson, line 335: Formatting only

src/main/java/org/sagebionetworks/web/client/view/EntityView.java
```diff
@@ -38,8 +38,4 @@ public interface Presenter {
   void setEntityPageTopVisible(boolean isVisible);
 
   void setOpenTeamInvitesVisible(boolean isVisible);
-
-  void injectDatasetJsonLd(String elementContent);
-
-  void removeDatasetJsonLdElement();
```
jay-hodgson, line 44: This is no longer done by the EntityPresenter.  It's done outside the GWT framework

src/main/java/org/sagebionetworks/web/server/servlet/JsonLdContentServlet.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @Override
+  public void service(ServletRequest arg0, ServletResponse arg1)
+    throws ServletException, IOException {
+    super.service(arg0, arg1);
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
```
jay-hodgson, line 80: Most of this file is boilerplate, with the exception of doGet().  We get the Entity (by ID), and if it's a Dataset then get the Wiki and EntityBundle to construct the Dataset json LD element content. Return this json response

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -337,28 +334,24 @@ private String getDisplayName(UserProfile profile) {
     return displayNameBuilder.toString();
   }
 
-  public static String getPlainTextWiki(
-    String entityId,
-    SynapseClientImpl synapseClient
-  ) {
+  public static String getPlainTextWiki(String entityId, WikiPage rootPage) {
```
jay-hodgson, line 337: Takes the WikiPage instead of a SynapseClientImpl.  Caller now retrieves the WikiPage

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -597,14 +595,15 @@ private String getThreadHtml(String threadId)
...
     html.append("Created by " + createdBy + "<br>");
-    PaginatedResults<DiscussionReplyBundle> replies = discussionForumClient.getRepliesForThread(
-      thread.getId(),
-      100L,
-      0L,
-      DiscussionReplyOrder.CREATED_ON,
-      false,
-      DiscussionFilter.EXCLUDE_DELETED
-    );
+    PaginatedResults<DiscussionReplyBundle> replies =
```
jay-hodgson, line 598: Formatting

xschildw (approved):

PR #5151, opened 2023-07-25
SWC-6502: upload issue with new CSP

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/aws/AwsSdk.java
```diff
@@ -89,11 +89,9 @@ private static native void _upload(
 		var listener = function(err, data) {
 			if (err) {
 				//error
-				console.log("Upload error, retrying:", err.code, err.message);
+				console.log("Upload error", err.code, err.message);
 				upload.abort();
-				setTimeout(function() {
-					upload.send(listener);
-				}, 5000);
```
jay-hodgson, line 96: Was never retrying, and user never receives an error message!  We should just send back the error so the UI is not stuck in infinite loading

src/main/webapp/Portal.html
```diff
@@ -11,7 +11,7 @@
 
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' *.synapse.org *.sagebase.org *.googleapis.com *.gstatic.com *.googletagmanager.com cdn.statuspage.io s3.amazonaws.com www.google-analytics.com kh896k90gyvg.statuspage.io *.youtube.com *.rstudio.com *.google.com *.shinyapps.io *.smartsheet.com 127.0.0.1:9876; img-src * data: blob:"
+      content="default-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' *.synapse.org *.sagebase.org *.googleapis.com *.gstatic.com *.googletagmanager.com cdn.statuspage.io s3.amazonaws.com www.google-analytics.com kh896k90gyvg.statuspage.io *.youtube.com *.rstudio.com *.google.com *.shinyapps.io *.smartsheet.com 127.0.0.1:9876; img-src * data: blob:; connect-src https://* http://127.0.0.1:8888"
```
jay-hodgson, line 14: [connect-src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src) : Synapse is a federated data store, and the client is unaware of all hosts
https://* : So let's set this to all https targets for now
http://127.0.0.1:8888 : Also the local host for development

xschildw (approved):

PR #5162, opened 2023-08-23
SWC-6514, SWC-6523, SWC-6524: end-to-end testing proof of concept

hallieswan:


.github/workflows/build-test-e2e.yml
```diff
@@ -0,0 +1,52 @@
+name: 'Build/Run End-to-End Tests'
+
+on: push
+jobs:
+  build-and-run-e2e:
+    if: ${{ github.ref == 'develop' || github.ref == 'release-**' || github.actor == github.repository_owner }}
```
hallieswan, line 6: Repo secrets can't be accessed in workflows triggered by pull requests from forked repos, as a security measure to prevent exfiltrating secrets. A workflow can access secrets if the `pull_request_target` trigger is used, but would use the base (develop) branch workflow and code, rather than the PR code. Since the e2e tests wouldn't be run against the PR code, this wouldn't be very useful for our use case. So, we will only run end-to-end tests on push to `develop` and `release-*` branches in the `Sage-Bionetworks` repo.

However, since tests that pass locally can be flaky in CI, we would like devs to know the status of e2e tests before merging a PR to develop. So, also run the e2e tests whenever a dev pushes changes to their forked repo. The workflow will be able to access secrets stored within the forked repo, since they are the repo owner.

.github/workflows/build-test-e2e.yml
```diff
@@ -0,0 +1,52 @@
...
+    runs-on: macos-latest
```
hallieswan, line 7: Tests failed intermittently in CI with GitHub standard Linux runner. Using the standard macOS runner, which is larger, eliminated flake in CI.

.github/workflows/build-test-e2e.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Setup docker (missing on MacOS)
+        if: runner.os == 'macos'
+        run: |
+          # workaround for https://github.com/actions/runner-images/issues/8104
+          export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
+          brew update
+          brew reinstall qemu
+
+          brew install docker
+          colima start
```
hallieswan: Docker is installed by default on Linux runners, but [not on macOS runners](https://stackoverflow.com/a/76490257), so install it here.

.github/workflows/build-test-e2e.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Build SWC
+        uses: ./.github/workflows/build
+      - name: Run SWC on Tomcat
+        run: |
+          docker run \
+            --name "${CONTAINER_NAME}" \
+            -d --rm \
+            -p 8888:8080 \
+            -v "/$(pwd)/target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war" \
```
hallieswan, line 33: Rename to`ROOT.war` so tomcat serves app at the base URL, rather than including the war name in the URL.

.github/workflows/build-test-e2e.yml
```diff
@@ -0,0 +1,52 @@
...
+            -v "/$(pwd)/e2e_workflow/settings.xml":/root/.m2/settings.xml \
```
hallieswan, line 34: Provide a settings file for the built app

.github/workflows/build-test-e2e.yml
```diff
@@ -0,0 +1,52 @@
...
+            tomcat:9.0
+      - name: Install Playwright Browsers
+        run: yarn playwright install --with-deps chromium
```
hallieswan, line 37: Only install `chromium` browser to speed up workflow

.github/workflows/build-test-e2e.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Run Playwright tests
+        env:
+          ADMIN_USERNAME: ${{ secrets.ADMIN_USERNAME }}
+          ADMIN_PASSWORD: ${{ secrets.ADMIN_PASSWORD }}
```
hallieswan, line 41: We'll need to add these secrets to the Sage SWC repo before merging these changes to develop

.github/workflows/build/action.yml
```diff
@@ -0,0 +1,14 @@
+name: 'Build SWC'
```
hallieswan, line 1: Move existing SWC build job into a composite action, so it can be used in the existing workflow and in the new e2e workflow.

.github/workflows/maven.yml
```diff
@@ -4,8 +4,6 @@
 name: Java CI with Maven
 
 on:
-  push:
-    branches: [develop, release-**]
```
hallieswan, line 8: The e2e workflow also includes the SWC build and runs on push to `develop` and `release-*`. Stop running this action on push, so SWC build is only run once.

.github/workflows/build-test-e2e.yml
hallieswan: If we decide to use a GitHub-hosted larger runner, tweak this so that a standard runner is still used when devs run the workflow in their forked repo. Otherwise, they will be charged for using the larger runner in their personal account.

e2e/auth.cleanup.ts
```diff
@@ -0,0 +1,58 @@
...
+
+cleanup.describe('Clean up', () => {
+  cleanup.beforeAll(async ({ browser }) => {
+    adminPage = await browser.newPage({
+      storageState: ADMIN_STORAGE_STATE,
+    })
+    userPage = await browser.newPage({
+      storageState: USER_STORAGE_STATE,
+    })
+  })
```
hallieswan, line 23: Create new pages in `beforeAll`, so they can be closed in `afterAll`, even if a step within the test fails. Create the pages using the stored state, so the user and admin are logged in on their respective pages.

e2e/auth.setup.ts
```diff
@@ -0,0 +1,110 @@
...
+let userPage: Page
+
+setup.describe('Setup', () => {
+  setup.beforeAll(async ({ browser }) => {
+    adminPage = await browser.newPage()
+    userPage = await browser.newPage()
+  })
+
+  setup('authenticate users', async () => {
+    setup.slow()
```
hallieswan, line 40: This may be the first test to run and trigger SWC compilation, which may take 15+ seconds. Mark this [test as slow](https://playwright.dev/docs/api/class-test#test-slow-1), so test will be given triple timeout.

e2e/auth.setup.ts
```diff
@@ -0,0 +1,110 @@
...
+    const { testUserName, testUserPassword } = await setup.step(
+      'create test user',
+      async () => {
+        const testUserName = 'swc-e2e-' + uuidv4()
+        const testUserPassword = 'password-' + uuidv4()
+        /* 
+        const testUserName = 'swc-e2e-test'
+        const testUserPassword = process.env.USER_PASSWORD!
+        expect(testUserPassword).not.toBeUndefined() 
+        */
```
hallieswan, line 60: By default, create a user with randomly generated username/password, so that concurrent tests are user-namespaced. Commented out code demonstrates how to create a user with a static credentials, which can be useful for troubleshooting when writing a new test.

e2e/auth.setup.ts
```diff
@@ -0,0 +1,110 @@
...
+  })
+
+  setup('authenticate users', async () => {
+    setup.slow()
+
+    await setup.step('authenticate admin', async () => {
+      const { adminUserName, adminUserPassword } = getAdminUserCredentials()
+
+      await loginTestUser(adminPage, adminUserName!, adminUserPassword!)
+      await adminPage.context().storageState({ path: ADMIN_STORAGE_STATE })
```
hallieswan, line 46: Store the [authenticated state](https://playwright.dev/docs/auth#basic-shared-account-in-all-tests), so that it can be reused across tests.

e2e/auth.setup.ts
```diff
@@ -0,0 +1,110 @@
...
+    // ...so we delete here, so we can retry setup without leaving behind test users
+    if (status !== expectedStatus && testUserId !== undefined) {
+      const result = await deleteTestUser(
+        getEndpoint(),
+        testUserId!,
+        adminAccessToken,
+      )
+      expect(result).toEqual(testUserId)
+    }
+  })
```
hallieswan, line 113: We create the test user in `auth.setup` and delete the test user in `auth.cleanup`. However, if `auth.setup` is retried, then `auth.cleanup` won't be run until after the last retry. Delete the test user here if the test failed or was skipped, so that setup can be retried without leaving behind test users.

e2e/helpers/http.ts
```diff
@@ -0,0 +1,141 @@
...
+  base64EncodedSecretKey: string,
+) {
+  const hash = createHmac(
+    'sha1',
+    Buffer.from(base64EncodedSecretKey, BASE64_ENCODING),
+  )
+    .update(data)
+    .digest(BASE64_ENCODING)
+  return hash
+}
```
hallieswan, line 26: Used for admins using an API key, rather than an access token

e2e/helpers/localStorage.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export async function setLocalStorage(page: Page, key: string, value: string) {
+  await navigateToHomepageIfPageHasNotBeenLoaded(page)
+  await page.evaluate(
+    ([key, value]) => {
+      localStorage.setItem(key, value)
+    },
+    [key, value],
+  )
+}
```
hallieswan, line 18: Navigate to the home page if the page hasn't been loaded, since localStorage isn't available until the page has initially loaded.

e2e/helpers/srcFetch.ts
```diff
@@ -0,0 +1,161 @@
+export const NETWORK_UNAVAILABLE_MESSAGE =
```
hallieswan, line 1: The SynapseClient doesn't expose the `fetchWithExponentialTimeout` method or `doGet` etc. Replicated SRC code here, so new API calls (e.g. creating a user, deleting a message) can be made.

e2e/helpers/testUser.ts
```diff
@@ -0,0 +1,104 @@
...
+  await page.getByLabel('Username or Email Address').fill(testUserName)
+  await page.getByLabel('Password').fill(testUserPassword)
+  await page.getByRole('button', { name: 'Sign in' }).click()
+
+  // Wait until the page reaches a state where all cookies are set
+  await expect(page.getByLabel('Search')).toBeVisible()
+  await expect(page.getByLabel('Projects')).toBeVisible()
+  await expect(page.getByLabel('Your Account')).toBeVisible({
+    timeout: 30 * 1000,
+  })
```
hallieswan, line 76: Wait for the various sidebar buttons that indicate a logged in state, in the order in which they appear

e2e/helpers/types.ts
```diff
@@ -0,0 +1,73 @@
+/* USERS */
+export type LoginResponse = {
```
hallieswan, line 2: Replicated various types from [synapse-types](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-types) (LoginResponse, FileHandleStatus, FileHandle, PaginatedResults) -- is there a way to pull these types from the package directly?

e2e/projects.loggedin.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { Page, expect, test } from '@playwright/test'
+import { v4 as uuidv4 } from 'uuid'
+import { USER_STORAGE_STATE } from '../playwright.config'
+import { goToDashboard } from './helpers/testUser'
+
+const PROJECT_NAME = 'swc-e2e-project-' + uuidv4()
```
hallieswan, line 6: Use a unique project name per test run

e2e/teams.loggedin.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+test.describe('Teams', () => {
+  test.beforeAll(async ({ browser }) => {
+    adminPage = await browser.newPage({
+      storageState: ADMIN_STORAGE_STATE,
+    })
+    userPage = await browser.newPage({
+      storageState: USER_STORAGE_STATE,
+    })
+  })
```
hallieswan, line 31: Could refactor to create [authenticated page fixtures](https://sagebionetworks.jira.com/jira/software/c/projects/SWC/issues/SWC-6528), which would remove the need to create new pages in the `beforeAll` hook

e2e/teams.loggedin.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+      await userPage.getByLabel('Invitation Message').type(INVITATION_MESSAGE)
+
+      const spinner = userPage.locator('.modal-body > .spinner')
+      await Promise.all([
+        expect(spinner).toBeVisible(),
+        userPage.getByRole('button', { name: 'Send Invitation(s)' }).click(),
+      ])
+
+      await expect(spinner).not.toBeVisible()
```
hallieswan, line 85: After the user clicks on the send invitation button, a spinner appears while waiting for some work to finish, then disappears when the work is done. The test expects the spinner to appear and disappear. However, depending on how quickly the work finishes, the spinner may appear and disappear before Playwright has checked for its existence. Use `Promise.all` to ensure that the check for the spinner is fired before the button is clicked and the spinner's appearance isn't missed. See more discussion [here](https://github.com/microsoft/playwright/issues/5470#issuecomment-1285640689) and an illustrated example [here](https://tally-b.medium.com/the-illustrated-guide-to-using-promise-all-in-playwright-tests-af7a98af3f32).

e2e/teams.loggedin.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+      // handle case where occasionally two team links are shown
+      // ...before resolving to one team link
+      await expect
+        .poll(async () => {
+          return await teamLink.count()
+        })
+        .toBe(1)
+
+      await teamLink.click()
```
hallieswan, line 128: Playwright locators are [strict](https://playwright.dev/docs/locators#strictness), so an exception will be thrown if we try to click on a locator that matches more than one element. Poll will wait for the link to resolve to a single element before trying to click on it.

e2e/teams.loggedin.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    await goToDashboard(userPage)
+    await userPage.getByLabel('Teams').click()
+    await userPage.getByRole('link', { name: TEAM_NAME }).click()
+    await userPage.getByRole('button', { name: 'Team Actions' }).click()
+    await userPage.getByRole('link', { name: 'Delete Team' }).click()
+    await userPage.getByRole('button', { name: 'Delete' }).click()
+    await expect(userPage.getByText('Team successfully deleted')).toBeVisible()
+    await expect(
+      userPage.getByRole('link', { name: TEAM_NAME }),
+    ).not.toBeVisible()
```
hallieswan, line 164: Clean up the team in the `afterAll` hook, so that team is cleaned up even if the test fails part way through execution. Allows us to retry the test without cluttering the dev stack with teams.

If Teams are commonly used across tests, could consider creating a [Team fixture](https://playwright.dev/docs/test-fixtures#creating-a-fixture), which encapsulates creating and deleting a team.

e2e_workflow/settings.xml
```diff
@@ -0,0 +1,32 @@
+<settings
```
hallieswan, line 2: Settings file used when serving war with tomcat

package.json
```diff
@@ -13,25 +13,32 @@
...
     "react-transition-group": "2.6.0",
-    "react-plotly.js": "2.2.0",
     "sass": "^1.63.6",
     "synapse-react-client": "3.1.49",
     "universal-cookie": "^4.0.4"
   },
   "devDependencies": {
+    "@playwright/test": "1.37.0",
     "@prettier/plugin-xml": "^3.1.0",
+    "dotenv": "^16.3.1",
```
hallieswan, line 27: Manages environment variables locally

package.json
```diff
@@ -13,25 +13,32 @@
...
     "husky": ">=6",
     "lint-staged": ">=10",
     "prettier": "^3.0.0",
-    "prettier-plugin-java": "^2.2.0"
+    "prettier-plugin-java": "^2.2.0",
+    "uuid": "^9.0.0"
```
hallieswan, line 32: Create random UUIDs for projects and teams

playwright.config.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    /* {
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    }, */
```
hallieswan, line 90: Currently, we only test against chromium, but could test against other browsers in the future

playwright.config.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    /* Test against mobile viewports. */
+    // {
+    //   name: 'Mobile Chrome',
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
```
hallieswan, line 100: And against different viewport sizes

playwright.config.ts
```diff
@@ -0,0 +1,117 @@
...
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
+  webServer: {
+    command: process.env.CI ? '' : 'mvn gwt:run',
+    url: baseURL,
+    reuseExistingServer: true, // on CI, will use the tomcat server
+  },
```
hallieswan, line 116: If running Playwright locally, start SWC if it isn't already running. However, in CI, always wait for tomcat to serve SWC.

yarn.lock
```diff
@@ -1045,6 +1055,11 @@
   dependencies:
     lz-string "*"
 
+"@types/node@*":
+  version "20.4.4"
+  resolved "https://registry.yarnpkg.com/@types/node/-/node-20.4.4.tgz#c79c7cc22c9d0e97a7944954c9e663bcbd92b0cb"
+  integrity sha512-CukZhumInROvLq3+b5gLev+vgpsIqC2D0deQr/yS1WnxvmYLlJXZpaQrQiseMY+6xusl79E04UjWoqyr+t1/Ew==
```
nickgros, line 1061: We might want to lock down the NodeJS version and these types, but that's not necessary in this PR

playwright.config.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: './e2e',
+  /* Timeout to allow portal enough time to compile when running locally */
+  timeout: 5 * 60 * 1000,
+  /* Increase expectation timeout on CI */
+  expect: { timeout: process.env.CI ? 30 * 1000 : 5 * 1000 },
```
nickgros, line 25: That's handy

playwright.config.ts
```diff
@@ -0,0 +1,117 @@
...
+  /* Run tests in files in parallel */
+  fullyParallel: true,
```
nickgros, line 27: Whoa, great that this works (locally)

playwright.config.ts
```diff
@@ -0,0 +1,117 @@
...
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Limit the number of failures on CI to save resources */
+  maxFailures: process.env.CI ? 10 : undefined,
```
nickgros, line 31: So the job will abort after 10 failures? If so, that's reasonable

playwright.config.ts
```diff
@@ -0,0 +1,117 @@
...
+    // Setup project
+    { name: 'setup', testMatch: /.*\.setup\.ts/, teardown: 'cleanup' },
+
+    // Tests that don't require authentication
+    {
+      name: 'chromium - logged out',
+      use: {
+        ...devices['Desktop Chrome'],
+      },
+      testMatch: /.*\.loggedout\.spec.ts/,
```
nickgros, line 60: Is there another way to add metadata to these test suites instead of manipulating the path/filename? This doesn't scale if we add more flags.

We don't need to figure this out to merge, but maybe create a new ticket to follow up later.

If such a mechanism also works in Jest, I think it would also be useful in SRC.

e2e_workflow/settings.xml
nickgros: Maybe add that as a comment in this XML file

e2e_workflow/README.md
nickgros, line 1: Fantastic doc

e2e/auth.setup.ts
nickgros: Can you add this as a comment above this line?

e2e/auth.setup.ts
nickgros: Great work handling this edge case

e2e/create_account.loggedout.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  test('should show an alert when an invalid email address is used', async ({
+    page,
+  }) => {
+    await page.goto('/')
+    // Allow time for SWC compilation
+    await expect(
+      page.getByRole('heading', { name: 'Loading…' }),
+    ).not.toBeVisible(
+      { timeout: 2 * 60 * 1000 }, // ...wait 2 minutes if necessary
+    )
```
nickgros: Why this instead of `.slow()` as used above? If this is more reliable, maybe make this a reusable function that we can call at the beginning of each test suite

e2e/helpers/testUser.ts
```diff
@@ -0,0 +1,104 @@
...
+  testUserName: string,
+  testUserPassword: string,
+) {
+  // Perform authentication steps
+  await page.goto('/')
+
+  // Allow time for SWC compilation
+  await expect(page.getByRole('heading', { name: 'Loading…' })).not.toBeVisible(
+    { timeout: 2 * 60 * 1000 }, // ...wait 2 minutes if necessary
+  )
```
nickgros: Refactor into separate helper as suggested above

e2e/helpers/types.ts
nickgros: Nope because (for now) it's private to the monorepo. I think copying it is fine for now.

Kind of want to wait for https://sagebionetworks.jira.com/browse/PLFM-7768 to be done to spend more time wrangling types or SynapseClient. 

I guess in the meantime maybe we could publish the @sage-bionetworks/synapse-types package as v0.0.1 or something like that--feel free to create a ticket for us to track that

e2e/create_account.loggedout.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    await page.getByRole('link', { name: 'Register Now' }).first().click()
+    await page.getByRole('textbox').first().fill('test123')
+    await page.getByPlaceholder('Your email address').fill('test123')
+    await page.getByRole('button', { name: 'Send registration info' }).click()
+    await expect(page.locator('#rootPanel').getByRole('alert')).toHaveText(
```
nickgros, line 15: Some of these selectors might be too specific to be resilient (what if there's >1 alert on the page?) but we can loosen them over time, no change needed.

e2e/teams.loggedin.spec.ts
nickgros: Thanks for creating the issue to track

e2e/teams.loggedin.spec.ts
nickgros: I see--so you're basically priming the visibility check to start looking before the action is invoked, which ensures that you don't miss it.

This is a useful pattern, and I think it might make sense to use in other contexts--one that comes to mind immediately is using mock-service-worker to update mocked server behavior when simulating a data mutation.

e2e/teams.loggedin.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+      await row.getByRole('button', { name: 'Join' }).click()
+      await expect(row).not.toBeVisible()
+    })
+
+    await test.step('admin should view team page', async () => {
+      const teamLink = adminPage.getByRole('link', { name: TEAM_NAME })
+
+      // handle case where occasionally two team links are shown
+      // ...before resolving to one team link
```
nickgros, line 121: Not totally clear to me why this might happen

.github/workflows/build-test-e2e.yml
nickgros: I'll go ahead and do that

nickgros (changes requested):
This looks great--for me the only issues are
1. Couple of PR comments that I think should be moved into the code (called out)
2. Is it possible to refactor/reuse code that waits for the initial compilation/"Loading..." splash screen?

e2e/helpers/types.ts
hallieswan: Got it, thanks! I've created a ticket to track creating the synapse-types package [here](https://sagebionetworks.jira.com/browse/SWC-6529).

playwright.config.ts
hallieswan: I'm not sure whether we can add metadata directly to the test suites themselves. However, _I think_ that [SWC-6528](https://sagebionetworks.jira.com/jira/software/c/projects/SWC/issues/SWC-6528) will allow us to stop using the file name to separate logged in vs logged out tests -- we would use a standard page fixture for logged out tests and an authenticated page fixture for logged in tests, but related tests could live within the same file.

e2e/create_account.loggedout.spec.ts
hallieswan: `test.slow()` increases the [timeout of the test](https://playwright.dev/docs/next/test-timeouts#test-timeout), but I don't believe it extends the timeout of a particular expectation. So if SWC compilation took longer than the standard expectation timeout, the test would fail, even if the test timeout was extended. Instead, we can just increase the timeout of this assertion to account for SWC compilation. I'll break it out into a reusable function!

e2e/teams.loggedin.spec.ts
hallieswan: I saw it intermittently when I was using the smaller runner -- I'm not sure if it was an artifact of the runner being resource constrained or a real state. Happy to remove the count check if seeing two team links isn't something we expect to see regularly

e2e/helpers/srcFetch.ts
jay-hodgson: What do you think about exposing SRC.SynapseClient fetchWithExponentialTimeout (and any other functions that we need)?  I'd hate to have fixes made to these SynapseClient functions (or srcFetch functions) and have them drift apart

hallieswan: Agreed that there's potential for drift! Exposing SRC's HttpClient would remove the need for duplicated functions, since we would have access to `doGet`, `doPost`, etc... which use `fetchWithExponentialTimeout`. Would you like me to include those changes as part of this PR or split the work into a separate ticket? 

We would still need to duplicate a few types (LoginResponse, FileHandleStatus, FileHandle, PaginatedResults) until the synpase-types package is also exposed, as discussed with @nickgros [here](https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5162/files#r1303339880) and associated ticket [here](https://sagebionetworks.jira.com/browse/SWC-6529).

nickgros (approved):

jay-hodgson (approved):

e2e/helpers/srcFetch.ts
hallieswan: I created a [ticket](https://sagebionetworks.jira.com/browse/SWC-6531) to track exposing the HttpClient and removing the duplicate code.

PR #5163, opened 2023-08-24
SWC-6514: use the short ref name of the branch, rather than the full ref

hallieswan:


.github/workflows/build-test-e2e.yml
```diff
@@ -3,7 +3,7 @@ name: 'Build/Run End-to-End Tests'
 on: push
 jobs:
   build-and-run-e2e:
-    if: ${{ github.ref == 'develop' || github.ref == 'release-**' || github.actor == github.repository_owner }}
```
hallieswan, line 6: Per [GitHub docs](https://docs.github.com/en/actions/learn-github-actions/contexts):

`github.ref` is "The fully-formed ref of the branch or tag that triggered the workflow run. For workflows triggered by push, this is the branch or tag ref that was pushed...The ref given is fully-formed, meaning that for branches the format is refs/heads/<branch_name>, for pull requests it is refs/pull/<pr_number>/merge, and for tags it is refs/tags/<tag_name>. For example, refs/heads/feature-branch-1."

`github.ref_name` is "The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, feature-branch-1."

Use `github.ref_name` so the conditional evaluates the correct values.

nickgros (commented):
LGTM

PR #5164, opened 2023-08-25
SWC-6537: remove reference to /table/sql/transform service call

jay-hodgson:


jay-hodgson:
Verified that the sql, that incorporates any selected facets, is accurate in the programmatic download options:
Screen Shot 2023-08-25 at 12 30 14 PM

nickgros (approved):

PR #5168, opened 2023-08-25
Bump synapse-react-client to v3.1.51

nickgros:


jay-hodgson:
This has a bug fix on top of 3.1.50.  
[See release notes for 3.1.50](https://github.com/Sage-Bionetworks/synapse-web-monorepo/releases/tag/synapse-react-client%2Fv3.1.50)

This includes a SynapseTable rewrite (uses tanstack) from @nickgros , and adding files from any table or entity view (that has files) to the Download Cart from @jay-hodgson .

jay-hodgson (approved):

PR #5172, opened 2023-08-31
SWC-6545 - fix broken (?) query cache keys

nickgros:
I discovered this defect when preparing to upgrade to @tanstack/react-query v4, but this change does not require that upgrade. I would expect this to be a source of bugs, but I can't actually (re?)produce any of the bugs I would expect to see...

- Use SRC's KeyFactory to generate query keys for invalidation


src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -331,24 +320,23 @@ public Widget asWidget() {
...
   ) {
-    List<String> queryKey = Arrays.asList(
-      authenticationController.getCurrentUserAccessToken(),
-      QueryKeyConstants.DOWNLOAD_LIST
+    KeyFactory keyFactory = new KeyFactory(
+      authenticationController.getCurrentUserAccessToken()
     );
-    queryClient.invalidateQueries(queryKey);
+    queryClient.invalidateQueries(keyFactory.getDownloadListBaseQueryKey());
   }
```
nickgros, line 340: This invalidation recently broke when we changed `KeyFactory` to use the token hash instead of the token itself.

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -331,24 +320,23 @@ public Widget asWidget() {
...
-      );
-    queryClient.resetQueries(queryKey);
+    KeyFactory keyFactory = new KeyFactory(
+      authenticationController.getCurrentUserAccessToken()
+    );
+
+    List<?> queryKey = keyFactory.getEntityQueryKey(event.getEntityId());
+    queryClient.invalidateQueries(queryKey);
     globalAppState.refreshPage();
   }
```
nickgros, line 330: This invalidation doesn't include the token in the query key, so who knows how long it's been broken...let's use `KeyFactory` to reliably generate the correct key.

src/main/java/org/sagebionetworks/web/client/jsinterop/JSON.java
```diff
@@ -7,4 +7,6 @@
 public class JSON {
 
   public static native Object parse(String json);
+
+  public static native String stringify(Object obj);
```
nickgros, line 11: generally useful for debugging and inspecting JS native objects

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -331,24 +324,23 @@ public Widget asWidget() {
...
-      );
-    queryClient.resetQueries(queryKey);
+    KeyFactory keyFactory = keyFactoryProvider.getKeyFactory(
+      authenticationController.getCurrentUserAccessToken()
+    );
+
+    List<?> queryKey = keyFactory.getEntityQueryKey(event.getEntityId());
+    queryClient.invalidateQueries(queryKey);
     globalAppState.refreshPage();
   }
```
nickgros, line 334: What's unusual is that I can't reproduce the bug I expect, where stale data would be shown in certain places after this event is fired.

This invalidation is missing the hashed token as the first argument in the query key, let's use SRC's KeyFactory to reliably generate the correct key.

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -331,24 +324,23 @@ public Widget asWidget() {
...
   ) {
-    List<String> queryKey = Arrays.asList(
-      authenticationController.getCurrentUserAccessToken(),
-      QueryKeyConstants.DOWNLOAD_LIST
+    KeyFactory keyFactory = keyFactoryProvider.getKeyFactory(
+      authenticationController.getCurrentUserAccessToken()
     );
-    queryClient.invalidateQueries(queryKey);
+    queryClient.invalidateQueries(keyFactory.getDownloadListBaseQueryKey());
   }
```
nickgros, line 344: Can't reproduce the expected bug here either, but we recently changed KeyFactory to use the token hash in the query key, instead of the token itself. Either way, just use the same class to generate the key as SRC.

src/main/java/org/sagebionetworks/web/client/context/KeyFactoryProvider.java
```diff
@@ -0,0 +1,7 @@
+package org.sagebionetworks.web.client.context;
+
+import org.sagebionetworks.web.client.jsinterop.KeyFactory;
+
+public interface KeyFactoryProvider {
+  public KeyFactory getKeyFactory(String currentUserAccessToken);
+}
```
nickgros, line 7: Create a mockable/injectable KeyFactoryProvider so I can easily mock KeyFactory in tests that don't support JsInterop

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/SynapseReactClientEntityQueryKey.java
nickgros, line 1: No longer need this object that represented one type of query key, since we can now just get a query key from `KeyFactory`.

nickgros:
#5230

PR #5178, opened 2023-09-05
SWC-6531: refactor SWC e2e helpers to use exposed SRC methods and remove duplicate code

hallieswan:


e2e/helpers/http.ts
```diff
@@ -1,141 +1,107 @@
...
-}
-
-export function getUserIdFromJwt(token: string) {
-  const payload = JSON.parse(
-    Buffer.from(token.split('.')[1], BASE64_ENCODING).toString(),
-  )
-  return payload.sub
+export enum BackendDestinationEnum {
+  REPO_ENDPOINT,
+  PORTAL_ENDPOINT,
```
hallieswan, line 6: Replicate [`BackendDestinationEnum`](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/bc092bdea2b0fd67f6bb897fd6f22f556e3a0ab2/packages/synapse-react-client/src/utils/functions/getEndpoint.ts#L1), so that the endpoint can be passed to `doGet`, etc... without evaluating the page to access SRC.

e2e/helpers/http.ts
```diff
@@ -1,141 +1,107 @@
...
-  }
+export async function waitForSrcEndpointConfig(page: Page) {
+  // window only available after page has initially loaded
+  await navigateToHomepageIfPageHasNotBeenLoaded(page)
+  // ensure that endpoint config is set,
+  // ...so API calls point to the correct stack
+  await expect(async () => {
+    const response = await page.evaluate('window.SRC.OVERRIDE_ENDPOINT_CONFIG')
+    expect(response).not.toBeUndefined()
+  }).toPass()
```
hallieswan, line 17: Method moved from e2e/helpers/verification.ts, so it is colocated with other methods that rely on evaluating the page to access SRC

e2e/helpers/http.ts
```diff
@@ -1,141 +1,107 @@
...
+        endpoint
+      ]
+      // @ts-expect-error: Cannot find name 'SRC'
+      return await SRC.HttpClient.doPost(
+        url,
+        requestJsonObject,
+        accessToken,
+        srcEndpoint,
+        additionalOptions,
+      )
```
hallieswan, line 48: Maintain a `doPost` method that evaluates the page to call the `SRC.HttpClient.doPost` method to reduce code duplication in API calls. Plan to remove this method (and `doGet` and `doDelete`) once SynapseClient is its own package that can be used directly instead

e2e/helpers/testUser.ts
```diff
@@ -12,34 +13,39 @@ export function getAdminPAT() {
   return adminPAT
 }
 
+function getUserIdFromJwt(token: string) {
+  const payload = JSON.parse(
+    Buffer.from(token.split('.')[1], BASE64_ENCODING).toString(),
+  )
+  return payload.sub
+}
```
hallieswan, line 21: Moved from e2e/helpers/http.ts, since method is only used in this file

e2e/helpers/http.ts
```diff
@@ -1,141 +1,107 @@
...
 }
 
 export async function doPost<T>(
-  endpoint: string,
-  uri: string,
-  requestContent: string,
-  accessToken?: string,
-  userName?: string,
-  apiKey?: string,
+  page: Page,
```
nickgros, line 21: Instead of passing a `Page` around, you could consider configuring the test process to point to the correct stack using an environment variable, probably by reading the Maven settings.xml so there's just one source of truth. Then I think a lot of this code could run without requiring the browser and loaded page.

You wouldn't have confidence that the `SRC.OVERRIDE_ENDPOINT_CONFIG` is properly set, but I don't think that's such a big deal.

What do you think?

nickgros (approved):
Generally looks good, I had one substantial design question to think about

e2e/helpers/http.ts
hallieswan: I like that idea for looking up the correct stack! Since we'll still need to evaluate the page to use SRC's HttpClient and SynapseClient, I might hold off on implementing that for now. However, I've opened [a ticket](https://sagebionetworks.jira.com/browse/SWC-6550) to track the work and capture this idea!

PR #5179, opened 2023-09-05
SWC-6536: use larger runner for e2e tests

hallieswan:


.github/workflows/build-test-e2e.yml
```diff
@@ -4,7 +4,7 @@ on: push
 jobs:
   build-and-run-e2e:
     if: ${{ github.ref_name == 'develop' || github.ref_name == 'release-**' || github.actor == github.repository_owner }}
-    runs-on: macos-latest
+    runs-on: ${{ github.repository_owner == 'Sage-Bionetworks' && 'ubuntu-22.04-4core-16GBRAM-150GBSSD' || 'macos-latest' }}
```
hallieswan, line 7: Use larger runner when e2e tests are run in Sage-Bionetworks organization, but continue using `macos` runner in dev forks

nickgros: Good idea. Let's do this in the monorepo as well

.github/workflows/build-test-e2e.yml
```diff
@@ -4,7 +4,7 @@ on: push
 jobs:
   build-and-run-e2e:
```
nickgros, line 5: I think you had a timeout at the application level, so maybe it's not necessary, but do you think one should be added at the job level?

nickgros (approved):

.github/workflows/build-test-e2e.yml
hallieswan: Oh good call! I'll add that now

.github/workflows/build-test-e2e.yml
nickgros: https://sagebionetworks.jira.com/browse/PORTALS-2798

PR #5180, opened 2023-09-06
SWC-6536: ensure that userId is set before saving storage state

hallieswan:


e2e/auth.setup.ts
```diff
@@ -47,6 +48,10 @@ for (const {
 
       await setup.step('authenticate test user', async () => {
         await loginTestUser(userPage, user.username, user.password)
+        // ensure that session_marker, i.e. user id, is set before saving state
+        await expect(async () => {
+          await getUserIdFromLocalStorage(userPage)
+        }).toPass()
         await userPage.context().storageState({ path: storageStatePath })
```
hallieswan, line 55: Tests run more quickly on the [larger runner](https://github.com/Sage-Bionetworks/SynapseWebClient/actions/runs/6100063541/job/16553606126), so ensure that userId is set before saving state

nickgros: It's so funny to me when slow machines hide race conditions

nickgros (approved):

PR #5181, opened 2023-09-07
SWC-6531/SWC-6536: fixes for using a larger runner and SRC methods

hallieswan:


e2e/helpers/testUser.ts
```diff
@@ -82,6 +75,11 @@ export async function loginTestUser(
   await page.getByLabel('Password').fill(testUserPassword)
   await page.getByRole('button', { name: 'Sign in' }).click()
 
+  // Wait for redirect
+  await expect(async () => {
+    expect(page.url()).not.toContain('LoginPlace')
+  }).toPass()
+
```
hallieswan, line 82: As Lawrence suggested, wait for the login redirect so if a test fails, we can more easily tell on which page the failure occurred. 

Per [the docs](https://playwright.dev/docs/api/class-frame#frame-wait-for-navigation), `waitForNavigation` is inherently racy and is deprecated in favor of `waitForURL`. However, since SWC login redirects to different pages depending on the last place, we may not know the URL of the redirect. Instead, we can wait for the URL to change from the LoginPlace to another value.

e2e/helpers/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+  )
+}
+
+export async function navigateToHomepageIfPageHasNotBeenLoaded(page: Page) {
+  if (page.url() === 'about:blank') {
+    // Navigate to homepage, so that localStorage and SRC are available
+    await page.goto('/')
+    await waitForInitialPageLoad(page)
+  }
+}
```
hallieswan, line 17: The first test that runs (`auth.setup`) creates a test user by evaluating the page to call SRC's doPost. I'm wondering if the flaky test in [this run](https://github.com/Sage-Bionetworks/SynapseWebClient/actions/runs/6113921796/job/16594399197) was related to evaluating the page before the initial page load finished. I added an await for the initial page load here to hopefully avoid that issue in the future.

nickgros (approved):

PR #5182, opened 2023-09-07
SWC-6547: move papaparse to sage cdn and monitor in package.json

jay-hodgson:
This PR updates the version of papaparse that we are using (from 5.2.0 to ^5.4.1), moves the js to the Synapse CloudFront CDN, and moves the import into package.json where we can be alerted to vulnerabilities and updates.


nickgros (approved):
Looks good!

PR #5190, opened 2023-09-15
SWC-6528: use fixtures for authenticated user pages

hallieswan:
- Use worker fixtures to create/delete users per worker/browser and store authenticated user state
- Use test fixtures to create/close authenticated user pages
- Run tests in parallel locally, but serially in CI
- Run tests against chromium, firefox, and webkit browsers
- Bump playwright version


.github/workflows/build-test-e2e.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             -v "/$(pwd)/e2e_workflow/settings.xml":/root/.m2/settings.xml \
             tomcat:9.0
       - name: Install Playwright Browsers
-        run: yarn playwright install --with-deps chromium
+        run: yarn playwright install --with-deps
```
hallieswan, line 38: Install all Playwright browsers and dependencies, rather than just chromium

e2e/auth.cleanup.ts
hallieswan, line 1: Moved into worker fixture

e2e/auth.setup.ts
hallieswan, line 1: Moved into worker fixture

e2e/favorites.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { expect, test } from '@playwright/test'
+import { testAuth } from './fixtures/authenticatedUserPages'
+import { goToDashboard } from './helpers/testUser'
+
+test.describe('Favorites', () => {
+  testAuth('should be visible', async ({ userPage }) => {
```
hallieswan, line 6: Use the `userPage` test fixture, rather than configuring the 'loggedin' project to have a `page` with user's `storageState`. Allows tests to be organized by subject, rather than by authentication state, i.e. file renamed from `favorites.loggedin.spec.ts` to `favorites.spec.ts`.

e2e/fixtures/authenticatedUserPages.ts
```diff
@@ -0,0 +1,141 @@
...
+type AuthenticatedUserPageWorkerFixtures = {
+  createUsers: void
+  storageStatePaths: StorageStatePaths
+}
+
+const createPage = async (browser: Browser) => {
+  const page = await browser.newPage()
+  // specify baseURL so we call this method outside of a test
+  // ...i.e. where the baseURL test fixture is not available
+  await page.goto(baseURL)
```
hallieswan, line 35: We use the page to call SRC methods during worker fixture cleanup, which is outside of a test.

e2e/fixtures/authenticatedUserPages.ts
```diff
@@ -0,0 +1,141 @@
...
+    const context = await browser.newContext({
+      storageState: storageStatePaths[userPrefix],
+    })
+    const validatedUserPage = await context.newPage()
+    await use(validatedUserPage)
+    await context.close()
+  }
+}
+
+export const testAuth = base.extend<
```
hallieswan, line 56: Use `testAuth` for these fixtures to distinguish between unauthenticated and authenticated tests in the same test file.

e2e/fixtures/authenticatedUserPages.ts
```diff
@@ -0,0 +1,141 @@
...
+          await loginTestUser(userPage, user.username, user.password)
+
+          // ensure that session_marker, i.e. user id, is set before saving state
+          await expect(async () => {
+            await getUserIdFromLocalStorage(userPage)
+          }).toPass()
+        })
+
+        await base.step(`save storageState: ${user.username}`, async () => {
+          const storageStateFileName = `${browserName}_${process.env.TEST_PARALLEL_INDEX}_${user.username}.json`
```
hallieswan, line 120: Worker fixtures are setup/torn down per browser + worker, so include those identifiers in the storageState file for clarity when testing locally.

e2e/helpers/userConfig.ts
```diff
@@ -29,37 +27,22 @@ const generateUserEmail = (prefix: string) => {
...
-  localStorageKey: string
-  user: TestUser
+  [key in UserPrefixes]: TestUser
 }
 
-export const userConfigs: UserConfig[] = [
-  {
-    testName: userNamePrefix,
-    storageStatePath: USER_STORAGE_STATE,
-    localStorageKey: USER_NAME_LOCALSTORAGE_KEY,
```
hallieswan, line 44: Instead of storing the username in local storage, look up the username using the test user's prefix.

e2e/helpers/utils.ts
```diff
@@ -1,11 +1,25 @@
...
+
 // Use after initially navigating to baseURL
 // ...to give SWC time to compile, if necessary
 export async function waitForInitialPageLoad(page: Page) {
   await expect(page.getByRole('heading', { name: 'Loading…' })).not.toBeVisible(
     { timeout: 2 * 60 * 1000 }, // ...wait 2 minutes if necessary
   )
+  // ensure that the SRC endpoint has been set
+  // ...before navigation initiates API calls
+  await waitForSrcEndpointConfig(page)
```
hallieswan, line 22: Ensure that SRC endpoint has been set before navigating away from homepage -- hope this resolves occasional flakiness when navigating to dashboard, as in [this run](https://github.com/Sage-Bionetworks/SynapseWebClient/actions/runs/6127124526/job/16632396261), where Projects button doesn't appear after clicking on 'View Your Dashboard'.

e2e/homepage.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { expect, test } from '@playwright/test'
+import { testAuth } from './fixtures/authenticatedUserPages'
+import { waitForInitialPageLoad } from './helpers/utils'
+
+test.describe('Homepage', () => {
+  test('should show Log In To Synapse button when logged out', async ({
```
hallieswan, line 6: Demonstrate unauthenticated and authenticated tests in the same file, although `testAuth` could be used for both unauthenticated and authenticated tests since only the fixtures needed for a particular test are setup for that test.

e2e/projects.spec.ts
hallieswan, line 1: File renamed (projects.loggedin.spec.ts -> projects.spec.ts) and updated to use the authenticated user page test fixture. Otherwise, mostly unchanged -- I'll call out changes.

e2e/projects.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+  await projectNameInput.fill(projectName)
+  await expect(projectNameInput).toHaveValue(projectName)
+
+  await page.getByRole('button', { name: 'Save' }).click()
+}
+
+// Run multiple describes in parallel, but run tests inside each describe in order
+// ...tests within describe expect afterAll to be run with the same users, i.e. on the same worker
+// https://playwright.dev/docs/api/class-test#test-describe-configure
+test.describe.configure({ mode: 'serial' })
```
hallieswan, line 22: Adding this block so that other test files can be [fully parallelized](https://playwright.dev/docs/test-parallel#parallelize-tests-in-a-single-file), i.e. tests within the same file can be split across workers, but this test describe will run serially on the same worker -- the `afterAll` should only run once, even if other tests are added to this describe. If this test describe was split across workers, then `afterAll` would be run for each test.

e2e/projects.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+      async () => {
+        await createProject(userPage, PROJECT_NAME)
+        await expect(
+          userPage.getByText(/an entity with the name.*already exists/i),
+        ).toBeVisible()
+      },
+    )
+  })
+
+  testAuth.afterAll(async ({ userPage }) => {
```
hallieswan, line 48: Use authenticated page test fixture, rather than opening a new page in `afterAll`

e2e/teams.spec.ts
hallieswan, line 1: File renamed (teams.loggedin.spec.ts -> teams.spec.ts) and updated to use the authenticated user page test fixture. Otherwise, mostly unchanged -- I'll call out changes.

e2e/teams.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+
+// Run multiple describes in parallel, but run tests inside each describe in order
+// ...tests within describe expect afterAll to be run with the same users, i.e. on the same worker
+// https://playwright.dev/docs/api/class-test#test-describe-configure
+test.describe.configure({ mode: 'serial' })
+
+test.describe('Teams', () => {
+  testAuth(
+    'should exercise team lifecycle',
+    async ({ userPage, validatedUserPage }, testInfo) => {
```
hallieswan, line 31: Use authenticated user page test fixtures, rather than opening new pages for the test

e2e/teams.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+      const { userName, validatedUserName } = await testAuth.step(
+        'should get user names',
+        async () => {
+          const userName = userConfigs[userPrefix].username
+          const validatedUserName = userConfigs[userValidatedPrefix].username
```
hallieswan, line 36: Get usernames from config file, rather than looking up in local storage

e2e/teams.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+
+      await testAuth.step(
+        'user should invite validated user to team',
+        async () => {
+          await userPage.getByRole('button', { name: 'Team Actions' }).click()
+          await userPage.getByRole('link', { name: 'Invite User' }).click()
+
+          await userPage
+            .getByRole('textbox', { name: 'Enter a name or email address...' })
+            .pressSequentially(validatedUserName)
```
hallieswan, line 72: `locator.type` was deprecated in favor of `locator.fill` in [Playwright v1.38.0](https://github.com/microsoft/playwright/releases/tag/v1.38.0). However, user search box requires only finds username when entered letter by letter, so use `locator.pressSequentially` instead.

package.json
```diff
@@ -24,8 +24,9 @@
     "papaparse": "^5.4.1"
   },
   "devDependencies": {
-    "@playwright/test": "1.37.0",
+    "@playwright/test": "1.38.0",
     "@prettier/plugin-xml": "^3.1.0",
+    "@types/node": "^20.6.0",
```
hallieswan: I needed to add node types after upgrading Playwright to 1.38.0

playwright.config.ts
```diff
@@ -49,46 +45,23 @@ export default defineConfig({
 
   /* Configure projects for major browsers */
   projects: [
-    // Setup project
-    { name: 'setup', testMatch: /.*\.setup\.ts/, teardown: 'cleanup' },
```
hallieswan, line 53: No longer use setup project, instead create test users in fixtures

playwright.config.ts
```diff
@@ -49,46 +45,23 @@ export default defineConfig({
...
-    // Tests that require authentication
     {
-      name: 'chromium - logged in',
-      use: {
-        ...devices['Desktop Chrome'],
-        // Use prepared auth state.
-        storageState: USER_STORAGE_STATE,
-      },
-      testMatch: /.*\.loggedin\.spec.ts/,
-      dependencies: ['setup'],
```
hallieswan, line 73: No longer need to separate tests by authentication state

playwright.config.ts
```diff
@@ -49,46 +45,23 @@ export default defineConfig({
...
+      use: { ...devices['Desktop Chrome'] },
+      testMatch: /.*\.spec.ts/,
     },
 
-    // Clean up project
     {
-      name: 'cleanup',
-      use: { storageState: USER_STORAGE_STATE },
-      testMatch: /.*\.cleanup\.ts/,
-    },
```
hallieswan, line 81: Fixtures clean up after themselves

playwright.config.ts
```diff
@@ -49,46 +45,23 @@ export default defineConfig({
...
-    /* {
       name: 'firefox',
       use: { ...devices['Desktop Firefox'] },
+      testMatch: /.*\.spec.ts/,
     },
 
     {
       name: 'webkit',
       use: { ...devices['Desktop Safari'] },
-    }, */
```
hallieswan, line 91: Run tests against firefox and webkit

package.json
nickgros: In pom.xml we configure to download Node 18 at build-time: https://github.com/Sage-Bionetworks/SynapseWebClient/blob/73d467eea7cecfba65b3a65a057235919da13dc1/pom.xml#L361

Recommend using 18 (LTS) if we can. If not, change pom.xml to use v20

e2e/fixtures/authenticatedUserPages.ts
nickgros: So if/when we can avoid the CJS/ESM issue and import an API client, we can get rid of this dependency on a `page` in the `testAuth` fixture, because we won't need to wait for the `window.SRC` object to be loaded?

e2e/homepage.spec.ts
nickgros: Just to make sure I understand, you mean that we could just use the `page` fixture from `testAuth`  for the unauthenticated tests?

e2e/projects.spec.ts
nickgros: Hmm...which do you think is more intuitive behavior, this, or serial-by-default and opt-in to parallel?

e2e/teams.spec.ts
nickgros: Wonder if this is a bug in UserSearchBox

nickgros (approved):
Looks great to me...still wrapping my head around fixtures, but it definitely looks like an improvement over splitting tests by file!

package.json
hallieswan: Oh good call! I'll switch this to use 18.6

e2e/teams.spec.ts
hallieswan: I think the `InviteWidget` is using the `SynapseSuggestBox`, which seems to require a `keydown` event to trigger the users dropdown list -- [`locator.fill`](https://playwright.dev/docs/api/class-locator#locator-fill) only sends an `input` event after filling the input, so the dropdown isn't triggered. However, [`locator.pressSequentially`](https://playwright.dev/docs/api/class-locator#locator-press-sequentially) sends `keydown`, `keypress`/`input`, and `keyup` events for each character in the text, so the dropdown appears.

I'm not sure whether this is an issue that a real user would run into. Should I open a ticket to fix in `SynapseSuggestBox` or consider switching to `UserSearchBoxV2` (which doesn't seem to have the same issue)?

e2e/projects.spec.ts
hallieswan: Good question! I'm not sure -- is there a common default behavior in other test runners? I'll also pose the question to the team and see what everyone thinks!

e2e/homepage.spec.ts
hallieswan: Yeah, exactly -- since `page` is a default test fixture, and we haven't overridden it in `testAuth`, we could still use `page` from `testAuth` in unauthenticated tests. 

Also, since the worker fixtures are setup lazily for the first test that needs them, the unauthenticated tests wouldn't have to wait for the worker fixtures to be setup before running. 

I was mostly using this to organize tests (probably as a holdover from creating separate `loggedin`/`loggedout` test files), but if not's useful/helpful, we don't have to separate tests this way!

e2e/fixtures/authenticatedUserPages.ts
hallieswan: Yes, the fixtures that only make SRC API calls would be able to stop creating new pages, such as the `createUsers` worker fixture, which creates and deletes test users by making API calls via SRC's HttpClient.

However, we'd still need a `Page` to set up other fixtures -- for example, the new test fixtures (`userPage`, `validatedUserPage`) return a `Page` with a storage state for an authenticated user. Similarly, one of the new worker fixtures (`storageStatePaths`) uses a page to authenticate each user through the UI.

So `testAuth` (our [extension](https://playwright.dev/docs/api/class-test#test-extend) of `test`) would still need to create new `Page` objects, even after we can call the API from node.

e2e/teams.spec.ts
nickgros: It seems to not work in Firefox when I paste using the right-click context menu (pasting using keyboard shortcut works normally). This is probably not a huge deal, but it would be a good idea to create a ticket with these details so we can decide if it's worth the effort later.

e2e/projects.spec.ts
nickgros: I guess since `serial` is not recommended, let's go with parallel by default as you've set it up already

e2e/teams.spec.ts
hallieswan: Good call! Issue created [here](https://sagebionetworks.jira.com/browse/SWC-6569).

package.json
```diff
@@ -38,8 +38,12 @@
     "*.{*}": "prettier --write"
   },
   "scripts": {
+    "build": "mvn -B package --file pom.xml",
+    "docker:start": "docker run --name swc-tomcat -d --rm -p 8888:8080 -v \"/$(pwd)/target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war\" -v \"/$(pwd)/e2e_workflow/settings.xml\":/root/.m2/settings.xml tomcat:9.0",
+    "docker:stop": "docker stop swc-tomcat",
```
hallieswan, line 43: Convenience scripts for local e2e tests

playwright.config.ts
```diff
@@ -29,7 +29,7 @@ export default defineConfig({
   /* Retries */
   retries: process.env.CI ? 2 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  workers: process.env.CI ? 1 : 2,
```
hallieswan, line 32: Specify a reasonable number of workers for local environment

package.json
```diff
@@ -26,7 +26,7 @@
   "devDependencies": {
     "@playwright/test": "1.38.0",
     "@prettier/plugin-xml": "^3.1.0",
-    "@types/node": "^20.6.0",
+    "@types/node": "18.6.5",
```
hallieswan, line 29: Use the correct version of `@types/node` for the node version specified in `pom.xml`

nickgros (approved):

PR #5191, opened 2023-09-22
SWC-6572

jay-hodgson:


jay-hodgson:
Couple of targeted css fixes.
Results in the following output:
Screen Shot 2023-09-21 at 5 08 39 PM
Screen Shot 2023-09-21 at 5 15 27 PM

src/main/webapp/sass/_core.scss
```diff
@@ -3299,6 +3304,10 @@ div.border {
   transition: opacity 0.3s ease-in;
 }
 
+#body .SRC-grid-item .UserBadge {
+  width: 100%;
+}
```
jay-hodgson, line 3309: Let the grid item parent determine the width.

src/main/webapp/sass/_core.scss
```diff
@@ -2592,9 +2600,6 @@ Looks for any child element under .modal-fullscreen that has a class that contai
 
 #body .inline-user-badge {
   display: inline-flex;
-  div {
-    display: inline;
```
jay-hodgson, line 2596: display inline-flex

src/main/webapp/sass/_core.scss
```diff
@@ -2592,9 +2600,6 @@ Looks for any child element under .modal-fullscreen that has a class that contai
...
-  }
   .SRC-userImgSmall,
   .SRC-userImgMedium,
   .SRC-userImgLarge,
```
nickgros, line 2605: These classes are no longer used

nickgros (commented):
Thanks @jay-hodgson !

PR #5199, opened 2023-10-10
SWC-6577: after a new website version is detected, automatically relo…

jay-hodgson:
…ad the site after 30 minutes


nickgros (approved):
2 questions:
1. Out of curiosity, what did loading the iframe accomplish before this change?
2. Will this reload get blocked by an open editor (i.e. GlobalApplicationState.isEditing)?

jay-hodgson:
> 1. Out of curiosity, what did loading the iframe accomplish before this change?

Loading the new code in an iframe preloaded the new web app assets.

> 2. Will this reload get blocked by an open editor (i.e. GlobalApplicationState.isEditing)?

Yes, I think it will!  And if that editor has some UI driven event loop that calls remote services, then SWC-6577 could still occur.  But this change should drastically reduce the space where this bug lives.

PR #5200, opened 2023-10-13
SWC-6555

hallieswan:
- add end-to-end test for File sharing 
  - certify test users by default
  - add methods for uploading/deleting a file and creating/deleting projects
- bump playwright version
- update CI workflow to build SWC once, then shard tests per browser across different machines

### DIFF
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -1,9 +1,27 @@
 name: 'Build/Run End-to-End Tests'
 
 on: push
+env:
+  BUILD_ARTIFACT_NAME: swc-war
+  BUILD_DIR: target
+  BUILD_NAME: portal-develop-SNAPSHOT.war
+  PW_ALL_BLOBS_DIR: all-blob-reports
 jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build SWC
+        uses: ./.github/workflows/build
+      - name: Upload build to GitHub Actions Artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.BUILD_ARTIFACT_NAME }}
+          path: '${{ env.BUILD_DIR }}/${{ env.BUILD_NAME }}'
+          retention-days: 1
   playwright-tests:
     if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
+    needs: [build]
     runs-on: ${{ github.repository_owner == 'Sage-Bionetworks' && 'ubuntu-22.04-4core-16GBRAM-150GBSSD' || 'macos-latest' }}
     timeout-minutes: 60
     strategy:
@@ -20,17 +38,22 @@ jobs:
         run: |
           brew install docker
           colima start
-      - name: Build SWC
-        uses: ./.github/workflows/build
+      - name: Download build from GitHub Actions Artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ env.BUILD_ARTIFACT_NAME }}
+          path: ${{ env.BUILD_DIR }}
       - name: Run SWC on Tomcat
         run: |
           docker run \
-            --name "${CONTAINER_NAME}" \
+            --name "${{ env.CONTAINER_NAME }}" \
             -d --rm \
             -p 8888:8080 \
-            -v "/$(pwd)/target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war" \
+            -v "/$(pwd)/${{ env.BUILD_DIR }}/${{ env.BUILD_NAME }}:/usr/local/tomcat/webapps/ROOT.war" \
             -v "/$(pwd)/e2e_workflow/settings.xml":/root/.m2/settings.xml \
             tomcat:9.0
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
       - name: Install Playwright Browsers
         run: yarn playwright install --with-deps
       - name: Run Playwright tests
@@ -41,12 +64,12 @@ jobs:
         uses: actions/upload-artifact@v3
         if: always()
         with:
-          name: all-blob-reports
+          name: ${{ env.PW_ALL_BLOBS_DIR }}
           path: blob-report
           retention-days: 1
       - name: Stop Tomcat
         if: always()
-        run: docker stop "${CONTAINER_NAME}"
+        run: docker stop "${{ env.CONTAINER_NAME }}"
       - name: Stop Colima
         if: runner.os == 'macos'
         run: colima stop
@@ -65,10 +88,10 @@ jobs:
       - name: Download blob reports from GitHub Actions Artifacts
         uses: actions/download-artifact@v3
         with:
-          name: all-blob-reports
-          path: all-blob-reports
+          name: ${{ env.PW_ALL_BLOBS_DIR }}
+          path: ${{ env.PW_ALL_BLOBS_DIR }}
       - name: Merge into HTML Report
-        run: yarn playwright merge-reports --reporter html ./all-blob-reports
+        run: yarn playwright merge-reports --reporter html ./"${{ env.PW_ALL_BLOBS_DIR }}"
       - name: Upload HTML report
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -5,6 +5,7 @@ env:
   BUILD_ARTIFACT_NAME: swc-war
   BUILD_DIR: target
   BUILD_NAME: portal-develop-SNAPSHOT.war
+  NODE_VERSION: 18.16.0
   PW_ALL_BLOBS_DIR: all-blob-reports
 jobs:
   build:
@@ -52,6 +53,9 @@ jobs:
             -v "/$(pwd)/${{ env.BUILD_DIR }}/${{ env.BUILD_NAME }}:/usr/local/tomcat/webapps/ROOT.war" \
             -v "/$(pwd)/e2e_workflow/settings.xml":/root/.m2/settings.xml \
             tomcat:9.0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
       - name: Install dependencies
         run: yarn install --frozen-lockfile
       - name: Install Playwright Browsers
@@ -82,7 +86,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: ${{ env.NODE_VERSION }}
       - name: Install dependencies
         run: yarn install --frozen-lockfile
       - name: Download blob reports from GitHub Actions Artifacts
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -14,13 +14,6 @@ jobs:
       - name: Setup docker (missing on MacOS)
         if: runner.os == 'macos'
         run: |
-          # workaround for https://github.com/actions/runner-images/issues/8104
-          # ...and handle 8.1.0 regression: https://github.com/Homebrew/homebrew-core/issues/140244
-          export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
-          brew uninstall --ignore-dependencies qemu
-          curl -OSL https://raw.githubusercontent.com/Homebrew/homebrew-core/dc0669eca9479e9eeb495397ba3a7480aaa45c2e/Formula/qemu.rb
-          brew install ./qemu.rb
-
           brew install docker
           colima start
       - name: Build SWC
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -2,10 +2,14 @@ name: 'Build/Run End-to-End Tests'
 
 on: push
 jobs:
-  build-and-run-e2e:
+  playwright-tests:
     if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
     runs-on: ${{ github.repository_owner == 'Sage-Bionetworks' && 'ubuntu-22.04-4core-16GBRAM-150GBSSD' || 'macos-latest' }}
     timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1/3, 2/3, 3/3]
     env:
       CONTAINER_NAME: 'swc-tomcat'
       CI: true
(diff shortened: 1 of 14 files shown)


.github/workflows/build-test-e2e.yml
```diff
@@ -1,11 +1,34 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build SWC
+        uses: ./.github/workflows/build
+      - name: Upload build to GitHub Actions Artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.BUILD_ARTIFACT_NAME }}
+          path: '${{ env.BUILD_DIR }}/${{ env.BUILD_NAME }}'
+          retention-days: 1
```
hallieswan, line 22: Build SWC once, then upload WAR as an artifact

.github/workflows/build-test-e2e.yml
```diff
@@ -1,11 +1,34 @@
...
+  playwright-tests:
     if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
+    needs: [build]
     runs-on: ${{ github.repository_owner == 'Sage-Bionetworks' && 'ubuntu-22.04-4core-16GBRAM-150GBSSD' || 'macos-latest' }}
     timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1/3, 2/3, 3/3]
```
hallieswan, line 31: Shard e2e tests for each browser across machines. Set [`fail-fast`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) to false, so that all shards will finish running tests, even if tests on one shard fail.

.github/workflows/build-test-e2e.yml
```diff
@@ -14,42 +37,68 @@ jobs:
       - name: Setup docker (missing on MacOS)
         if: runner.os == 'macos'
         run: |
-          # workaround for https://github.com/actions/runner-images/issues/8104
-          # ...and handle 8.1.0 regression: https://github.com/Homebrew/homebrew-core/issues/140244
-          export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=1
-          brew uninstall --ignore-dependencies qemu
-          curl -OSL https://raw.githubusercontent.com/Homebrew/homebrew-core/dc0669eca9479e9eeb495397ba3a7480aaa45c2e/Formula/qemu.rb
-          brew install ./qemu.rb
-
```
hallieswan, line 23: Workaround no longer needed -- qemu has been updated in the GitHub runners

.github/workflows/build-test-e2e.yml
hallieswan: Specify the full file path, so that other files in the `target` directory are not included in the artifact

.github/workflows/build-test-e2e.yml
```diff
@@ -14,42 +37,68 @@ jobs:
...
+        run: yarn install --frozen-lockfile
       - name: Install Playwright Browsers
         run: yarn playwright install --with-deps
       - name: Run Playwright tests
         env:
           ADMIN_PAT: ${{ secrets.ADMIN_PAT }}
-        run: yarn playwright test
-      - name: Upload Playwright Report
+        run: yarn playwright test --shard ${{ matrix.shard }}
+      - name: Upload blob report to GitHub Actions Artifacts
```
hallieswan, line 67: Each shard will produce its own blob report, so upload these as an artifact

.github/workflows/build-test-e2e.yml
```diff
@@ -14,42 +37,68 @@ jobs:
...
+          node-version: ${{ env.NODE_VERSION }}
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Download blob reports from GitHub Actions Artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ env.PW_ALL_BLOBS_DIR }}
+          path: ${{ env.PW_ALL_BLOBS_DIR }}
+      - name: Merge into HTML Report
+        run: yarn playwright merge-reports --reporter html ./"${{ env.PW_ALL_BLOBS_DIR }}"
```
hallieswan, line 98: Combine the blob reports created by each shard into a single HTML report

e2e/files.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+    const userPage = await userContext.newPage()
+    const userAccessToken = await getAccessTokenFromCookie(userPage)
+
+    // create project
+    const userProjectName = generateEntityName('project')
+    const userProjectId = await createProject(
+      userProjectName,
+      userAccessToken,
+      userPage,
+    )
```
hallieswan, line 122: Create the Project with the test user's access token so that the project is initially only accessible by the test user

e2e/files.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+    async ({ userPage, validatedUserPage, browserName }, testInfo) => {
+      if (browserName === 'webkit') {
+        test.info().annotations.push({
+          type: 'very slow',
+          description: `webkit in CI only. May be related to the InviteWidget. 
+            Evaluate if timeout can be removed after addressing 
+            https://sagebionetworks.jira.com/browse/SWC-6569.`,
+        })
+        test.setTimeout(testInfo.timeout * 5)
+      }
```
hallieswan, line 157: The files test is intermittently very slow in webkit in CI (5-7 minutes in CI vs 35 seconds locally), but does eventually pass. Increase overall test timeout and timeout of particularly slow expectations to help this test pass when it is slow.

e2e/files.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+      )
+
+      await testAuth.step(
+        'First user shares the file with second user',
+        async () => {
+          const userSuggestBox = userPage.getByRole('textbox', {
+            name: 'Enter name...',
+          })
+          await userSuggestBox.fill(validatedUserName)
+          await userSuggestBox.press('Shift')
```
hallieswan, line 220: `press` and `pressSequentially` are particularly slow in webkit in CI, so fill the validated username to avoid calling `press` for each letter of the  validated user's name, then `press` Shift to trigger the user suggestion dropdown, as discussed in [SWC-6569](https://sagebionetworks.jira.com/browse/SWC-6569)

e2e/fixtures/authenticatedUserPages.ts
```diff
@@ -47,8 +48,8 @@ function createUserPageFixture<Page>(userPrefix: UserPrefixes) {
     const context = await browser.newContext({
       storageState: storageStatePaths[userPrefix],
     })
-    const validatedUserPage = await context.newPage()
-    await use(validatedUserPage)
+    const userPage = await context.newPage()
+    await use(userPage)
```
hallieswan: General method, so use a general variable name

e2e/fixtures/authenticatedUserPages.ts
```diff
@@ -68,10 +69,18 @@ export const testAuth = base.extend<
...
+          async () => {
+            const userId = await createTestUser(user, getAdminPAT(), page)
+            expect(userId).not.toBeUndefined()
+            userIds.push(userId)
+            return userId
+          },
+        )
+
+        await base.step(`certify test user - ${user.username}`, async () => {
+          await setCertifiedUserStatus(userId, true, getAdminPAT(), page)
```
hallieswan, line 90: Users must be certified to upload a file -- certify new users by default, similar to [backend integration tests](https://github.com/Sage-Bionetworks/Synapse-Repository-Services/blob/e84f3c2556c4458348c27cd11edfad82ceed788e/integration-test/src/test/java/org/sagebionetworks/SynapseClientHelper.java#L69C18-L69C18)

e2e/helpers/entities.ts
```diff
@@ -0,0 +1,203 @@
...
+  page: Page,
+) {
+  await doDelete(
+    page,
+    `/file/v1/fileHandle/${handleId}`,
+    accessToken,
+    BackendDestinationEnum.REPO_ENDPOINT,
+  )
+  return handleId
+}
```
hallieswan, line 175: Moved from e2e/helpers/messages.ts so these methods are colocated with other entity methods

e2e/helpers/messages.ts
```diff
@@ -43,38 +39,6 @@ async function deleteUserMessage(
...
-  page: Page,
-) {
-  await doDelete(
-    page,
-    `/file/v1/fileHandle/${handleId}`,
-    accessToken,
-    BackendDestinationEnum.REPO_ENDPOINT,
-  )
-  return handleId
-}
```
hallieswan, line 76: Moved into e2e/helpers/entities.ts

e2e/helpers/testUser.ts
```diff
@@ -84,9 +84,6 @@ export async function loginTestUser(
   await page.getByRole('button', { name: 'Sign in' }).click()
 
   // Ensure that correct username/password were received
-  const loadingButton = page.getByRole('button', { name: 'Logging you in' })
-  await expect(loadingButton).toBeVisible()
-  await expect(loadingButton).not.toBeVisible()
```
hallieswan, line 89: This state is very short-lived, so was intermittently missed and caused tests to fail, even when wrapped in a Promise.all. This expectation was originally added to help us understand why login intermittently failed, but removing it since actually caused tests to be flakier and the check for the URL and the username/password alert should help us evaluate, if needed

package.json
```diff
@@ -46,6 +46,8 @@
     "e2e:chromium": "yarn playwright test --project=chromium",
     "e2e:ui": "yarn playwright test --ui",
     "e2e:codegen": "yarn playwright codegen",
-    "e2e:report": "yarn playwright show-report"
+    "e2e:report": "yarn playwright show-report",
+    "e2e:report:blob": "yarn playwright merge-reports --reporter html ./blob-report && yarn e2e:report",
+    "playwright:update": "yarn add -D @playwright/test@latest; yarn playwright install"
```
hallieswan, line 51: Update Playwright browsers after updating the Playwright version

playwright.config.ts
```diff
@@ -32,7 +32,7 @@ export default defineConfig({
   workers: process.env.CI ? 1 : 2,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
   // Concise 'dot' for CI, default 'list' when running locally
-  reporter: process.env.CI ? [['list'], ['html']] : 'html',
+  reporter: process.env.CI ? [['list'], ['blob']] : 'html',
```
hallieswan, line 35: Use 'blob' format in CI, so that the report from each shard can be merged into a single HTML report

.github/workflows/build-test-e2e.yml
nickgros: This is unique per-workflow, right? i.e. if you have 2 PRs running this job, could you have this race condition? 

1. Job 1 builds + uploads WAR artifact 1
2. Job 2 builds + uploads WAR artifact 2
3. Job 1 shard(s) downloads the artifact, which ends up incorrectly being WAR artifact 2

.github/workflows/build-test-e2e.yml
nickgros: Just curious, is it smart enough to shard by browser engine?

e2e/files.spec.ts
nickgros: That's really weird. This annotations system is cool though, so thanks for adding all of this potentially helpful information.

e2e/files.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+          await expect(
+            userPage.getByRole('cell', { name: validatedUserName }),
+          ).toBeVisible({ timeout: testInfo.timeout * 3 })
+
+          // Don't send message, so we don't have to clean up the message
+          await userPage.getByText('Notify people via email').click()
+
+          // Cookies banner covers the button to save settings
+          // ...so close banner by accepting cookies
+          await acceptSiteCookies(userPage)
```
nickgros: Would it make sense to move this into a fixture, and just auto-accept for every new page (user?) instance? I don't think we should have to think about this in this test.

e2e/files.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+      await testAuth.step('First user can view a private file', async () => {
+        await userPage.goto(`${entityUrlPathname(userProject.id)}/files`)
+        await waitForInitialPageLoad(userPage)
+
+        const fileLink = userPage.getByRole('link', { name: fileName })
+        await expect(fileLink).toBeVisible()
+
+        await fileLink.click()
+        await expect(
+          userPage.getByText(`Discussion about ${fileName}`),
```
nickgros: Just curious, why look for "Discussion about" ?

e2e/files.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+      await openFileSharingSettings(userPage)
+      await confirmSharingSettings(userPage, userName, validatedUserName, false)
+
+      await testAuth.step('Second user cannot access the file', async () => {
+        await validatedUserPage.reload()
+        await expectNoAccessPage(validatedUserPage)
+      })
+    },
+  )
+})
```
nickgros, line 283: This test suite is exceptionally readable!!

nickgros (approved):
This looks great, though I had a few questions

.github/workflows/build-test-e2e.yml
hallieswan: That's a good question! I think artifacts are unique per workflow run (e.g. [each e2e workflow run](https://github.com/Sage-Bionetworks/SynapseWebClient/actions/workflows/build-test-e2e.yml) has its own 'playwright-report' artifact), but I couldn't find something in the docs that explicitly states this. 

Either way, I don't think we'd run into that race condition because per the [GH docs](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#about-workflow-artifacts), "You can only download artifacts that were uploaded during the same workflow run." So the Job 1 shard should only be able to download WAR artifact 1, even if another WAR artifact exists.

.github/workflows/build-test-e2e.yml
hallieswan: Good question! I don't see their sharding strategy explicitly defined in the docs, but from my tests, sharding is done by test, where tests are ordered by the project declaration order in the config file, then the alphabetized file name, then the test declaration order within each file. The tests are then distributed as evenly as possible across shards. In a simple example, running three projects (10 chromium tests, 10 firefox tests, and 5 webkit tests) on three shards results in shard 1 running 9 tests (all chromium), shard 2 running 8 tests (1 chromium, 7 firefox), and shard 3 running 8 tests (3 firefox, 5 webkit).

Since we have one project per browser that runs all tests, using an equal number of shards as browsers ensures that each shard only runs tests for one browser. However, if we start writing browser-specific tests or change how the projects are defined, we may need a more complicated strategy to ensure that each shard runs one browser (or decide whether that's necessary).

I've seen issues requesting other sharding strategies (e.g. sharding by test time), so it's also possible that more sophisticated strategies will be available in the future.

e2e/files.spec.ts
hallieswan: Ah, good question -- this was an easy locator to confirm that the File page had loaded, but I'll add some other expectations about the File page for completeness!

e2e/files.spec.ts
hallieswan: Sure, I'll add it in to the user page fixtures!

PR #5202, opened 2023-10-16
SWC-6555: fix

hallieswan:


e2e/helpers/testUser.ts
```diff
@@ -69,6 +69,10 @@ export async function loginTestUser(
   await page.goto('/')
   await waitForInitialPageLoad(page)
 
+  // Accept cookies, so banner doesn't obscure buttons in other tests
+  await acceptSiteCookies(page)
```
hallieswan, line 73: Accept site cookies during login, which fixes issue in Firefox where the user was logged out when using authenticated user page fixture.

nickgros (approved):

PR #5203, opened 2023-10-16
SWC-6558: add end-to-end test for Favorites workflow

hallieswan:


e2e/favorites.spec.ts
```diff
@@ -1,13 +1,120 @@
...
 import { testAuth } from './fixtures/authenticatedUserPages'
+import { entityUrlPathname } from './helpers/entities'
 import { goToDashboard } from './helpers/testUser'
+import { Project } from './helpers/types'
+import { waitForInitialPageLoad } from './helpers/utils'
+
+const PUBLIC_PROJECT: Project = {
+  name: 'swc-e2e-test-project',
+  id: 'syn12366772',
+}
```
hallieswan, line 13: For now, I set up a public project owned by a `swc-e2e-admin`. In the future, we could create/delete a new public project within the test instead.

e2e/favorites.spec.ts
```diff
@@ -1,13 +1,120 @@
...
+  page: Page,
+  projectName: string,
+  projectId: string,
+) {
+  await page.waitForURL(entityUrlPathname(projectId))
+  await expect(page.getByRole('heading', { name: projectName })).toBeVisible()
+  await page.waitForURL(
+    `${entityUrlPathname(PUBLIC_PROJECT.id)}/discussion/default`,
+  )
+  await expect(page.getByRole('heading', { name: 'Discussion' })).toBeVisible()
```
hallieswan, line 37: Wait for navigations to finish and project page to load -- after navigating to the base project page, the user is redirected to `/wiki/` and then `/discussion/default`.

e2e/favorites.spec.ts
```diff
@@ -1,13 +1,120 @@
...
+
+        await expect(addFavorite).toBeVisible()
+        await expect(removeFavorite).not.toBeVisible()
+      })
+
+      await testAuth.step(
+        'User does not have project listed on their favorites page',
+        async () => {
+          await goToFavorites(userPage)
+          await userPage.reload()
```
hallieswan, line 116: The page needs to be reloaded for the favorites list to update

nickgros: Can you add as a comment right above this line?

```suggestion
          // The page needs to be reloaded for the favorites list to update
          await userPage.reload()
```


At some point this might use a cache like react-query, where the reload won't be necessary--if/when that affects this integration test, the comment could help us find that and decide if it can be removed.

e2e/favorites.spec.ts
nickgros: Please add a comment so that way we know why this has started failing if this project gets deleted

nickgros (approved):
Looks great! Just want a few notes moved into code comments

PR #5215, opened 2023-10-25
SWC-6560: discussions e2e test

hallieswan:
- Add e2e test for discussions
- Pull project setup / teardown into reusable methods
- Update e2e README to reflect current process

### DIFF
--- a/e2e/discussions.spec.ts
+++ b/e2e/discussions.spec.ts
@@ -1,0 +1,465 @@
+import { Page, expect, test } from '@playwright/test'
+import { testAuth } from './fixtures/authenticatedUserPages'
+import { entityUrlPathname } from './helpers/entities'
+import {
+  setupProjectWithPermissions,
+  teardownProjectsAndFileHandles,
+} from './helpers/setupTeardown'
+import { dismissAlert } from './helpers/testUser'
+import { Project } from './helpers/types'
+import { UserPrefixes, userConfigs } from './helpers/userConfig'
+
+const discussionActionIconClasses = {
+  PIN: '.fa-thumb-tack.imageButton',
+  EDIT: '.fa-pencil.imageButton',
+  DELETE: '.syn-trash-o.imageButton',
+  TOGGLE_FOLLOW: '.fa-eye.imageButton',
+  LINK: '.fa-link.imageButton',
+}
+const discussionThreadSelector = '.discussionThread:visible'
+const discussionReplySelector = '.discussionReply:visible'
+
+const getDefaultDiscussionPath = (projectId: string) => {
+  return `${entityUrlPathname(projectId)}/discussion/default`
+}
+
+const getThreadTextbox = (page: Page) => {
+  return page.locator('.markdownEditor').getByRole('textbox')
+}
+
+const expectDiscussionPageLoaded = async (page: Page, projectId: string) => {
+  await testAuth.step('Default discussion page has loaded', async () => {
+    await page.waitForURL(getDefaultDiscussionPath(projectId))
+    await expect(
+      page.getByRole('heading', { name: 'Discussion' }),
+    ).toBeVisible()
+    await expect(page.getByRole('button', { name: 'New Thread' })).toBeVisible()
+    await expect(
+      page.getByRole('button', { name: 'Discussion Tools' }),
+    ).toBeVisible()
+  })
+}
+
+const expectThreadTableLoaded = async (
+  page: Page,
+  title: string,
+  author: string,
+  replies: string,
+  views: string,
+) => {
+  await testAuth.step('Thread table info is loaded', async () => {
+    const threadCells = page
+      .getByRole('row')
+      .filter({ hasText: title })
+      .getByRole('cell')
+
+    await expect(threadCells.first(), 'Should have correct title').toHaveText(
+      title,
+    )
+    await expect(threadCells.nth(1), 'Should have correct author').toHaveText(
+      `@${author}`,
+    )
+    await expect(threadCells.nth(3), 'Should have correct replies').toHaveText(
+      replies,
+    )
+    await expect(threadCells.nth(4), 'Should have correct views').toHaveText(
+      views,
+    )
+  })
+}
+
+const expectDiscussionThreadLoaded = async (page: Page, threadId: number) => {
+  await testAuth.step('Discussion thread has loaded', async () => {
+    await page.waitForURL(
+      `${entityUrlPathname(userProject.id)}/discussion/threadId=${threadId}`,
+    )
+    await expect(
+      page.getByRole('button', { name: 'Show All Threads' }),
+    ).toBeVisible()
+    await expect(
+      page.getByRole('button', { name: 'Date Posted' }),
+    ).toBeVisible()
+    await expect(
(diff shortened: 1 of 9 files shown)


e2e/files.spec.ts
```diff
@@ -167,40 +168,17 @@ let fileHandleIds: string[] = []
...
-    const userProjectName = generateEntityName('project')
-    const userProjectId = await createProject(
-      userProjectName,
-      userAccessToken,
-      userPage,
-    )
-    userProject = { name: userProjectName, id: userProjectId }
-
-    await userContext.close()
+    userProject = await setupProject(browser, 'swc-e2e-user', storageStatePaths)
```
hallieswan, line 171: Pull project setup into a method that can be reused across tests

e2e/helpers/ACCESS_TYPE.ts
```diff
@@ -0,0 +1,28 @@
+/**
```
hallieswan, line 1: Duplicates synapse-type's [ACCESS_TYPE.ts](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-types/src/ACCESS_TYPE.ts) -- can be dropped when SWC-6529 is complete.

e2e/helpers/setupTeardown.ts
```diff
@@ -0,0 +1,83 @@
...
+  const accessToken = await getAccessTokenFromCookie(page)
+
+  const projectName = generateEntityName('project')
+  const projectId = await createProject(projectName, accessToken, page)
+  await context.close()
+
+  return { name: projectName, id: projectId } as Project
+}
+
+export const setupProjectWithPermissions = async (
```
hallieswan, line 37: Create a new project and then update the ACL to give another user access.

e2e/helpers/setupTeardown.ts
```diff
@@ -0,0 +1,83 @@
...
+  )
+  await context.close()
+
+  return project
+}
+
+export const teardownProjectsAndFileHandles = async (
+  browser: Browser,
+  projects: Project[],
+  fileHandleIds: string[],
```
hallieswan, line 66: Handle multiple projects, so the test can create multiple projects in the `beforeAll` or within a test, then those projects can all be cleaned up in the `afterAll` using a single page.

e2e/helpers/types.ts
```diff
@@ -84,3 +86,19 @@ export type FileUploadComplete = {
...
+
+export type AccessControlList = {
+  id: string
+  createdBy?: string
+  creationDate?: string
+  modifiedBy?: string
+  modifiedOn?: string
+  etag?: string
+  resourceAccess: ResourceAccess[]
+}
```
hallieswan, line 104: Also duplicated from synapse-types: [AccessControlList.ts](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-types/src/AccessControlList.ts), [ResourceAccess.ts](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-types/src/ResourceAccess.ts)

e2e/discussions.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+import { Project } from './helpers/types'
+import { UserPrefixes, userConfigs } from './helpers/userConfig'
+
+const discussionActionIconClasses = {
+  PIN: '.fa-thumb-tack.imageButton',
+  EDIT: '.fa-pencil.imageButton',
+  DELETE: '.syn-trash-o.imageButton',
+  TOGGLE_FOLLOW: '.fa-eye.imageButton',
+  LINK: '.fa-link.imageButton',
+}
```
hallieswan, line 21: Discussion action icons are not buttons and don't have accessible names, so can't be selected by role + name or by label -- select by classname to differentiate between the icons.

e2e/discussions.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+const discussionThreadSelector = '.discussionThread:visible'
+const discussionReplySelector = '.discussionReply:visible'
```
hallieswan, line 23: Used to select the current thread (visible) instead of the deleted thread (hidden), since Playwright [css selectors](https://playwright.dev/docs/other-locators#css-matching-only-visible-elements) match all elements (hidden or visible).

e2e/discussions.spec.ts
```diff
@@ -0,0 +1,465 @@
...
+) => {
+  await testAuth.step('Thread table info is loaded', async () => {
+    const threadCells = page
+      .getByRole('row')
+      .filter({ hasText: title })
+      .getByRole('cell')
+
+    await expect(threadCells.first(), 'Should have correct title').toHaveText(
+      title,
+    )
```
hallieswan, line 61: Include a [custom expect message](https://playwright.dev/docs/test-assertions#custom-expect-message) to make it easier to understand a failure

e2e/discussions.spec.ts
nickgros: That's not good! This is fine for now. Really like that RTL/Playwright basically force you to do surface level a11y testing. Do you mind adding your note as a comment in a `// FIXME`?

e2e/discussions.spec.ts
nickgros: Nice!

e2e_workflow/README.md
nickgros, line 1: 🚀

e2e/helpers/setupTeardown.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  // delete projects
+  for await (const project of projects) {
+    await deleteProject(project.id, accessToken, page)
+  }
+
+  // delete fileHandles
+  for await (const fileHandleId of fileHandleIds) {
+    await deleteFileHandleWithRetry(accessToken, fileHandleId, page)
+  }
```
nickgros: I guess you already had `for await` here, but it's not making sense to me why you need it, it seems like `fileHandleId`/`fileHandleIds` are all available without `await`

e2e/helpers/setupTeardown.ts
```diff
@@ -0,0 +1,83 @@
...
+  getAccessTokenFromCookie,
+  getAdminPAT,
+  getUserIdFromLocalStorage,
+} from './testUser'
+import { Project } from './types'
+import { UserPrefixes } from './userConfig'
+
+export const setupProject = async (
+  browser: Browser,
+  projectCreator: UserPrefixes,
```
nickgros: Not super important but I think `UserPrefixes` should be renamed to `UserPrefix` since it only refers to one prefix

e2e/helpers/setupTeardown.ts
nickgros: Super slick

nickgros (approved):

e2e/helpers/setupTeardown.ts
hallieswan: Good point -- I agree the `await` doesn't seem necessary, I'll remove it!

e2e/discussions.spec.ts
hallieswan: Good call! Done in the next commit

nickgros (approved):

PR #5216, opened 2023-10-25
SWC-6556, SWC-6560

hallieswan:
- Conditionally skip entire job, rather than just the second step
-Attempt to reduce flakiness in file upload, teams, and discussions tests

### DIFF
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -25,7 +25,7 @@ jobs:
           retention-days: 1
   playwright-tests:
     needs: [build]
-    runs-on: 'macos-latest'
+    runs-on: ${{ github.repository_owner == 'Sage-Bionetworks' && 'ubuntu-22.04-4core-16GBRAM-150GBSSD' || 'macos-latest' }}
     timeout-minutes: 60
     strategy:
       fail-fast: false
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -9,6 +9,9 @@ env:
   PW_ALL_BLOBS_DIR: all-blob-reports
 jobs:
   build:
+    # Run in Sage repo on develop or release- branches
+    # and on all branches in user-owned forks
+    if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -21,9 +24,8 @@ jobs:
           path: '${{ env.BUILD_DIR }}/${{ env.BUILD_NAME }}'
           retention-days: 1
   playwright-tests:
-    if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
     needs: [build]
-    runs-on: ${{ github.repository_owner == 'Sage-Bionetworks' && 'ubuntu-22.04-4core-16GBRAM-150GBSSD' || 'macos-latest' }}
+    runs-on: 'macos-latest'
     timeout-minutes: 60
     strategy:
       fail-fast: false
@@ -79,7 +81,8 @@ jobs:
         run: colima stop
   merge-reports:
     # Merge reports after playwright-tests, even if some shards have failed
-    if: always()
+    # But skip this job if the previous job was cancelled
+    if: ${{ !cancelled() }}
     needs: [playwright-tests]
     runs-on: ubuntu-latest
     steps:
--- a/e2e/discussions.spec.ts
+++ b/e2e/discussions.spec.ts
@@ -72,7 +72,12 @@ const expectThreadTableLoaded = async (
   })
 }
 
-const expectDiscussionThreadLoaded = async (page: Page, threadId: number) => {
+const expectDiscussionThreadLoaded = async (
+  page: Page,
+  threadId: number,
+  threadTitle: string,
+  threadBody: string,
+) => {
   await testAuth.step('Discussion thread has loaded', async () => {
     await page.waitForURL(
       `${entityUrlPathname(userProject.id)}/discussion/threadId=${threadId}`,
@@ -90,6 +95,12 @@ const expectDiscussionThreadLoaded = async (page: Page, threadId: number) => {
     await expect(
       page.getByRole('button', { name: 'Most Recent' }),
     ).toBeVisible()
+
+    const discussionThread = page.locator(discussionThreadSelector)
+    await expect(discussionThread.getByText(threadTitle)).toBeVisible()
+    await expect(discussionThread.getByText(threadBody)).toBeVisible({
+      timeout: 60_000,
+    })
   })
 }
 
@@ -224,7 +235,12 @@ test.describe('Discussions', () => {
 
           await testAuth.step('Go to thread', async () => {
             await threadLink.click()
-            await expectDiscussionThreadLoaded(userPage, threadId)
+            await expectDiscussionThreadLoaded(
+              userPage,
+              threadId,
+              threadTitle,
+              threadBody,
+            )
           })
 
           await testAuth.step('Check thread info', async () => {
@@ -239,8 +255,6 @@ test.describe('Discussions', () => {
               }),
             ).toBeVisible()
             await expect(discussionThread.getByText('Moderator')).toBeVisible()
-            await expect(discussionThread.getByText(threadTitle)).toBeVisible()
-            await expect(discussionThread.getByText(threadBody)).toBeVisible()
           })
 
           return threadId
@@ -254,6 +268,7 @@ test.describe('Discussions', () => {
 
         await expectDiscussionPageLoaded(userPage, userProject.id)
 
+        // retry if the view count hasn't updated
         await expect(async () => {
           // reload is necessary for view count to update
           await userPage.reload()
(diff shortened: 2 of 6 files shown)


.github/workflows/build-test-e2e.yml
```diff
@@ -9,6 +9,9 @@ env:
   PW_ALL_BLOBS_DIR: all-blob-reports
 jobs:
   build:
+    # Run in Sage repo on develop or release- branches
+    # and on all branches in user-owned forks
+    if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
```
hallieswan, line 14: Move conditional to first job so that subsequent jobs are skipped when first job is skipped -- will properly skip the entire workflow for tagged branches, rather than running the first job, skipping the second, and then failing the last job, as in [475.0](https://github.com/Sage-Bionetworks/SynapseWebClient/actions/runs/6595041108)

.github/workflows/build-test-e2e.yml
```diff
@@ -21,9 +24,8 @@ jobs:
           path: '${{ env.BUILD_DIR }}/${{ env.BUILD_NAME }}'
           retention-days: 1
   playwright-tests:
-    if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
     needs: [build]
-    runs-on: ${{ github.repository_owner == 'Sage-Bionetworks' && 'ubuntu-22.04-4core-16GBRAM-150GBSSD' || 'macos-latest' }}
+    runs-on: 'macos-latest'
```
hallieswan: Always run playwright tests on a macos runner so environment is consistent between forks and in Sage repo

.github/workflows/build-test-e2e.yml
```diff
@@ -79,7 +81,8 @@ jobs:
         run: colima stop
   merge-reports:
     # Merge reports after playwright-tests, even if some shards have failed
-    if: always()
+    # But skip this job if the previous job was cancelled
+    if: ${{ !cancelled() }}
```
hallieswan, line 85: If playwright-tests is skipped, then skip this job too. Otherwise, run this job regardless of whether playwright-tests succeeds or fails

.github/workflows/build-test-e2e.yml
nickgros: Remind me, why do we only run on macos?

hallieswan: We used the macOS runner because it's more powerful than the free Linux runner. 

@nickgros noticed that there are [usage limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) on GitHub-hosted runners, which currently allow a maximum of 5 concurrent macOS jobs for a GitHub “Team” plan. Before switching over to the macOS runners, we would like to investigate the potential impact on the org's macOS runner resource availability -- tracking in SWC-6598.

For now, I'll change this PR to continue using the GitHub-hosted larger Linux runner.

nickgros (approved):

e2e/files.spec.ts
```diff
@@ -114,9 +114,19 @@ const uploadFile = async (
...
+    await expect(page.getByRole('button', { name: 'Browse...' })).toBeVisible()
   })
 
   await testAuth.step('choose file', async () => {
+    // Ensure that upload type is specified before attempting upload
+    // to prevent upload error: "Unsupported external upload type specified: undefined"
+    await expect(
+      page.getByText(/all uploaded files will be stored in synapsestorage/i),
+    ).toBeVisible()
+
```
hallieswan, line 129: Ensure that the upload type is specified in the upload modal before attempting upload, so that the upload type is specified -- the flakiness on the Sage larger Linux runner may be associated with this error. 

SWC-6556_uploadError

e2e/files.spec.ts
```diff
@@ -140,10 +150,7 @@ const uploadFile = async (
   await testAuth.step('ensure there was not an upload error', async () => {
     const uploadError = page.getByRole('heading', { name: 'Upload Error' })
     if (await uploadError.isVisible()) {
-      const uploadErrorText = await page
-        .getByRole('dialog')
-        .getByText('Unable to upload the file')
-        .textContent()
+      const uploadErrorText = await page.getByRole('dialog').textContent()
```
hallieswan, line 153: Output the upload error text for any type of upload error, not only errors associated with the MultipartUploader

e2e/discussions.spec.ts
```diff
@@ -71,20 +72,35 @@ const expectThreadTableLoaded = async (
...
       `${entityUrlPathname(userProject.id)}/discussion/threadId=${threadId}`,
     )
     await expect(
-      page.getByRole('button', { name: 'Show All Threads' }),
+      page.getByRole('heading', { name: 'Discussion' }),
     ).toBeVisible()
+
+    await expect(
+      page.getByRole('button', { name: /show all threads/i }),
+    ).toBeVisible({ timeout: 60_000 })
```
hallieswan, line 91: This button doesn't appear until the thread network response returns -- increase timeout to handle cases where the network connection is slow.

e2e/discussions.spec.ts
```diff
@@ -71,20 +72,35 @@ const expectThreadTableLoaded = async (
...
     ).toBeVisible()
     await expect(
       page.getByRole('button', { name: 'Most Recent' }),
     ).toBeVisible()
+
+    const discussionThread = page.locator(discussionThreadSelector)
+    await expect(discussionThread.getByText(threadTitle)).toBeVisible()
+    await expect(discussionThread.getByText(threadBody)).toBeVisible({
+      timeout: 60_000,
+    })
```
hallieswan, line 103: Wait for the thread title and body to be visible before considering the thread to be "loaded"

e2e/discussions.spec.ts
```diff
@@ -163,6 +179,8 @@ test.describe('Discussions', () => {
   testAuth(
     'should allow discussion and reply CRUD',
     async ({ userPage, validatedUserPage }) => {
+      test.slow()
```
hallieswan, line 182: The test intermittently times out, so give the test more time to finish, e.g. when network connection is slow

e2e/discussions.spec.ts
```diff
@@ -182,12 +200,18 @@ test.describe('Discussions', () => {
...
         await testAuth.step('Enter thread information', async () => {
-          await userPage.getByPlaceholder('Title').fill(threadTitle)
-          await getThreadTextbox(userPage).fill(threadBody)
+          const threadTitleInput = userPage.getByPlaceholder('Title')
+          await threadTitleInput.fill(threadTitle)
+          await expect(threadTitleInput).toHaveValue(threadTitle)
+
+          const threadTextbox = getThreadTextbox(userPage)
+          await threadTextbox.fill(threadBody)
+          await expect(threadTextbox).toHaveValue(threadBody)
```
hallieswan, line 209: Ensure that input values have been set before submitting the thread info

e2e/discussions.spec.ts
```diff
@@ -226,28 +255,32 @@ test.describe('Discussions', () => {
...
+          await userPage.reload()
+          await expectDiscussionPageLoaded(userPage, userProject.id)
+          await expectThreadTableLoaded(
+            userPage,
+            threadTitle,
+            userConfigs['swc-e2e-user'].username,
+            '0',
+            '1',
+          )
+        }).toPass()
```
hallieswan, line 283: The test intermittently failed because the view count wasn't updated, so retry the reload to give the view count more time to update

e2e/discussions.spec.ts
```diff
@@ -282,10 +320,18 @@ test.describe('Discussions', () => {
...
-              .getByRole('textbox', { name: 'Write a reply...' })
-              .click()
-            await getThreadTextbox(validatedUserPage).fill(threadReply)
+            const replyInput = discussionThread.getByRole('textbox', {
+              name: 'Write a reply...',
+            })
+
+            // A second reply input briefly appears when the network
+            // connection is slow. Wait for the second input to be hidden.
+            await expect(replyInput).toHaveCount(1)
```
hallieswan, line 329: Handle case where a second reply input briefly appears, as shown in Chrome when using "Slow 3G" network throttling --

https://github.com/Sage-Bionetworks/SynapseWebClient/assets/26949006/4063b7ae-c15a-43e0-a75d-0ba311cb11e3

e2e/teams.spec.ts
```diff
@@ -161,22 +167,26 @@ test.describe('Teams', () => {
...
-    await userPage.getByLabel('Teams').click()
-    await userPage.getByRole('link', { name: TEAM_NAME }).click()
-    await userPage.getByRole('button', { name: 'Team Actions' }).click()
-    await userPage.getByRole('link', { name: 'Delete Team' }).click()
-    await userPage.getByRole('button', { name: 'Delete' }).click()
-    await expect(userPage.getByText('Team successfully deleted')).toBeVisible()
-    await expect(
-      userPage.getByRole('link', { name: TEAM_NAME }),
-    ).not.toBeVisible()
-
```
hallieswan, line 179: This `afterAll` intermittently times out -- delete the team via the UI within the test itself, rather than in the `afterAll`. However, if the test fails before the team is deleted, then delete the team via the API in the `afterAll`, so that the users can still be cleaned up.

e2e/teams.spec.ts
```diff
@@ -29,6 +31,8 @@ test.describe('Teams', () => {
   testAuth(
     'should exercise team lifecycle',
     async ({ userPage, validatedUserPage }, testInfo) => {
+      test.slow()
```
hallieswan, line 34: Test intermittently times out, so mark as slow to increase test timeout

e2e/discussions.spec.ts
nickgros: I don't think the view count is that important, so if it keeps flaking, then we should remove those checks IMO

e2e/discussions.spec.ts
nickgros: Super weird case! Nice catch!!

nickgros (approved):

PR #5217, opened 2023-10-25
SWC-6587: attempt to fix intermittent issue with display column names

jay-hodgson:


jay-hodgson:
See https://sagebionetworks.jira.com/browse/SWC-6587?focusedCommentId=187693

nickgros (approved):
Nice sleuthing!

PR #5225, opened 2023-11-07
SWC-6606

jay-hodgson:


nickgros (approved):

nickgros:
Dropping the JSON dependency is being pulled from develop to staging with this PR but I don't think that's a big deal, just wanted to point that out.

PR #5229, opened 2023-11-14
SWC-6610 - Add ColumnModelsEditorV2Widget

nickgros:
- Add ColumnModelsEditorV2Widget which will be shown when editing table columns in experimental mode
 - Update ColumnModelsWidget and CreateTableViewWizardStep2 to utilize new editor
 - Add JSONEntityUtils.toJsInteropCompatibleObject to simplify converting a JSONEntity to a JsInterop-compatible native JavaScript object
 - Update JsInterop definitions for React to support handling component refs

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <webappDirectory>
       ${project.build.directory}/${project.build.finalName}
     </webappDirectory>
-    <synapse.version>462.0</synapse.version>
+    <synapse.version>475.0</synapse.version>
     <gwtVersion>2.10.0</gwtVersion>
     <org.springframework.version>5.3.28</org.springframework.version>
     <guiceVersion>3.0</guiceVersion>
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/AccessRequirementListProps.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/AccessRequirementListProps.java
@@ -5,7 +5,7 @@ import jsinterop.annotations.JsFunction;
 import jsinterop.annotations.JsOverlay;
 import jsinterop.annotations.JsPackage;
 import jsinterop.annotations.JsType;
-import org.sagebionetworks.schema.adapter.JSONObjectAdapter;
+import org.sagebionetworks.repo.model.AccessRequirement;
 
 @JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
 public class AccessRequirementListProps extends ReactComponentProps {
@@ -23,7 +23,7 @@ public class AccessRequirementListProps extends ReactComponentProps {
   @JsOverlay
   public static AccessRequirementListProps create(
     Callback onHide,
-    List<JSONObjectAdapter> accessRequirements,
+    List<AccessRequirement> accessRequirements,
     String entityId
   ) {
     AccessRequirementListProps props = new AccessRequirementListProps();
@@ -33,7 +33,7 @@ public class AccessRequirementListProps extends ReactComponentProps {
     props.accessRequirementFromProps =
       accessRequirements
         .stream()
-        .map(o -> JSON.parse(o.toJSONString()))
+        .map(JSONEntityUtils::toJsInteropCompatibleObject)
         .toArray();
     return props;
   }
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
@@ -25,6 +25,8 @@ public class React {
     ReactNode[] children
   );
 
+  public static native <T> T createRef();
+
   /**
    * Similar to {@link #createElementWithSynapseContext} but only includes the theme. Any components rendered will NOT get the full Synapse context, including
    * access token + auth state, experimental mode status, and time display settings.
@@ -36,10 +38,11 @@ public class React {
     ReactComponentType<P> component,
     P props
   ) {
-    SynapseReactClientFullContextProviderProps emptyContext = SynapseReactClientFullContextProviderProps.create(
-      SynapseContextJsObject.create(null, false, false),
-      null
-    );
+    SynapseReactClientFullContextProviderProps emptyContext =
+      SynapseReactClientFullContextProviderProps.create(
+        SynapseContextJsObject.create(null, false, false),
+        null
+      );
     return createElementWithSynapseContext(component, props, emptyContext);
   }
 
--- a/src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
@@ -7,8 +7,7 @@ import com.google.gwt.user.client.ui.IsWidget;
 import com.google.gwt.user.client.ui.SimplePanel;
 import com.google.gwt.user.client.ui.Widget;
 import com.google.inject.Inject;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Collections;
 import javax.annotation.Nullable;
 import org.gwtbootstrap3.client.ui.Alert;
 import org.gwtbootstrap3.client.ui.Button;
@@ -19,7 +18,6 @@ import org.sagebionetworks.repo.model.ManagedACTAccessRequirement;
 import org.sagebionetworks.repo.model.RestrictableObjectDescriptor;
 import org.sagebionetworks.repo.model.RestrictableObjectType;
 import org.sagebionetworks.schema.adapter.JSONObjectAdapter;
-import org.sagebionetworks.schema.adapter.JSONObjectAdapterException;
 import org.sagebionetworks.web.client.DisplayUtils;
 import org.sagebionetworks.web.client.GlobalApplicationState;
 import org.sagebionetworks.web.client.context.SynapseReactClientFullContextPropsProvider;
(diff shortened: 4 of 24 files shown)


src/main/java/org/sagebionetworks/web/client/jsinterop/AccessRequirementListProps.java
```diff
@@ -33,7 +33,7 @@ public static AccessRequirementListProps create(
     props.accessRequirementFromProps =
       accessRequirements
         .stream()
-        .map(o -> JSON.parse(o.toJSONString()))
+        .map(JSONEntityUtils::toJsInteropCompatibleObject)
```
nickgros, line 36: The new TableColumnSchemaForm component also serialize objects in this way, so add and use a common utility.

src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
```diff
@@ -25,6 +25,8 @@ public static native <P extends ReactComponentProps> ReactNode createElement(
     ReactNode[] children
   );
 
+  public static native <T> T createRef();
+
```
nickgros, line 29: Expose `createRef`

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
```diff
@@ -363,38 +361,29 @@ public void showRequestAccessModal(
     ManagedACTAccessRequirement accessRequirement,
     @Nullable RestrictableObjectDescriptor targetSubject
   ) {
-    try {
```
nickgros, line 366: Refactored so JSON serialization happens in the new utility, so the `try`/`catch` could be removed.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep2.java
```diff
@@ -118,30 +123,45 @@ public void configure(Table entity, TableType tableType) {
...
+      editorV2.configure(
+        EntityTypeUtils.getEntityType(entity),
+        getViewScope(),
+        defaultColumns
+      );
+    } else {
+      editor.configure(tableType, new ArrayList<>());
+      boolean isView = !TableType.table.equals(tableType);
+      this.editor.setAddDefaultColumnsButtonVisible(isView);
+      this.editor.setAddAnnotationColumnsButtonVisible(isView);
```
nickgros, line 138: These visibility conditions are handled internally by the React component, so the methods on the new widget are not 1:1 with the methods on the old widget.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorV2Widget.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.gwt.user.client.ui.IsWidget;
+import com.google.gwt.user.client.ui.Widget;
+import com.google.inject.Inject;
+import java.util.List;
+import org.sagebionetworks.repo.model.EntityType;
+import org.sagebionetworks.repo.model.table.ColumnModel;
+import org.sagebionetworks.repo.model.table.ViewScope;
+
+public class ColumnModelsEditorV2Widget implements IsWidget {
```
nickgros, line 11: I considered extracting a common interface between ColumnModelsEditorWidget (v1) and ColumnModelsEditorV2Widget, which would have simplified initial changes to the classes that use the editor. It would have been a fairly complicated refactor because

1. ColumnModelsEditorWidget (v1) also implements its view's presenter interface, which would be inappropriate to implement in v2. I'm not sure if the view is the only class to access the presenter methods, so it may have been tough to extract a new non-presenter interface.
2. The v2 implementation handles more logic internally, so it wouldn't have as many methods, so we would have ended up needing to implement many no-op methods. 
3. I don't immediately know if/how to use Guice to accomplish conditional dependency injection based on the experimental mode cookie. The logic could be in the constructor would work but it's not simple to reason with since some of our widgets are persistent and just get reconfigured while others are regularly destroyed and re-created


I ultimately decided that it wasn't worth it for temporary code.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorV2WidgetViewImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    List<ColumnModel> asList = new ArrayList<>();
+    for (Object o : editedColumnModels) {
+      try {
+        ColumnModel cm = new ColumnModel(
+          adapterFactory.createNew(JSON.stringify(o))
+        );
+        asList.add(cm);
+      } catch (JSONObjectAdapterException e) {
+        e.printStackTrace();
+      }
```
nickgros, line 86: I could add an alert container to the widget, or propagate this up, but I don't think this is common. Thoughts?

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorV2WidgetViewImpl.java
```diff
@@ -0,0 +1,100 @@
...
+    this.adapterFactory = adapterFactory;
+  }
+
+  @Override
+  public void configure(
+    EntityType entityType,
+    ViewScope viewScope,
+    List<ColumnModel> startingModels
+  ) {
+    componentRef = React.createRef();
```
nickgros, line 59: I was so happy to find out that this worked to give us a stable component ref

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsWidget.java
```diff
@@ -171,18 +166,28 @@ public void getPossibleColumnModelsForViewScope(String nextPageToken) {
...
+        );
       }
       scope.setViewEntityType(ViewEntityType.dataset);
     } else if (entity instanceof DatasetCollection) {
       scopeIds = new ArrayList<>();
       for (EntityRef item : ((DatasetCollection) entity).getItems()) {
         scopeIds.add(item.getEntityId());
+        scopeIds.add(
+          item.getEntityId() + "." + item.getVersionNumber().toString()
+        );
```
nickgros, line 180: Fixes SWC-6368

src/main/java/org/sagebionetworks/web/server/servlet/SynapseClientImpl.java
```diff
@@ -1,6 +1,7 @@
 package org.sagebionetworks.web.server.servlet;
 
 import au.com.bytecode.opencsv.CSVReader;
+import com.google.gwt.core.client.GWT;
```
nickgros:
```suggestion
```

src/test/java/org/sagebionetworks/web/unitclient/widget/table/modal/fileview/CreateTableViewWizardStep2Test.java
```diff
@@ -153,9 +167,12 @@ public class CreateTableViewWizardStep2Test {
...
     MockitoAnnotations.initMocks(this);
-    when(mockTableEntity.getId()).thenReturn(ENTITY_ID);
-    when(mockEntityView.getId()).thenReturn(ENTITY_ID);
-    when(mockSubmissionView.getId()).thenReturn(ENTITY_ID);
+    mockTableEntity = new TableEntity();
+    mockEntityView = new EntityView();
+    mockSubmissionView = new SubmissionView();
+    mockTableEntity.setId(ENTITY_ID);
+    mockEntityView.setId(ENTITY_ID);
+    mockSubmissionView.setId(ENTITY_ID);
```
nickgros: We are now passing these to EntityTypeUtils, which checks the class to give us an `EntityType`. That check doesn't work with mocks, so just use "real" data objects.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep2.java
```diff
@@ -118,30 +123,45 @@ public void configure(Table entity, TableType tableType) {
...
+      return Collections.emptyList();
+    }
+  }
+
+  public ViewScope getViewScope() {
     ViewScope scope = new ViewScope();
     List<String> scopeIds = null;
-    if (entity instanceof EntityView) {
+    if (entity instanceof TableEntity) {
+      return new ViewScope();
```
jay-hodgson: A TableEntity has a ViewScope?

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorV2Widget.java
jay-hodgson: Totally agree, no need to share a common interface.  In general, I think the more logic that's in new implementation, the less complicated the interface will be.

jay-hodgson: The separation will make deletion of v1 that much easier.

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorV2WidgetViewImpl.java
jay-hodgson: I'm a bit surprised that this worked!

src/main/java/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorV2WidgetViewImpl.java
jay-hodgson: I don't expect this parsing error to ever happen, I think sending it to the console is ok

src/main/resources/org/sagebionetworks/web/client/widget/table/v2/schema/ColumnModelsEditorV2WidgetViewImpl.ui.xml
```diff
@@ -0,0 +1,10 @@
+<ui:UiBinder
+  xmlns:ui="urn:ui:com.google.gwt.uibinder"
+  xmlns:g="urn:import:com.google.gwt.user.client.ui"
+  xmlns:b="urn:import:org.gwtbootstrap3.client.ui"
+  xmlns:b.html="urn:import:org.gwtbootstrap3.client.ui.html"
+  xmlns:t="urn:import:org.sagebionetworks.web.client.view.bootstrap.table"
+  xmlns:w="urn:import:org.sagebionetworks.web.client.widget"
```
jay-hodgson, line 3: Could clean these imports (I think you just need w)

src/test/java/org/sagebionetworks/web/unitclient/widget/table/modal/fileview/CreateTableViewWizardStep2Test.java
```diff
@@ -153,9 +167,12 @@ public class CreateTableViewWizardStep2Test {
   @Before
   public void before() {
     MockitoAnnotations.initMocks(this);
-    when(mockTableEntity.getId()).thenReturn(ENTITY_ID);
-    when(mockEntityView.getId()).thenReturn(ENTITY_ID);
-    when(mockSubmissionView.getId()).thenReturn(ENTITY_ID);
+    mockTableEntity = new TableEntity();
+    mockEntityView = new EntityView();
+    mockSubmissionView = new SubmissionView();
```
jay-hodgson: These could be renamed to be more appropriate (tableEntityStub for example)

pom.xml
```diff
@@ -19,7 +19,7 @@
     <webappDirectory>
       ${project.build.directory}/${project.build.finalName}
     </webappDirectory>
-    <synapse.version>462.0</synapse.version>
+    <synapse.version>475.0</synapse.version>
```
jay-hodgson, line 22: Thank you for updating the synapse version as well!

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/table/modal/fileview/CreateTableViewWizardStep2.java
nickgros: I think I hit an NPE but the right thing to do would be to return null and explicitly handle the null case, thanks for catching this.

pom.xml
nickgros: It was necessary to properly serialize the JSON Subcolumns!

PR #5230, opened 2023-11-20
SWC-6575 - Fix issue where stale cache data was shown after creating a stable version of a dataset

nickgros:
SWC-6575

 - Add Java bindings for SRC KeyFactory to ensure reliable cache invalidation
 - Fire EntityUpdatedEvent when view/dataset snapshots are created to ensure cache is invalidated and prevent stale version data from being shown


src/main/java/org/sagebionetworks/web/client/jsinterop/KeyFactory.java
```diff
@@ -0,0 +1,16 @@
+package org.sagebionetworks.web.client.jsinterop;
+
+import java.util.List;
+import jsinterop.annotations.JsConstructor;
+import jsinterop.annotations.JsType;
+
+@JsType(isNative = true, namespace = "SRC.SynapseQueries")
+public class KeyFactory {
```
nickgros, line 8: Bindings for KeyFactory, which provides query keys used for `react-query`

src/main/java/org/sagebionetworks/web/client/jsinterop/KeyFactory.java
```diff
@@ -0,0 +1,16 @@
...
+
+  @JsConstructor
+  public KeyFactory(String accessToken) {}
+
+  public native List<?> getDownloadListBaseQueryKey();
+
+  public native List<?> getEntityQueryKey(String entityId);
```
nickgros, line 15: There are many other methods, but these are the only keys we use today in SWC

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -401,24 +403,23 @@ public Widget asWidget() {
...
-        QueryKeyConstants.ENTITY,
-        event.getEntityId()
-      );
-    queryClient.resetQueries(queryKey);
+    KeyFactory keyFactory = keyFactoryProvider.getKeyFactory(
+      authenticationController.getCurrentUserAccessToken()
+    );
+
+    List<?> queryKey = keyFactory.getEntityQueryKey(event.getEntityId());
+    queryClient.invalidateQueries(queryKey);
```
nickgros, line 411: Use the tool to generate the cache key instead of relying on a consistent implementation

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2527,6 +2527,7 @@ public void onComplete(TableUpdateTransactionResponse response) {
                 if (!errors.isEmpty()) {
                   view.showErrorMessage(errors);
                 } else {
+                  fireEntityUpdatedEvent();
```
nickgros, line 2530: Invalidate the cache when a snapshot of a view/dataset is created

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/controller/EntityActionControllerImplTest.java
```diff
@@ -2468,9 +2468,9 @@ public void testOnCreateEntityViewSnapshot() {
     handler.onComplete(mockTableUpdateTransactionResponse);
     verify(mockView).hideCreateVersionDialog();
     verify(mockPopupUtils).notify(any(), any(), any(ToastMessageOptions.class));
+    // EntityUpdatedEvent is fired to invalidate caches
+    verify(mockEventBus).fireEvent(any(EntityUpdatedEvent.class));
     verify(mockPlaceChanger).goTo(any());
-    // Don't fire an updated event; we used placeChanger to go to a new place.
-    verify(mockEventBus, never()).fireEvent(any(EntityUpdatedEvent.class));
```
nickgros, line 2473: Alternatively we could have placechanger invalidate all caches, but that's pretty heavy-handed

src/main/java/org/sagebionetworks/web/client/context/KeyFactoryProviderImpl.java
```diff
@@ -0,0 +1,11 @@
...
+
+import org.sagebionetworks.web.client.jsinterop.KeyFactory;
+
+public class KeyFactoryProviderImpl implements KeyFactoryProvider {
+
+  @Override
+  public KeyFactory getKeyFactory(String currentUserAccessToken) {
+    return new KeyFactory(currentUserAccessToken);
+  }
+}
```
nickgros, line 11: Create a mockable/injectable KeyFactoryProvider so I can easily mock KeyFactory in tests that don't support JsInterop

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/SynapseReactClientEntityQueryKey.java
nickgros, line 1: No longer need this object that represented one type of query key, since we can now just get a query key from KeyFactory.

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/controller/EntityActionControllerImplTest.java
jay-hodgson: Like to avoid this option

jay-hodgson (approved):

PR #5241, opened 2023-12-07
SWC-6557

hallieswan:
- add end-to-end test for table CRUD
- bump Playwright version

### DIFF
--- a/e2e/files.spec.ts
+++ b/e2e/files.spec.ts
@@ -170,7 +170,7 @@ const dismissFileUploadAlert = async (page: Page) => {
 const getFileMD5 = async (page: Page) => {
   return await testAuth.step('get file MD5', async () => {
     const row = page.getByRole('row').filter({ hasText: 'MD5' })
-    expect(row.getByRole('cell')).toHaveCount(2)
+    await expect(row.getByRole('cell')).toHaveCount(2)
     return row.getByRole('cell').filter({ hasNotText: 'MD5' }).textContent()
   })
 }
--- a/e2e/fixtures/authenticatedUserPages.ts
+++ b/e2e/fixtures/authenticatedUserPages.ts
@@ -94,7 +94,11 @@ export const testAuth = base.extend<
 
       const cleanupPage = await createPage(browser)
       for (const userId of userIds) {
-        await cleanupTestUser(userId, cleanupPage)
+        try {
+          await cleanupTestUser(userId, cleanupPage)
+        } catch (error) {
+          console.error(`Failed to clean up userId: ${userId}:\n${error}`)
+        }
       }
       await cleanupPage.close()
     },
--- a/e2e/helpers/tableColumnSchemaEditor.ts
+++ b/e2e/helpers/tableColumnSchemaEditor.ts
@@ -104,7 +104,7 @@ const enterColumnRestrictValues = async (
 
     await page.getByRole('button', { name: 'Paste new values' }).click()
     await page
-      .getByPlaceholder('Place comma or tab delimited')
+      .getByPlaceholder('Place comma or tab delimited values here')
       .fill(values.join(','))
     await page.getByRole('button', { name: 'Add', exact: true }).click()
     await page.getByRole('button', { name: 'OK' }).click()
--- a/e2e/helpers/tableColumnSchemaEditor.ts
+++ b/e2e/helpers/tableColumnSchemaEditor.ts
@@ -166,12 +166,16 @@ const enterColumnSchema = async (
 export const addColumnsFromColumnSchemaConfig = async (
   page: Page,
   columnConfigs: ColumnSchemaConfig[],
-  columnStartingIndex: number = 0,
 ) => {
   await test.step('add columns from column schema config', async () => {
     const addColumnButton = page.getByRole('button', {
       name: 'Add Column',
     })
+
+    const columnStartingIndex =
+      await test.step('calculate starting index', async () => {
+        return await page.getByPlaceholder('Column name').count()
+      })
 
     for (
       let columnIndex = columnStartingIndex;
--- a/e2e/helpers/tableColumnSchemaEditor.ts
+++ b/e2e/helpers/tableColumnSchemaEditor.ts
@@ -1,0 +1,290 @@
+import { Locator, Page, expect, test } from '@playwright/test'
+
+// Types -----------------------------------------------------------------------
+type FacetType = 'Range' | 'Values'
+export type ColumnSchemaConfig = {
+  name: string
+  type: string
+  size?: number
+  maxListLength?: number
+  defaultValue?: string
+  restrictValues?: string[]
+  facet?: FacetType
+}
+type ColumnSchemaKey = keyof ColumnSchemaConfig
+const columnSchemaIndexMap: Record<ColumnSchemaKey, number> = {
+  name: 0,
+  type: 1,
+  size: 2,
+  maxListLength: 3,
+  defaultValue: 4,
+  restrictValues: 5,
+  facet: 6,
+}
+
+// Set Individual Column Schema ------------------------------------------------
+const enterColumnName = async (
+  page: Page,
+  columnIndex: number,
+  columnName: string,
+) => {
+  await test.step('enter column name', async () => {
+    const columnNameFields = page.getByPlaceholder('Column name')
+    expect(columnNameFields).toHaveCount(columnIndex + 1)
+    await columnNameFields.nth(columnIndex).fill(columnName)
+  })
+}
+
+const selectColumnType = async (
+  page: Page,
+  columnIndex: number,
+  columnType: string,
+) => {
+  await test.step('select column type', async () => {
+    const columnTypeFields = page.getByLabel('Column Type')
+    await columnTypeFields.nth(columnIndex).click()
+    await page.getByRole('option', { name: columnType, exact: true }).click()
+  })
+}
+
+const setColumnMaxSize = async (
+  page: Page,
+  columnIndex: number,
+  maxSize: number,
+) => {
+  await test.step('set column size', async () => {
+    const columnMaxSizeFields = page.getByLabel('Maximum Size')
+    await columnMaxSizeFields.nth(columnIndex).fill(maxSize.toString())
+  })
+}
+
+const setColumnMaxListLength = async (
+  page: Page,
+  columnIndex: number,
+  maxListLength: number,
+) => {
+  await test.step('set column max list length', async () => {
+    const columnMaxListLengthFields = page.getByLabel('Maximum List Length')
+    await columnMaxListLengthFields
+      .nth(columnIndex)
+      .fill(maxListLength.toString())
+  })
+}
+
+const setColumnDefaultValue = async (
+  page: Page,
+  columnIndex: number,
+  defaultValue: string,
+  columnType: string,
+) => {
+  await test.step('set column default value', async () => {
+    const columnDefaultValueField = page
+      .getByLabel('Default Value')
+      .nth(columnIndex)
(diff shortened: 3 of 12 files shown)


e2e/fixtures/authenticatedUserPages.ts
```diff
@@ -94,7 +94,11 @@ export const testAuth = base.extend<
 
       const cleanupPage = await createPage(browser)
       for (const userId of userIds) {
-        await cleanupTestUser(userId, cleanupPage)
+        try {
+          await cleanupTestUser(userId, cleanupPage)
+        } catch (error) {
+          console.error(`Failed to clean up userId: ${userId}:\n${error}`)
+        }
```
hallieswan, line 101: Allow test to pass, even if we fail to clean up the test user. Handles users that have [created tables](https://sagebionetworks.jira.com/browse/PLFM-8179).

e2e/helpers/tableColumnSchemaEditor.ts
```diff
@@ -0,0 +1,294 @@
...
+  defaultValue: string,
+  columnType: string,
+) => {
+  await test.step('set column default value', async () => {
+    const columnDefaultValueField = page
+      .getByLabel('Default Value')
+      .nth(columnIndex)
+    if (columnType === 'Boolean') {
+      await columnDefaultValueField.click()
+      await page.getByRole('option', { name: defaultValue }).click()
```
hallieswan, line 87: `Boolean` fields provide a drop-down for default values, rather than an input field

e2e/helpers/tableColumnSchemaEditor.ts
```diff
@@ -0,0 +1,294 @@
...
+  }
+  if (columnConfig.facet) {
+    await selectColumnFacet(page, columnIndex, columnConfig.facet)
+  }
+}
+
+export const addColumnsFromColumnSchemaConfig = async (
+  page: Page,
+  columnConfigs: ColumnSchemaConfig[],
+  columnStartingIndex: number = 0,
```
hallieswan: By default, enter all of the columns in the provided config into the schema editor. However, optionally specify the config index at which to start entering columns in the schema editor. Allows this function to be reused when updating an existing column schema.

e2e/helpers/tables.ts
```diff
@@ -0,0 +1,420 @@
...
+    await expect(
+      page
+        .locator('div')
+        .filter({ hasText: new RegExp(`^${tableDescription}$`, 'i') }),
+    ).toBeVisible()
+
+    await expect(page.getByText('Loading provenance...')).not.toBeVisible()
+    await expect(
+      page.getByRole('progressbar').locator('visible=true'),
+    ).toHaveCount(0, { timeout: defaultExpectTimeout * 2 })
```
hallieswan, line 45: Allow time for the table to fully load

e2e/helpers/tables.ts
```diff
@@ -0,0 +1,420 @@
...
+
+// Validate table data ---------------------------------------------------------
+const getTableRows = async (page: Page) => {
+  return page.getByTestId('SynapseTable').getByRole('row')
+}
+
+export const expectTableColumnHeaders = async (
+  tableRows: Locator,
+  columnsSchemaConfig: ColumnSchemaConfig[],
+  columnIndexOffset: number,
```
hallieswan, line 208: Provide a `columnIndexOffset`, so this function can be used to check the table column headers when entering data into the table (and there is a column of checkboxes, so offset is 1) and when viewing the table (and the offset is 0).

e2e/helpers/tables.ts
```diff
@@ -0,0 +1,420 @@
...
+  tableData: TableDataConfig[],
+) => {
+  await test.step('confirm table data', async () => {
+    const tableRows = await getTableRows(page)
+    await expectTableColumnHeaders(tableRows, columnsSchemaConfig, 0)
+
+    for (let dataIndex = 0; dataIndex < tableData.length; dataIndex++) {
+      await expectTableRowCorrect(
+        tableRows,
+        dataIndex + 1,
```
hallieswan, line 301: Skips the first table row, which contains the column names

e2e/tables.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+const noTablesText =
+  'There are no Tables associated with this project. Any Tables you create in this project will appear here'
+
+test.describe('Tables', () => {
+  testAuth.beforeAll(async ({ browser, storageStatePaths }) => {
+    userProject = await setupProject(browser, 'swc-e2e-user', storageStatePaths)
+  })
+
+  testAuth.afterAll(async ({ browser }) => {
+    test.slow()
```
hallieswan, line 52: Test takes 1-2 minutes. Mark as slow, so the test has enough time to finish when CI is slow.

e2e/tables.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+        ).toBeVisible()
+        await expect(createTableHeader).not.toBeVisible()
+      })
+    })
+
+    await test.step('Table has been created with expected schema', async () => {
+      await expectTablePageLoaded(userPage, tableName, tableDescription)
+
+      await test.step('get table entity id', async () => {
+        tableEntityId =
```
hallieswan, line 193: Table entity will be moved to the Trash Can in the test, but deleted via API in the `afterAll`

e2e/helpers/tableColumnSchemaEditor.ts
nickgros: Would it make sense to try to calculate the number of existing columns in this function? Then there would be no need to pass a starting index, if I'm understanding correctly

e2e/helpers/tables.ts
```diff
@@ -0,0 +1,420 @@
...
+
+      if (decadeDifference !== 0) {
+        await test.step('change decade', async () => {
+          const arrowSelector =
+            decadeDifference > 0 ? '.prev:visible' : '.next:visible'
+          for (let i = 0; i < decadeDifference; i++) {
+            await page.locator(arrowSelector).click()
+          }
+        })
+      }
```
nickgros: Sorry, what's going on here? Not clear to me why we're changing the decade (conditionally?)

e2e/tables.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+  testAuth('should allow table CRUD', async ({ userPage }) => {
+    test.slow()
+
+    const tableName = generateEntityName('table')
+    const tableDescription = 'table test description'
+
+    const initialColumnsSchemaConfig: ColumnSchemaConfig[] = [
+      { name: 'item', type: 'String', size: 20 },
+      { name: 'quantity', type: 'Integer', defaultValue: '500' },
+      // { name: 'on_sale', type: 'Boolean', defaultValue: 'false' }, - currently broken, SWC-6630
```
nickgros, line 76: Thanks for writing + commenting these cases and mentioning the associated ticket! I attached comments to each ticket as a reminder (for me) to come back through and fix all of these!

nickgros (approved):
Looks great, had some small questions

e2e/helpers/tableColumnSchemaEditor.ts
hallieswan: Oh that's an interesting idea! I'd thought that allowing the user to pass a starting index would provide the flexibility to use this function to update existing columns as well as adding new columns, but maybe that's overly complicated. What do you think?

e2e/helpers/tables.ts
hallieswan: Sorry this wasn't clear! By default, the years in the current decade are shown when picking a year. If the desired year is contained within the current decade, the user can just click that year.

However, if the current year's decade doesn't contain the desired year, then the user needs to choose whether to click the left arrow to view previous decades or the right arrow to view future decades. The user can then click the arrow the correct number of times to view the decade containing the desired year. See an example below, where the current year is 2023, but the desired year is 1991.

Let me know if there's a way I can improve the code to make it clearer!

https://github.com/Sage-Bionetworks/SynapseWebClient/assets/26949006/36cd8f54-53d3-4e84-85c2-a02a745c918f

e2e/tables.spec.ts
hallieswan: Thank you!!

e2e/helpers/tableColumnSchemaEditor.ts
nickgros: Doesn't it always click the "Add column" button? If you edited an existing column, you'd end up with an extra empty column at the end, right?

hallieswan: Oh good call, that's right. I'll change the function to calculate the current number of columns, so the starting index doesn't have to be passed. If we need to edit the schema, we can add an "editSchema..." function. Thanks!

e2e/helpers/tables.ts
nickgros: That makes sense! It makes it super-resilient. My suggestion is to consider encapsulating this logic in a function that is dedicated to picking a date specifically using the `goog.ui.DatePicker` implementation, then just call that here.

hallieswan: Good idea! Will do

e2e/files.spec.ts
```diff
@@ -170,7 +170,7 @@ const dismissFileUploadAlert = async (page: Page) => {
 const getFileMD5 = async (page: Page) => {
   return await testAuth.step('get file MD5', async () => {
     const row = page.getByRole('row').filter({ hasText: 'MD5' })
-    expect(row.getByRole('cell')).toHaveCount(2)
+    await expect(row.getByRole('cell')).toHaveCount(2)
```
hallieswan, line 173: `toHaveCount` is an [auto-retrying assertion](https://playwright.dev/docs/test-assertions#auto-retrying-assertions), so should be awaited.

e2e/helpers/tableColumnSchemaEditor.ts
```diff
@@ -166,13 +166,17 @@ const enterColumnSchema = async (
...
     const addColumnButton = page.getByRole('button', {
       name: 'Add Column',
     })
 
+    const columnStartingIndex =
+      await test.step('calculate starting index', async () => {
+        return await page.getByPlaceholder('Column name').count()
+      })
+
     for (
```
hallieswan, line 180: Calculate starting index from which to add new columns, rather than requiring user to pass the starting index.

e2e/helpers/tables.ts
```diff
@@ -140,56 +140,11 @@ export const selectTableDate = async (
...
-      })
-
-      // handle case where day may also be visible in previous or next month
-      // e.g. 1 or 30
-      if ((await dayLocator.count()) > 1) {
-        dayLocator = day > 15 ? dayLocator.nth(1) : dayLocator.first()
-      }
-      await dayLocator.click()
-    })
+    await selectDateUsingGoogUiDatePicker(page, year, month, day)
```
hallieswan, line 147: Encapsulate logic specific to the goog.ui.DatePicker implementation in a separate function.

e2e/tables.spec.ts
```diff
@@ -276,10 +276,16 @@ test.describe('Tables', () => {
         })
 
       await test.step('add new columns', async () => {
+        // wait for previously entered column schemas to appear
+        // ...so addColumnsFromColumnSchemaConfig calculates correct starting column index
+        await expect(
+          userPage.getByPlaceholder('Column name'),
+          'should load previously entered column schemas',
+        ).toHaveCount(initialColumnsSchemaConfig.length)
```
hallieswan, line 284: In webkit in CI, the table column schema editor loaded inconsistently, so the starting column index would intermittently be calculated before the previous column schemas had loaded and result in test failures.

nickgros (approved):

PR #5242, opened 2023-12-07
SWC-6614

jay-hodgson:
Existing SubjectsWidget is now the TeamSubjectsWidget, and create a new EntitySubjectsWidget that uses the SRC EntityHeaderTable (need to update SRC before merge!)


jay-hodgson:
AR with Team subject:
Screen Shot 2023-12-07 at 10 06 00 AM

Editing still works:
Screen Shot 2023-12-07 at 10 06 11 AM

With single entity:
Screen Shot 2023-12-07 at 10 35 31 AM
Editing:
Screen Shot 2023-12-07 at 10 41 20 AM

Unsaved changes error shown if clicking Next with value in text field:
Screen Shot 2023-12-07 at 11 03 56 AM

ADKP real-world large subject list example:
Screen Shot 2023-12-07 at 11 06 17 AM

Editing:
Screen Shot 2023-12-07 at 11 07 10 AM

src/main/java/org/sagebionetworks/web/client/jsinterop/EntityHeaderTableProps.java
```diff
@@ -0,0 +1,79 @@
...
+  @FunctionalInterface
+  @JsFunction
+  public interface OnUpdateReferenceList {
+    void onUpdate(ReferenceJsObject[] referenceListJsObject);
+  }
+
+  @FunctionalInterface
+  @JsFunction
+  public interface OnUpdateString {
+    void onUpdatString(String newValue);
```
nickgros, line 29: This typo shouldn't matter since this is a functional interface
```suggestion
    void onUpdateString(String newValue);
```

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/createaccessrequirement/CreateAccessRequirementStep1.java
```diff
@@ -212,4 +200,12 @@ public void setModalPresenter(ModalPresenter modalPresenter) {
     this.modalPresenter = modalPresenter;
     modalPresenter.setPrimaryButtonText(DisplayConstants.NEXT);
```
nickgros, line 201: When editing an existing AR, can you verify that the changes to the AR are saved after clicking this button, and if so, change the text to something like "Save Changes and Continue"?

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/EntitySubjectsWidget.java
```diff
@@ -0,0 +1,102 @@
...
+  public void configure(
+    final List<RestrictableObjectDescriptor> subjects,
+    CallbackP<List<RestrictableObjectDescriptor>> onUpdate
+  ) {
+    this.onUpdate = onUpdate;
+    isACTMemberAsyncHandler.isACTActionAvailable(
+      new CallbackP<Boolean>() {
+        @Override
+        public void invoke(Boolean isACT) {
+          view.setVisible(isACT);
```
nickgros, line 55: Why only show subjects for ACT? This could be useful for data administrators, Sage employees, etc

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/EntitySubjectsWidget.java
```diff
@@ -0,0 +1,102 @@
...
+    List<Reference> referenceList = new ArrayList<>();
+    for (RestrictableObjectDescriptor rod : subjects) {
+      if (rod.getType().equals(RestrictableObjectType.ENTITY)) {
+        Reference ref = new Reference();
+        ref.setTargetId(rod.getId());
+        referenceList.add(ref);
+      }
+    }
+    entityReferences.setReferences(referenceList);
+    boolean isEditable = onUpdate != null;
```
nickgros, line 77: If we can show this to non-ACT (above comment), also make ACT membership a requirement for `isEditable`

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/createaccessrequirement/CreateAccessRequirementStep1.java
```diff
@@ -130,38 +121,36 @@ private void refreshSubjects() {
...
   public void onPrimary() {
     // SWC-3700: validate that subjects have been set
     if (subjects.size() == 0) {
       modalPresenter.setErrorMessage(EMPTY_SUBJECT_LIST_ERROR_MESSAGE);
       return;
     }
+    if (!entitySubjectsWidget.isEntityIDsTextboxEmpty()) {
+      modalPresenter.setErrorMessage(UNSAVED_ENTITY_IDS_ERROR_MESSAGE);
+      return;
+    }
```
nickgros, line 153: This is a nice addition!

nickgros (approved):
Looks great pending https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/599 and upgrading the version of synapse-react-client here.

I attached a few questions

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/EntitySubjectsWidget.java
jay-hodgson: This is a good question.  Generally speaking, I don't think most users will ever see the old AR renderers (because the link to the AR page is an ACT-only command).  Since it wasn't shown before, I'm not going to change it in this PR.  But happy to revisit

PR #5245, opened 2023-12-12
SWC-6624: store Playwright reports with traces in S3 when e2e workflow runs in Sage repo

hallieswan:
| **..** | **Sage-Bionetworks repo** | **forked repos**
:---:|:---:|:---:
traces | included in report | not included in report 
report location | S3 bucket | GitHub artifact
saved report format | one blob per shard | archive of blobs and merged HTML report
trigger | push to `develop` and `release-**` branches | push
job concurrency | restricted - allow at most one `playwright-tests` job to run at a time within the repo | unrestricted
matrix strategy | 3 shards | 3 shards
matrix parallelization |  restricted - allow at most one shard to run at a time per workflow run | allow all three shards to run at the same time per workflow run

### DIFF
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -9,9 +9,9 @@ env:
   PW_ALL_BLOBS_DIR: all-blob-reports
 jobs:
   build:
-    # Run in Sage repo on release- branches
+    # Run in Sage repo on develop or release- branches
     # and on all branches in user-owned forks
-    if: ${{ startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
+    if: ${{ github.ref_name == 'develop' || startsWith(github.ref_name, 'release-') || github.actor == github.repository_owner }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -27,6 +27,15 @@ jobs:
     needs: [build]
     runs-on: 'macos-latest'
     timeout-minutes: 60
+    # Ensure that at most one playwright-tests job will run at a time
+    # Per https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
+    #  "When a concurrent job or workflow is queued, if another job or workflow
+    #   using the same concurrency group in the repository is in progress,
+    #   the queued job or workflow will be pending. Any previously pending job or
+    #   workflow in the concurrency group will be cancelled."
+    # Related discussion here: https://github.com/orgs/community/discussions/5435
+    concurrency:
+      group: ${{ github.workflow }}-playwright-tests
     strategy:
       max-parallel: ${{ github.repository_owner == 'Sage-Bionetworks' && 1 || 3 }}
       fail-fast: false
--- a/.github/workflows/build-test-e2e.yml
+++ b/.github/workflows/build-test-e2e.yml
@@ -44,6 +44,13 @@ jobs:
     env:
       CONTAINER_NAME: 'swc-tomcat'
       CI: true
+      DESTINATION_BUCKET: e2e-reports-bucket-bucket-1p1qz6p48t4uy
+      REPORT_ID: ${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}
+      IS_SAGE_REPO: ${{ github.repository_owner == 'Sage-Bionetworks' }}
+    # Allow OIDC Integration (so we can assume the AWS role to deploy)
+    permissions:
+      id-token: write
+      contents: read
     steps:
       - uses: actions/checkout@v3
       - name: Setup docker (missing on MacOS)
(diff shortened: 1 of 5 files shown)


.github/workflows/build-test-e2e.yml
```diff
@@ -27,6 +27,16 @@ jobs:
...
+    # Ensure that at most one playwright-tests job will run at a time in Sage-Bionetworks repo,
+    # but allow multiple playwright-test jobs to run concurrently in forked repos
+    # Per https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
+    #  "When a concurrent job or workflow is queued, if another job or workflow
+    #   using the same concurrency group in the repository is in progress,
+    #   the queued job or workflow will be pending. Any previously pending job or
+    #   workflow in the concurrency group will be cancelled."
+    # Related discussion here: https://github.com/orgs/community/discussions/5435
+    concurrency:
+      group: ${{ github.repository_owner == 'Sage-Bionetworks' && format('${0}-${1}', github.workflow, '-playwright-tests') || format('${0}-${1}', github.run_id, matrix.shard ) }}
```
hallieswan, line 39: Apply a concurrency group to `playwright-tests` jobs so that the e2e workflow uses at most one macOS runner at a time. GitHub's current concurrency implementation cancels previously queued pending jobs, rather than queuing and executing all pending jobs. So if more than two e2e workflow runs are triggered at the same time, depending on when each `playwright-tests` job starts, a subset or none of the e2e shards may run per workflow run. However, once all other jobs have finished running, the cancelled job(s) could be manually re-triggered, if needed.

For now, plan to monitor the e2e tests and the frequency of cancelled jobs, then decide whether another solution should be investigated.

Note: I applied concurrency at the `playwright-tests` job level, so that the `build` job (which builds SWC and runs unit tests on a Linux runner) will always run, regardless of how many e2e workflows have been triggered.

.github/workflows/build-test-e2e.yml
```diff
@@ -66,10 +82,23 @@ jobs:
       - name: Run Playwright tests
         env:
           ADMIN_PAT: ${{ secrets.ADMIN_PAT }}
-        run: yarn playwright test --shard ${{ matrix.shard }}
+          TRACE_TOGGLE: ${{ github.repository_owner == 'Sage-Bionetworks' && 'on' || 'off'}}
+        run: yarn playwright test --shard ${{ matrix.shard }} --trace=${{ env.TRACE_TOGGLE }}
```
hallieswan, line 86: Only record a trace when running the workflow in the Sage-Bionetworks repo

.github/workflows/build-test-e2e.yml
```diff
@@ -66,10 +82,23 @@ jobs:
...
+          role-session-name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
+          role-duration-seconds: 1200
+      - name: Upload blob report to S3
+        if: github.repository_owner == 'Sage-Bionetworks'
+        run: |
+          aws s3 sync ./blob-report --region us-east-1 s3://${{ env.DESTINATION_BUCKET }}/${{ env.REPORT_ID }}/
       - name: Upload blob report to GitHub Actions Artifacts
         uses: actions/upload-artifact@v3
-        if: always()
+        if: ${{ github.repository_owner != 'Sage-Bionetworks' && always() }}
```
hallieswan, line 101: Only attach the blob report as a GitHub artifact in forked repos

.github/workflows/build-test-e2e.yml
```diff
@@ -83,7 +112,7 @@ jobs:
   merge-reports:
     # Merge reports after playwright-tests, even if some shards have failed
     # But skip this job if the previous job was cancelled or skipped
-    if: ${{ !cancelled() && needs.playwright-tests.result != 'skipped' }}
+    if: ${{ github.repository_owner != 'Sage-Bionetworks' && !cancelled() && needs.playwright-tests.result != 'skipped' }}
```
hallieswan, line 115: Similarly, only merge the blob reports into an HTML report in forked repos

e2e_workflow/README.md
```diff
@@ -30,7 +30,40 @@ Notes:
...
+
+#### Forked Repositories
+
+In your forked repository, ensure that [Actions are enabled](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository), then create [an Actions secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository) for `ADMIN_PAT`. The tests will run on each push to your forked repository. The Playwright report will be saved as a GitHub Artifact.
+
+The GitHub UI or CLI can be used to view the reports:
+
+- GitHub UI
+  - Navigate to the Action run summary page.
+  - Download the report named "html-report--atttempt-{number}". _Note:_ only the report from the latest attempt will be available.
```
nickgros:
```suggestion
  - Download the report named "html-report--attempt-{number}". _Note:_ only the report from the latest attempt will be available.
```

.github/workflows/build-test-e2e.yml
```diff
@@ -66,10 +82,23 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::449435941126:role/sagebase-github-oidc-synapsedev-e2e-infra
+          role-session-name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
+          role-duration-seconds: 1200
+      - name: Upload blob report to S3
+        if: github.repository_owner == 'Sage-Bionetworks'
+        run: |
+          aws s3 sync ./blob-report --region us-east-1 s3://${{ env.DESTINATION_BUCKET }}/${{ env.REPORT_ID }}/
```
nickgros, line 98: Would a retention policy be specified here? Or is it specified on the bucket?

e2e_workflow/view_github_report.sh
```diff
@@ -0,0 +1,57 @@
...
+for file in $(find "${BLOB_DIR}" -type f) $(find "${REPORT_DIR}" -type f); do 
+  [ -f "${file}" ] && echo "delete: ${file}" && rm "${file}"
+done
+
+# Get workflow artifact id and name
+ARTIFACT=$(gh api \
+  -H "Accept: application/vnd.github+json" \
+  -H "X-GitHub-Api-Version: 2022-11-28" \
+  "/repos/${REPO_OWNER}/${REPO_NAME}/actions/runs/${RUN_ID}/artifacts" \
+  | jq '.artifacts[] | select(.name | contains("html-report")) | {id, name}')
```
nickgros, line 40: `jq` may not be installed on dev machines, might want to add that to instructions (unless maybe it's a dependency of `gh`?)

nickgros (approved):
LGTM! The bash scripts will be especially helpful, and the README updates are very much appreciated.

.github/workflows/build-test-e2e.yml
hallieswan: The policy is specified on the bucket -- objects are set to expire after 14 days, which matches the retention set for reports stored as GitHub artifacts.

e2e_workflow/view_github_report.sh
hallieswan: Oh good call! I don't believe `jq` is a dependency of `gh`, so I'll make a note to install it in the README.

xschildw (approved):
Nice! Couple of questions offline.

PR #5250, opened 2024-01-06
SWC-6621

hallieswan:
- lint e2e tests
- fix errors raised by linter
- use types from @sage-bionetworks/synapse-types


.eslintrc.js
```diff
@@ -0,0 +1,43 @@
...
+    'prefer-const': 'warn',
+    '@typescript-eslint/restrict-template-expressions': [
+      'warn',
+      { allowBoolean: true },
+    ],
+    '@typescript-eslint/unbound-method': 'warn',
+    '@typescript-eslint/no-inferrable-types': 'off',
+    '@typescript-eslint/ban-ts-comment': 'warn',
+    '@typescript-eslint/no-empty-interface': 'warn',
+    '@typescript-eslint/no-floating-promises': 'error',
```
hallieswan, line 30: For the most part, these rules are a direct subset of rules in [synapse-web-monorepo](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/.eslintrc.cjs). However, `@typescript-eslint/no-floating-promises` is set to error to catch issues in Playwright tests, as recommended in their [best practices](https://playwright.dev/docs/best-practices#lint-your-tests) guide.

.github/workflows/build-test-e2e.yml
```diff
@@ -85,6 +99,12 @@ jobs:
         run: yarn install --frozen-lockfile
       - name: Install Playwright Browsers
         run: yarn playwright install --with-deps
+      - name: Confirm Docker container is running
```
hallieswan, line 102: Trying to troubleshoot the `TypeError: The argument 'file' cannot be empty. Received ''` error, as seen in [this run](https://github.com/hallieswan/SynapseWebClient/actions/runs/7425928398/job/20208782799). I can reproduce the error when the SWC war does not exist or when the docker container is not running (likely because in CI, we pass a stub to the [PW webServer command](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/6f689abdaaa696d8664a380c1e2089fe43304d2f/playwright.config.ts#L93)).

e2e/discussions.spec.ts
```diff
@@ -79,10 +79,7 @@ const expectThreadReplyVisible = async (
   })
 }
 
-const getDiscussionParentActionButtons = async (
-  page: Page,
-  threadTitle: string,
-) => {
+const getDiscussionParentActionButtons = (page: Page, threadTitle: string) => {
```
hallieswan, line 82: `Async arrow function 'getDiscussionParentActionButtons' has no 'await' expression`

e2e/discussions.spec.ts
```diff
@@ -113,7 +107,7 @@ const getDiscussionReplyActionButtons = async (
 }
 
 let userProject: Project
-let fileHandleIds: string[] = []
+const fileHandleIds: string[] = []
```
hallieswan, line 110: `fileHandleIds` is never reassigned - prefer-const

e2e/files.spec.ts
```diff
@@ -344,72 +344,85 @@ test.describe('Files', () => {
     },
   )
 
-  testAuth(
-    'should create and delete a file',
-    async ({ userPage, browserName }) => {
```
hallieswan, line 349: I removed `browserName`, since it wasn't used. This resulted in many whitespace changes, so the diff isn't very helpful -- no other changes were made in this file

e2e/fixtures/authenticatedUserPages.ts
```diff
@@ -110,8 +114,9 @@ export const testAuth = base.extend<
   // ...per worker + browserName
   // ...depends on createUsers worker fixture to create each test user
   storageStatePaths: [
+    // eslint-disable-next-line  @typescript-eslint/no-unused-vars
     async ({ browser, browserName, createUsers }, use) => {
```
hallieswan, line 124: `createUsers` is not used, but must be included so that the createUser worker fixture creates each test user before this worker fixture runs. Ignore this type error.

e2e/helpers/ACCESS_TYPE.ts
hallieswan, line 1: Import `ACCESS_TYPE` from @sage-bionetworks/synapse-types  instead

e2e/helpers/acl.ts
```diff
@@ -9,11 +12,15 @@ async function getEntityACL(page: Page, entityId: string, accessToken: string) {
   const acl = await page.evaluate(
     async ({ entityId, accessToken }) => {
       // @ts-expect-error: Cannot find name 'SRC'
-      return await SRC.SynapseClient.getEntityACL(entityId, accessToken)
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+      return (await SRC.SynapseClient.getEntityACL(
+        entityId,
+        accessToken,
+      )) as AccessControlList
```
hallieswan, line 19: Defining an object in SWC that represents SRC would be difficult, so ignore these errors for now since we plan to clean this up [once SRC](https://sagebionetworks.jira.com/browse/SWC-6550) can be called directly from the e2e tests.

e2e/helpers/tables.ts
```diff
@@ -95,15 +100,17 @@ export const enterTableData = async (
...
+  const tableRows = getTableEditorRows(page)
 
   // skip initial empty column
   await expectTableColumnHeaders(tableRows, columnsSchemaConfig, 1)
 
   await test.step('enter test data', async () => {
     const addRowButton = page.getByRole('button', { name: 'Add Row' })
     for (let index = 0; index < tableData.length; index++) {
-      await test.step(`enter test data: ${tableData[index].item}`, async () => {
+      const item = tableData[index].item.toString()
```
hallieswan, line 111: `item` can be a string, string[], or number, so convert to a string before using in the template expression

e2e/helpers/tables.ts
```diff
@@ -239,33 +246,31 @@ const expectTableRowCorrect = async (
         .nth(colIndex)
         .textContent()
       await test.step(`confirm ${columnName}`, async () => {
-        const expectedText =
-          await test.step('format expected text', async () => {
```
hallieswan, line 243: Removed the unnecessary `async`, which resulted in whitespace changes. No other changes in this chunk

e2e/helpers/types.ts
```diff
@@ -1,12 +1,4 @@
-import ACCESS_TYPE from './ACCESS_TYPE'
```
hallieswan, line 1: Types are imported from @sage-bionetworks/synapse-types, rather than duplicated here

tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
```
hallieswan, line 1: Other than the files/paths in "files" and "include", this is almost an exact copy of [`tsconfig.base.json`](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/shared/tsconfig.base.json) in synapse-react-client. I'm not very familiar with TypeScript compiler options, so if there are more appropriate options, happy to adjust this as needed.

.eslintrc.js
nickgros: Do you think we should create a ticket to enable this rule in synapse-web-monorepo? I think it makes sense since we're writing tests in essentially the same way

e2e/files.spec.ts
nickgros: Thanks, this was a catastrophic diff

e2e/fixtures/authenticatedUserPages.ts
nickgros: It's not intuitive to me why `createUsers` needs to be destructured in the signature to accomplish different behavior. Could you explain or point me to some documentation? And also leave a comment alongside the comment used to disable the rule?

e2e/helpers/acl.ts
nickgros: This is still primarily blocked by ESM/CJS mismatch issues, correct?

nickgros (approved):
This is a great improvement, just had one question/request to add a comment in code.

.eslintrc.js
hallieswan: Agreed that we're writing tests in the same way! I initially enabled the rule for the entire monorepo, but since this caused many errors and I wasn't sure whether the rule should apply to non-e2e-test code, I ended up setting the rule as [an override](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/41097cf35073d7f899929506b717afd37c2ca926/.eslintrc.cjs#L63) only for the Portals e2e tests directory. Happy to create a ticket if we'd like to enable the rule for the entire monorepo though!

e2e/fixtures/authenticatedUserPages.ts
hallieswan: Sure! The destructured fixtures are the dependencies of the current fixture, as described in the [Execution Order docs](https://playwright.dev/docs/test-fixtures#execution-order). Since the test users must be created before the test user can be authenticated, the `storageStatePaths` worker fixture depends on the `createUsers` worker fixture. This tells Playwright to set up `createUsers` before executing `storageStatePaths`. I'll add a comment to clarify in the code!

e2e/helpers/acl.ts
hallieswan: Yes, importing SRC directly is blocked by ESM/CJS issues. However, we'd also considered  moving SynapseClient into its own package that could run in browser or node.js environments, so that we wouldn't have to resolve the ESM/CJS issues for all of SRC. I think this work was blocked by the OpenAPI work.

.eslintrc.js
nickgros: I think it would be good to create the ticket since we can use the `void` keyword to be explicit about the places where we don't want to handle the promise. Do you mind creating it?

hallieswan: Sure! https://sagebionetworks.jira.com/browse/PORTALS-2894

PR #5253, opened 2024-01-11
SWC-6638: show both certification and validation badges on profile pages

hallieswan:
- depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/649
- requires SRC bump

type | profile
:---:|:---:
registered | SWC_registered
certified | SWC_certified
certified and validated | SWC_certified_validated SWC_tooltip

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
@@ -431,6 +431,7 @@ public class UserProfileEditorWidgetViewImpl
           accountTypeContainer.setVisible(showAccountType);
         }
 
+        // error will be handled/shown by React component
         @Override
         public void onFailure(Throwable caught) {}
       }
--- a/src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
@@ -1,8 +1,12 @@
 package org.sagebionetworks.web.client.widget.profile;
 
+import static org.sagebionetworks.web.client.presenter.ProfilePresenter.IS_CERTIFIED;
+import static org.sagebionetworks.web.client.presenter.ProfilePresenter.IS_VERIFIED;
+
 import com.google.gwt.event.dom.client.KeyDownHandler;
 import com.google.gwt.uibinder.client.UiBinder;
 import com.google.gwt.uibinder.client.UiField;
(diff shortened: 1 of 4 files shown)


src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
```diff
@@ -388,13 +400,14 @@ public void setSynAlert(IsWidget w) {
...
       React.createElementWithSynapseContext(
-        SRC.SynapseComponents.AccountLevelBadge,
-        AccountLevelBadgeProps.create(userId),
+        SRC.SynapseComponents.AccountLevelBadges,
+        AccountLevelBadgesProps.create(userId),
         propsProvider.getJsInteropContextProps()
       );
-    accountLevelBadgeContainer.render(accountLevelBadgeComponent);
+    accountLevelBadgesContainer.render(accountLevelBadgesComponent);
+    setAccountTypeVisibility(Long.parseLong(userId));
```
hallieswan, line 410: When the user doesn't have any badges (i.e. not certified nor validated), the Account Type row should be hidden

src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
```diff
@@ -405,6 +418,25 @@ public void setOwnerId(String userId) {
...
+      new AsyncCallback<UserBundle>() {
+        @Override
+        public void onSuccess(UserBundle bundle) {
+          boolean showAccountType =
+            bundle.getIsCertified() | bundle.getIsVerified();
+          accountTypeContainer.setVisible(showAccountType);
+        }
+
+        @Override
+        public void onFailure(Throwable caught) {}
```
hallieswan, line 436: If an error occurs when getting the UserBundle, the `AccountLevelBadges` component will display an error. Since the `accountTypeContainer` defaults to visible, I think that the `AccountLevelBadges` component error will be visible, so we may not need to handle the failure here. Happy to add in handling though, if that'd be preferred!

nickgros: Maybe just add a comment `// error will be handled/shown by React component`

nickgros (approved):

package.json
```diff
@@ -18,7 +18,7 @@
     "react-router-dom": "^5.3.4",
     "react-transition-group": "2.6.0",
     "sass": "^1.63.6",
-    "synapse-react-client": "3.1.70",
+    "synapse-react-client": "3.1.71",
```
hallieswan: Badges after SRC release: 
SWC_certified_validated_after_update
SWC_certified_after_update

PR #5261, opened 2024-01-17
SWC-6651: downgrade colima/lima to resolve intermittent port forwarding issue

hallieswan:


.github/workflows/build-test-e2e.yml
```diff
@@ -67,6 +67,35 @@ jobs:
...
       - uses: actions/checkout@v3
+      - name: Check colima and lima version
+        if: runner.os == 'macos'
+        run: |
+          echo "- runner colima version: $(colima version)" >> $GITHUB_STEP_SUMMARY
+          LIMA_VERSION=$(limactl --version)
+          echo "- runner lima version: ${LIMA_VERSION}" >> $GITHUB_STEP_SUMMARY
+          if [[ !("${LIMA_VERSION}" == "limactl version 0.19.0" || "${LIMA_VERSION}" == "limactl version 0.19.1") ]]; then 
+            echo "::warning::runner lima version has changed. Consider removing lima/colima downgrade, see SWC-6655."
+          fi
```
hallieswan, line 78: Add a warning if lima version included in GitHub macOS runner changes

package.json
```diff
@@ -45,7 +45,7 @@
   },
   "scripts": {
     "build": "mvn -B package --file pom.xml",
-    "docker:start": "docker run --name swc-tomcat -d --rm -p 8888:8080 -v \"/$(pwd)/target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war\" -v \"/$(pwd)/e2e_workflow/settings.xml\":/root/.m2/settings.xml tomcat:9.0",
+    "docker:start": "docker pull tomcat:9.0; docker run --name swc-tomcat -d --rm -p 8888:8080 -v \"/$(pwd)/target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war\" -v \"/$(pwd)/e2e_workflow/settings.xml\":/root/.m2/settings.xml tomcat:9.0",
```
hallieswan, line 48: Pull latest tomcat:9.0 image before starting the container to prevent the image used locally and in CI from significantly diverging

playwright.config.ts
```diff
@@ -90,8 +90,11 @@ export default defineConfig({
     // },
   ],
   webServer: {
-    command: process.env.CI ? '' : 'mvn gwt:run',
+    command: process.env.CI
+      ? 'echo SWC not available at baseURL'
+      : 'mvn gwt:run',
```
hallieswan, line 95: Before running e2e tests, Playwright makes a request to the baseURL to determine if the server is ready. If the server is not ready, Playwright attempts to run the `command` to start a new server. Passing an empty string here resulted in an obscure error (`TypeError: The argument 'file' cannot be empty. Received ''`); instead, output a more useful message for troubleshooting in CI.

.github/workflows/build-test-e2e.yml
nickgros: Good idea!

.github/workflows/build-test-e2e.yml
```diff
@@ -99,10 +122,34 @@ jobs:
...
+              exit 1
+          fi
+          if [[ "$(docker exec "${{ env.CONTAINER_NAME }}" bash -c "curl -sS -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8080")" != "200" ]]; then
+            echo "::error:: SWC container port not accessible from container"
+            exit 1
+          fi
+          if [[ "$(docker exec "${{ env.CONTAINER_NAME }}" bash -c "curl -sS -o /dev/null -w '%{http_code}\n' http://172.17.0.1:8888")" != "200" ]]; then
+            echo "::error:: SWC host port not accessible from container"
+            exit 1
+          fi
```
nickgros, line 146: Are these the same?

.github/workflows/build-test-e2e.yml
```diff
@@ -99,10 +122,34 @@ jobs:
         run: yarn install --frozen-lockfile
       - name: Install Playwright Browsers
         run: yarn playwright install --with-deps
-      - name: Confirm Docker container is running
+      - name: Check for common errors
         run: |
```
nickgros, line 126: At some point maybe we just put this script into a file and run it. I think it's fine in this action for now

nickgros (approved):
Thanks for the effort to troubleshoot this!

.github/workflows/build-test-e2e.yml
hallieswan: They're a bit different -- the first attempts to connect to port 8080 on the local loopback address of the container itself - `http://127.0.0.1:8080`. The second attempts to connect to port 8888 on the Docker host from within the container - `http://172.17.0.1:8888`. If both return a 200, then SWC is accessible at both expected ports from within the container.

.github/workflows/build-test-e2e.yml
hallieswan: Good call!

PR #5264, opened 2024-01-22
SWC-6658:  Front-end support for re-taking certification quiz

jay-hodgson:
having a PassingRecord does not mean user is currently certified


nickgros (approved):
Looks good! This should just work when we get the backend change! 🤞

PR #5267, opened 2024-01-24
SWC-6645

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
```diff
@@ -89,6 +91,15 @@ public void setPlace(Search place) {
     String queryTerm = place.getSearchTerm();
     if (queryTerm == null) queryTerm = "";
     currentSearch = checkForJson(queryTerm);
+    if (currentSearch.getQueryTerm().size() == 1) {
```
jay-hodgson, line 94: If there's a single search term, check to see if we should redirect (to Synapse entity page or People Search)

src/main/java/org/sagebionetworks/web/client/presenter/SearchUtil.java
```diff
@@ -34,6 +34,9 @@ public static Place willRedirect(Search place) {
    * @return
    */
   public static Place willRedirect(String queryTerm) {
+    if (queryTerm == null || queryTerm.trim().length() == 0) {
```
jay-hodgson, line 37: Precondition check

src/main/java/org/sagebionetworks/web/client/presenter/TeamSearchPresenter.java
```diff
@@ -79,6 +79,11 @@ public void setPlace(TeamSearch place) {
     this.place = place;
     this.view.setPresenter(this);
     searchTerm = place.getSearchTerm();
+    Place redirectPlace = SearchUtil.willRedirect(searchTerm);
```
jay-hodgson, line 82: Do the same for Team Search

src/main/java/org/sagebionetworks/web/client/presenter/TeamSearchPresenter.java
```diff
@@ -98,25 +103,26 @@ public void goTo(Place place) {
   public void loadMore() {
     synAlert.clear();
     // execute search, and update view with the results
-    AsyncCallback<PaginatedResults<Team>> callback = new AsyncCallback<PaginatedResults<Team>>() {
```
jay-hodgson, line 101: No code change, auto-formatted on update

nickgros (approved):
Looks good!

PR #5269, opened 2024-01-26
SWC-6657

jay-hodgson:


xschildw (approved):
Thanks @jay-hodgson for the clarifications!

PR #5284, opened 2024-02-06
SWC-6631: entity updated event can cause place change.  On save, we a…

jay-hodgson:
…re no longer editing

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
@@ -682,8 +682,9 @@ public class TableEntityWidgetV2
           DisplayUtils.NotificationVariant.SUCCESS,
           toastOptions
         );
-        eventBus.fireEvent(new EntityUpdatedEvent(tableId));
+        ginInjector.getGlobalApplicationState().setIsEditing(false);
         closeItemsEditor();
+        eventBus.fireEvent(new EntityUpdatedEvent(tableId));
       },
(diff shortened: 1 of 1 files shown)


src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
```diff
@@ -682,8 +682,9 @@ public DatasetEditorProps getItemsEditorProps() {
           DisplayUtils.NotificationVariant.SUCCESS,
           toastOptions
         );
-        eventBus.fireEvent(new EntityUpdatedEvent(tableId));
```
jay-hodgson, line 685: Moved this line to the last thing that we do.

src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
```diff
@@ -682,8 +682,9 @@ public DatasetEditorProps getItemsEditorProps() {
...
+        ginInjector.getGlobalApplicationState().setIsEditing(false);
```
jay-hodgson, line 685: Explicitly clear the editing state on save.

src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
jay-hodgson: The EntityUpdatedEvent can cause a place change (which is recognized as a no-op later in the process).

nickgros (approved):

PR #5286, opened 2024-02-07
PORTALS-2818

nickgros:
- While in edit mode with annotations editor v2, toggle globalAppState.setIsEditing


jay-hodgson (approved):

nickgros:
This doesn't really require https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/699, the new prop will just do nothing until synapse-react-client is updated to include that change. So this can be merged before that change is pulled here.

PR #5288, opened 2024-02-08
PORTALS-2905: include changes affecting SWC from synapse-react-client by renaming to defaultShowPlots

jinjunoh:
Implemented changes on swc to reflect the name changes made on synapse-react-client


jay-hodgson:
Does this depend on a particular release of the react client library?

jay-hodgson:
Does it make sense to check in IntelliJ iml files into the codebase?  If not, they should be removed from the PR, and probably added to .gitignore

jay-hodgson (changes requested):
Added a couple of questions that I'd like to be answered before merge.  Changing the prop name is fine though!

jinjunoh:
> Does this depend on a particular release of the react client library?

jay-hodgson:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/682

jay-hodgson (approved):
Approved, but not to be merged until we are pointing to a compatible SRC release

nickgros:
This should have been pulled into 488, changing the base branch and merging

PR #5293, opened 2024-02-15
PORTALS-2943: Sage branding update

jay-hodgson:
Switch from Lato to DM Sans


jay-hodgson:
And update the logo in the footer:
Screen Shot 2024-02-15 at 3 17 09 PM

src/main/java/org/sagebionetworks/web/client/widget/footer/FooterViewImpl.java
```diff
@@ -68,6 +70,12 @@ public interface Binder extends UiBinder<Widget, FooterViewImpl> {}
   @UiField
   Image alphaModeStatusIcon;
 
+  @UiField
+  FocusPanel sageLogo1;
+
+  @UiField
+  FocusPanel sageLogo2;
```
nickgros, line 77: Do you mind renaming these so it's more clear that they are conditionally shown based on breakpoints?

nickgros (approved):

PR #5296, opened 2024-02-19
SWC-6695

jay-hodgson:


jay-hodgson:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/721 (as well as releasing a new SRC and pulling it into this PR).

nickgros (approved):

PR #5300, opened 2024-02-22
SWC-6545 - Upgrade to @tanstack/react-query v5

nickgros:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/739 and a new version of synapse-react-client

- Fetch @tanstack/react-query's ESM distribution from esm.sh, add esm.sh to CSP
- Account for react-query's breaking changes
- Remove headjs, modern browsers have been able to load scripts in parallel for about 15 years


package.json
```diff
@@ -14,7 +14,6 @@
     "react-dom": "18.2.0",
     "react-measure": "2.5.2",
     "react-plotly.js": "^2.6.0",
-    "react-query": "^3.39.3",
```
nickgros, line 17: We will load @tanstack/react-query v5 through a CDN. [See comment in Jira](https://sagebionetworks.jira.com/browse/SWC-6545?focusedCommentId=199010).

src/main/java/org/sagebionetworks/web/client/jsinterop/KeyFactory.java
```diff
@@ -3,14 +3,15 @@
...
+import org.sagebionetworks.web.client.jsinterop.reactquery.QueryKey;
 
 @JsType(isNative = true, namespace = "SRC.SynapseQueries")
 public class KeyFactory {
 
   @JsConstructor
   public KeyFactory(String accessToken) {}
 
-  public native List<?> getDownloadListBaseQueryKey();
+  public native QueryKey getDownloadListBaseQueryKey();
```
nickgros, line 14: Small improvement to our JsInterop types/classes - Our `KeyFactory` now returns a `QueryKey`

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/InvalidateQueryFilters.java
```diff
@@ -0,0 +1,18 @@
+package org.sagebionetworks.web.client.jsinterop.reactquery;
+
+import jsinterop.annotations.JsOverlay;
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+
+@JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
+public class InvalidateQueryFilters {
```
nickgros, line 8: Breaking change in react-query -- we need to put our `queryKey` in an `InvalidateQueryFilters` object when we invalidate queries.

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryClient.java
```diff
@@ -13,15 +13,9 @@ public QueryClient(QueryClientOptions config) {}
    */
   public native void resetQueries();
 
-  /**
-   * Removes the cached query data and triggers a refetch.
-   * @param queryKey
-   */
-  public native void resetQueries(List<?> queryKey);
```
nickgros, line 20: Unused

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryClient.java
```diff
@@ -13,15 +13,9 @@ public QueryClient(QueryClientOptions config) {}
...
-
   /**
    * Triggers a refetch of matching query data without removing the cached data.
-   * @param queryKey
+   * @param filters
    */
-  public native void invalidateQueries(List<?> queryKey);
+  public native void invalidateQueries(InvalidateQueryFilters filters);
```
nickgros, line 20: Handle breaking change in this method

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryKeyConstants.java
nickgros, line 1: Class was unused, because we have been using the `KeyFactory` exported by synapse-react-client

src/main/java/org/sagebionetworks/web/client/jsinterop/reactquery/QueryOptions.java
```diff
@@ -8,15 +8,15 @@
...
+  public long gcTime;
   public boolean retry;
   public boolean refetchOnWindowFocus;
 
   @JsOverlay
   public static QueryOptions create() {
     QueryOptions queryOptions = new QueryOptions();
     queryOptions.staleTime = 60 * 1000L; // 60s
-    queryOptions.cacheTime = 1000L * 60 * 30; // 30 min
+    queryOptions.gcTime = 1000L * 60 * 30; // 30 min
```
nickgros, line 19: Breaking change in react-query -- same functionality, new name

src/main/webapp/Portal.html
```diff
@@ -11,7 +11,7 @@
 
     <meta
       http-equiv="Content-Security-Policy"
-      content="default-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' *.synapse.org *.sagebase.org *.googleapis.com *.gstatic.com *.googletagmanager.com cdn.statuspage.io s3.amazonaws.com www.google-analytics.com kh896k90gyvg.statuspage.io *.youtube.com *.rstudio.com *.google.com *.shinyapps.io *.smartsheet.com sagebionetworks.jira.com 127.0.0.1:9876; img-src * data: blob:; connect-src https://* http://127.0.0.1:8888"
+      content="default-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' *.synapse.org *.sagebase.org *.googleapis.com *.gstatic.com *.googletagmanager.com cdn.statuspage.io s3.amazonaws.com www.google-analytics.com kh896k90gyvg.statuspage.io *.youtube.com *.rstudio.com *.google.com *.shinyapps.io *.smartsheet.com sagebionetworks.jira.com 127.0.0.1:9876 esm.sh; img-src * data: blob:; connect-src https://* http://127.0.0.1:8888"
```
nickgros, line 14: Add the new CDN

src/main/webapp/Portal.html
```diff
@@ -159,382 +159,70 @@
       loadCss('//fonts.googleapis.com/css?family=Lato')
       loadCss('https://fonts.googleapis.com/icon?family=Material+Icons')
 
-      /*! head.load.min.js - v1.0.3 */
```
nickgros, line 164: I removed headjs. I was always curious what `head.load` did in this script, and I learned that it was useful in IE6, IE7, and other browsers of that era where scripts were not necessarily loaded in parallel. This is not something we need to worry about now, so I removed the script.

src/main/webapp/Portal.html
```diff
@@ -159,382 +159,70 @@
...
-        i.ready(r, function () {
-          y() &&
-            u(h.ALL, function (n) {
-              f(n)
-            })
-          i.feature && i.feature('domloaded', !0)
-        })
-      })(window)
-
-      let reactPath, reactDomPath, reactQueryPath, srcPath
```
nickgros, line 475: Remove react-query from this set of loaded scripts, since we'll import it another way

src/main/webapp/Portal.html
```diff
@@ -159,382 +159,70 @@
...
         reactDomPath = cdnEndpoint + 'generated/react-dom.development.js'
-        // reactQueryPath = cdnEndpoint + 'generated/react-query.development.js'
         srcPath = cdnEndpoint + 'generated/synapse-react-client.development.js'
       }
 
       // TODO: Workaround for SWC-6664
       // Override the conditional above and just use the production distribution
-      reactQueryPath = cdnEndpoint + 'generated/react-query.production.min.js'
 
-      head.load(cdnEndpoint + 'generated/swc.css')
```
nickgros, line 494: replace `head.load` with `loadJs` or `loadCss`, as appropriate.

src/main/webapp/Portal.html
```diff
@@ -159,382 +159,70 @@
...
+      loadCss(cdnEndpoint + 'css/katex-0.10.1.min.css')
+      loadJs('//cdn.statuspage.io/se-v2.js')
+      // loadJs("//kh896k90gyvg.statuspage.io/embed/script.js");
+      loadJs(cdnEndpoint + 'generated/papaparse.min.js')
+    </script>
+    <script type="module">
+      // react-query only ships ESM. esm.sh is a CDN that will 'bundle' dependencies for us
+      import * as ReactQuery from 'https://esm.sh/@tanstack/react-query@5.22.2'
+      window.ReactQuery = ReactQuery
     </script>
```
nickgros, line 228: Load ReactQuery from the CDN, and store it in the `window.ReactQuery` variable just as the UMD version did.

src/main/webapp/Portal.html
jay-hodgson: Are we able to be more specific with esm.sh?  Something like `esm.sh/@tanstack/react-query*` ?

jay-hodgson (approved):

src/main/webapp/Portal.html
nickgros: Unfortunately not. The CSP only works at the host level-even if we could be more specific, the ESM bundle fetches dependencies of @tanstack/react-query from other paths of esm.sh, which we wouldn't want to block.

@jay-hodgson Is this a blocker for this change? I can look into adding a bundler before merging this.

PR #5304, opened 2024-02-23
SWC-6699: On ORCiD login SSO failure, send back to login place with special token

jay-hodgson:


jay-hodgson:
Screen Shot 2024-02-23 at 3 09 26 PM

src/main/java/org/sagebionetworks/web/server/servlet/oauth2/OAuth2SessionServlet.java
```diff
@@ -90,8 +89,12 @@ public void validateUser(
       resp.addCookie(cookie);
       resp.sendRedirect(LOGIN_PLACE + WebConstants.REDIRECT_TO_LAST_PLACE);
     } catch (SynapseNotFoundException e) {
```
jay-hodgson, line 91: If not found (associated to a Synapse user account) - if ORCID, then send back to the login page with a special token that causes the new error to pop up.  If anything else (aka Google right now), then do what it's always done and redirect to the registration page (where they can click sign up with Google!).

xschildw (approved):

PR #5317, opened 2024-03-06
SWC-6704

jay-hodgson:


jay-hodgson:
Did not encounter any issues building or running the client with the new version.

nickgros (approved):

PR #5321, opened 2024-03-09
SWC-6723: clean up end to end tests

hallieswan:
- SWC-6723: change `test` -> `testAuth` when running a test using the authenticated user fixtures, which resolves the error introduced in Playwright v1.42
- fix empty table expectation so that tables test passes

### DIFF
--- a/e2e/homepage.spec.ts
+++ b/e2e/homepage.spec.ts
@@ -2,7 +2,7 @@ import { expect, test } from '@playwright/test'
 import { testAuth } from './fixtures/authenticatedUserPages'
 import { waitForInitialPageLoad } from './helpers/utils'
 
-test.describe('Homepage', () => {
+test.describe('Homepage - Unauthenticated', () => {
   test('should show Log In To Synapse button when logged out', async ({
     page,
   }) => {
@@ -16,7 +16,9 @@ test.describe('Homepage', () => {
       page.getByRole('link', { name: 'View Your Dashboard' }),
     ).toHaveCount(0)
   })
+})
 
+testAuth.describe('Homepage - Authenticated', () => {
   testAuth(
     'should show View Your Dashboard button when logged in',
     async ({ userPage }) => {
--- a/e2e/tables.spec.ts
+++ b/e2e/tables.spec.ts
@@ -1,4 +1,4 @@
-import { expect, test } from '@playwright/test'
+import { expect } from '@playwright/test'
 import { testAuth } from './fixtures/authenticatedUserPages'
 import { deleteEntity, generateEntityName } from './helpers/entities'
 import {
@@ -44,13 +44,13 @@ let tableEntityId: string | undefined = undefined
 const noTablesText =
   'There are no Tables associated with this project. Any Tables you create in this project will appear here'
 
-test.describe('Tables', () => {
+testAuth.describe('Tables', () => {
   testAuth.beforeAll(async ({ browser, storageStatePaths }) => {
     userProject = await setupProject(browser, 'swc-e2e-user', storageStatePaths)
   })
 
   testAuth.afterAll(async ({ browser }) => {
-    test.slow()
+    testAuth.slow()
     if (tableEntityId) {
       const page = await browser.newPage()
       await deleteEntity(tableEntityId, true, getAdminPAT(), page)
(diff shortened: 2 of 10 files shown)


e2e/homepage.spec.ts
```diff
@@ -2,7 +2,7 @@ import { expect, test } from '@playwright/test'
 import { testAuth } from './fixtures/authenticatedUserPages'
 import { waitForInitialPageLoad } from './helpers/utils'
 
-test.describe('Homepage', () => {
+test.describe('Homepage - Unauthenticated', () => {
```
hallieswan, line 5: Separate these tests into different suites, so the unauthenticated suite can still be run without setting up the authenticated user page fixtures

e2e/tables.spec.ts
```diff
@@ -129,21 +129,24 @@ test.describe('Tables', () => {
...
       await expect(userPage.getByText(noTablesText)).toBeVisible()
     })
 
     // TODO: remove this step after table descriptions are visible outide of experimental mode
-    await test.step('Toggle into Experimental Mode to use react-based table column schema editor', async () => {
-      await toggleIntoExperimentalMode(userPage)
-      await expectTablesPageLoaded(userPage, userProject.id)
-      await expect(userPage.getByText(noTablesText)).toBeVisible()
-    })
+    await testAuth.step(
```
hallieswan, line 139: The rest of the changes in this file are 1) changing `test` -> `testAuth` and 2) associated whitespace/formatting changes.

e2e/teams.spec.ts
```diff
@@ -52,9 +52,9 @@ let teamId: string | undefined = undefined
 // Run multiple describes in parallel, but run tests inside each describe in order
 // ...tests within describe expect afterAll to be run with the same users, i.e. on the same worker
 // https://playwright.dev/docs/api/class-test#test-describe-configure
-test.describe.configure({ mode: 'serial' })
+testAuth.describe.configure({ mode: 'serial' })
```
hallieswan, line 55: Same in this file -- only changes are `test` -> `testAuth` and formatting changes

.github/workflows/build-test-e2e.yml
```diff
@@ -67,35 +67,6 @@ jobs:
...
-
-          # install colima
-          COLIMA_VERSION="v0.6.6"
-          COLIMA_VERSION_URL="https://github.com/abiosoft/colima/releases/download/${COLIMA_VERSION}/colima-$(uname -s)-$(uname -m)"
-          sudo mkdir -p /usr/local/bin
-          sudo curl -L -o /usr/local/bin/colima "${COLIMA_VERSION_URL}" && sudo chmod +x /usr/local/bin/colima
-
-          # Get colima and lima versions
-          limactl --version
-          colima version
```
hallieswan: SWC-6655: fixed versions of lima are used in the macOS GitHub runners, so this is no longer needed.

e2e/tables.spec.ts
```diff
@@ -157,48 +160,49 @@ test.describe('Tables', () => {
...
-        ).toBeVisible()
-        await expect(
-          userPage.getByText('This table is currently empty'),
-        ).toBeVisible()
-      })
+        await testAuth.step('table is currently empty', async () => {
+          await expect(
+            userPage.getByRole('heading', { name: 'Items (0)' }),
+          ).toBeVisible()
+          await expect(userPage.getByText('0 total rows')).toBeVisible()
```
hallieswan: Also change the expectation for an empty table in Synapse so the tables test passes -- a table with 0 rows is shown instead of text stating that "This table is currently empty": 

image

e2e/tables.spec.ts
hallieswan: ~~*Other than the specific change to the empty table expectation, as noted below~~ -- reverted to the original expectation, as the new state indicates a bug rather than a design choice.

.github/workflows/build-test-e2e.yml
hallieswan: E2E tests intermittently failed to run after removing the downgrade, even though lima was updated. Pushed that ticket back to open for more investigation and re-introduced the downgrade.

e2e/tables.spec.ts
nickgros: Is this expected? This seems like a new bug

hallieswan: Good point! I've reverted to the original expectation and opened a [new ticket](https://sagebionetworks.jira.com/browse/SWC-6725) to track!

hallieswan: We'll expect the tables test to fail until that ticket is resolved.

nickgros (approved):

PR #5329, opened 2024-03-18
Release-493 patch with SRC-3.2.6

jay-hodgson:


jay-hodgson:
https://github.com/Sage-Bionetworks/synapse-web-monorepo/releases/tag/synapse-react-client%2Fv3.2.6

xschildw (approved):

PR #5332, opened 2024-03-19
SWC-6670: use React-based editor for SubmissionViewScope

jinjunoh:


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityViewScopeEditorModalWidget.java
jinjunoh, line 1: deleted unused imports

nickgros (approved):
LGTM!

PR #5338, opened 2024-03-21
SWC-6746: just render paragraphs when not editing

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
```diff
@@ -182,17 +201,6 @@ public UserProfileEditorWidgetViewImpl(
     locationTextBox = locationSuggestBox.getTextBox();
     locationTextBox.addStyleName("form-control");
     locationSuggestBoxContainer.add(locationSuggestBox);
-    // note, not adding email since it's not editable here.
-    textBoxes =
```
jay-hodgson, line 186: Previously, this code would run through this array of text boxes and set them to readOnly if we were not editing.  Now, there are "renderer" (static, read-only) elements that are shown instead.

src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
```diff
@@ -215,6 +223,7 @@ public Widget asWidget() {
   @Override
   public void setUsername(String username) {
     this.username.setText(username);
+    usernameRenderer.setText(username);
```
jay-hodgson, line 226: Initialize both the editor and renderer with the current value

src/main/java/org/sagebionetworks/web/client/widget/profile/UserProfileEditorWidgetViewImpl.java
```diff
@@ -395,33 +409,46 @@ public void addKeyDownHandlerToFields(KeyDownHandler keyDownHandler) {
...
     currentPosition.setPlaceholder(isEditing ? "Enter current position" : "");
+    industry.setVisible(isEditing);
+    industryRenderer.setVisible(!isEditing);
     industry.setPlaceholder(isEditing ? "Enter industry/discipline" : "");
 
     editProfileButton.setVisible(!isEditing);
     saveProfileButton.setVisible(isEditing);
     cancelButton.setVisible(isEditing);
-    linkRenderer.setVisible(!isEditing);
-    link.setVisible(isEditing);
```
jay-hodgson, line 422: Moved up (next to where we set the placeholder) in the same method

hallieswan (approved):
Looks good to me!

PR #5341, opened 2024-03-28
SWC-6256 - Upgrade GWT to v2.11.0

nickgros:


pom.xml
```diff
@@ -497,6 +497,12 @@
       <groupId>com.google.gwt.eventbinder</groupId>
       <artifactId>eventbinder</artifactId>
       <version>1.1.0</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.gwt</groupId>
+          <artifactId>gwt-user</artifactId>
+        </exclusion>
+      </exclusions>
```
nickgros, line 505: A few of our dependencies transitively pull in `com.google.gwt:gwt-user`. We want them to instead use `org.gwtproject:gwt-user`, so we must manually exclude the `com.google.gwt:gwt-user` dependency.

pom.xml
```diff
@@ -199,7 +199,7 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>gwt-maven-plugin</artifactId>
-        <version>${gwtVersion}</version>
+        <version>2.10.0</version>
```
nickgros, line 202: v2.11.0 does not exist for this plugin.

jay-hodgson (approved):

PR #5342, opened 2024-04-01
SWC-6755 - Fix null storage location Long value when uploading to a wiki

nickgros:


src/main/java/org/sagebionetworks/web/client/widget/upload/MultipartUploaderImplV2.java
```diff
@@ -76,7 +90,7 @@ public void uploadFile(
       auth.getCurrentUserAccessToken(),
       fileName,
       blob,
-      storageLocationId.intValue(),
+      storageLocationIntValue,
```
nickgros, line 93: When the bug occurred, `storageLocationId` was null, so `storageLocationId.intValue()` yielded a NPE.

The SRC code that is called is not set up to accept `null`, so we have to supply the default storage location id here

xschildw (changes requested):
Looks good but there's a compile error, missing import in MultipartUploaderImplV2Test.

jay-hodgson (commented):
Thank you for tracking this down and deploying the fix Nick and Xa!  Sorry we didn't catch this last week

PR #5347, opened 2024-04-03
SWC-6145 - markdown-it & plugins pulled from npm

nickgros:


pom.xml
```diff
@@ -485,6 +485,71 @@
                   file="${project.basedir}/node_modules/papaparse/papaparse.min.js"
                   tofile="src/main/webapp/generated/papaparse.min.js"
                 />
+                <copy todir="src/main/webapp/generated/katex">
+                  <fileset dir="${project.basedir}/node_modules/katex/dist" />
+                </copy>
```
nickgros, line 507: Copy all of the katex/dist folder because it contains fonts referenced in the CSS

jay-hodgson:
This looks great, but there's a regression.  Namely, https://sagebionetworks.jira.com/browse/PORTALS-1974 when running this branch locally.
Screen Shot 2024-04-18 at 1 45 13 PM

nickgros:
@jay-hodgson nice catch! I was able to reproduce the issue. It even persisted after downgrading all of the markdown-it related packages to the version referenced in the originally-committed files. So more investigation is needed

nickgros:
PORTALS-1974 regression fixed locally: 
image

jay-hodgson (approved):

PR #5349, opened 2024-04-05
SWC-6759 - fix case where href attr was not sanitized

nickgros:


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -697,21 +696,25 @@ private static final native boolean initFilterXss() /*-{
...
-					} else {
-						return value;
-					}
-				}
-			};
-			$wnd.xss = new $wnd.filterXSS.FilterXSS(options);
-			return true;
+                safeAttrValue: function(tag, name, value) {
+                  // Apply default safeAttrValue filtering:
+                  value = $wnd.filterXSS.safeAttrValue(tag, name, value);
```
nickgros, line 701: By default `safeAttrValue` cleans the `href` attribute. But because we overrode it, we missed that case. Use the default `safeAttrValue` first to clean href, then apply our custom rules.

I did not notice any regressions in some light testing, but I would rather aggressively filter and fix those cases if they come up.

src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtilsImpl.java
```diff
@@ -697,21 +696,25 @@ private static final native boolean initFilterXss() /*-{
...
+                    ) {
+                      return ''
+                    }
+                  }
+                  value = $wnd.filterXSS.escapeAttrValue(value)
+                  return value
+                }
+              }
+			$wnd.xss = new $wnd.filterXSS.FilterXSS(options)
+			return true
```
nickgros, line 717: Mind the whitespace gap. We have the same code in synapse-react-client (for portals). Lets soon plan to import that config here so we just have a single config to update over time.

hallieswan (approved):

xschildw (approved):

PR #5354, opened 2024-04-10
SWC-6331: provide client-side error when renaming non-draft datasets

jinjunoh:
On Production
Screenshot 2024-04-10 at 11 07 44 AM

On Feature
Screenshot 2024-04-10 at 10 58 41 AM

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -310,15 +310,16 @@ public class EntityActionControllerImpl
     this.gwt = gwt;
     this.eventBus = eventBus;
     this.popupUtils = popupUtilsView;
-    entityUpdatedWizardCallback = new WizardCallback() {
-      @Override
-      public void onFinished() {
-        fireEntityUpdatedEvent();
-      }
+    entityUpdatedWizardCallback =
+      new WizardCallback() {
+        @Override
+        public void onFinished() {
+          fireEntityUpdatedEvent();
+        }
 
-      @Override
-      public void onCanceled() {}
-    };
+        @Override
+        public void onCanceled() {}
+      };
   }
 
   private void fireEntityUpdatedEvent() {
@@ -576,9 +577,8 @@ public class EntityActionControllerImpl
     this.entity = entityBundle.getEntity();
     this.isUserAuthenticated = authenticationController.isLoggedIn();
     this.isCurrentVersion = isCurrentVersion;
-    this.entityTypeDisplay = getFriendlyEntityTypeName(
-      entityBundle.getEntity()
-    );
+    this.entityTypeDisplay =
+      getFriendlyEntityTypeName(entityBundle.getEntity());
     this.currentArea = currentArea;
     this.addToDownloadListWidget = addToDownloadListWidget;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1544,7 +1540,7 @@ private void configureEditFileMetadataAction() {
     }
   }
 
-  private void configureRenameAction() {
+  private void configureRenameAction() { ////
```
nickgros:
```suggestion
  private void configureRenameAction() {
```

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2609,22 +2599,34 @@ public void onFailure(Throwable caught) {
   }
 
   private void onRename() {
-    checkUpdateEntity(() -> {
-      postCheckRename();
-    });
+    if (
+      entityBundle.getEntity() instanceof VersionableEntity &&
+      ((VersionableEntity) entityBundle.getEntity()).getIsLatestVersion()
+    ) {
```
nickgros: Can you add these new checks to the `checkUpdateEntity` method?

nickgros (changes requested):

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2609,22 +2599,38 @@ public void onFailure(Throwable caught) {
...
+
+  private boolean checkIsLatestVersion() {
+    if (
+      entityBundle.getEntity() instanceof VersionableEntity &&
+      ((VersionableEntity) entityBundle.getEntity()).getIsLatestVersion()
+    ) {
+      return true;
+    } else {
+      return false;
+    }
```
nickgros: Are you still able to rename non-VersionableEntities like Folders?

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2609,22 +2599,38 @@ public void onFailure(Throwable caught) {
   }
 
   private void onRename() {
-    checkUpdateEntity(() -> {
-      postCheckRename();
-    });
+    if (checkIsLatestVersion()) {
```
nickgros, line 2629: Can you add a test to verify that the error message is shown for datasets, and consider adding tests for any other cases that were missed (such as renaming a folder if your change caused that to break without a test failing)

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2614,9 +2614,30 @@ public void onFailure(Throwable caught) {
...
+      checkUpdateEntity(() -> {
+        postCheckRename();
+      });
+    } else {
+      view.showErrorMessage(
+        "Can only change the name of the most recent " +
+        entityBundle.getEntityType() +
+        " version."
+      );
```
nickgros: Now that I think about it, this will be confusing for datasets, because our terminology is different for those entities.

This text is fine for non-datasets, but if the entity is an instanceof Dataset, can you show an error message that says `"Can only change the name of the draft dataset"` ?

nickgros (approved):

PR #5356, opened 2024-04-15
SWC-6738: bump spring-web to 5.3.33

jay-hodgson:


jay-hodgson:
Verified that the client builds and runs without any issues with this change

hallieswan (approved):

PR #5365, opened 2024-04-22
Update to SRC 3.2.9

jay-hodgson:


src/main/webapp/Portal.html
```diff
@@ -214,7 +217,7 @@
       )
       loadJs(cdnEndpoint + 'js/linkify.min.js')
       loadJs(cdnEndpoint + 'js/linkify-string.min.js')
-      loadCss(cdnEndpoint + 'css/katex-0.10.1.min.css')
```
hallieswan, line 217: Could [this file](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/develop/src/main/webapp/css/katex-0.10.1.min.css) be removed or is it still used elsewhere?

hallieswan (approved):
Looks good to me! 

I also noticed that [katex-0.10.1](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/develop/src/main/webapp/js/katex-0.10.1.min.js) is specified in [ClientProperties](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/490c0e0836a22c6a5155d9db1f30e1fb3ec5f0dc/src/main/java/org/sagebionetworks/web/client/ClientProperties.java#L70-L72) and loaded by [MarkdownWidget](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/490c0e0836a22c6a5155d9db1f30e1fb3ec5f0dc/src/main/java/org/sagebionetworks/web/client/widget/entity/MarkdownWidget.java#L145-L178) -- if it's out of scope for this PR, should we open a ticket to track removing this older version of katex, so that SWC is only using the katex version noted in package.json?

jay-hodgson:
Verified latex is still rendered after the changes.
Screen Shot 2024-04-23 at 9 45 41 AM

hallieswan (approved):

PR #5373, opened 2024-04-26
SWC-6810

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/AccessRequirementPresenter.java
```diff
@@ -59,37 +60,44 @@ public void setPlace(AccessRequirementPlace place) {
     String id = place.getParam(AccessRequirementsPlace.ID_PARAM);
     String typeString = place.getParam(AccessRequirementsPlace.TYPE_PARAM);
 
-    // Note: configuring the Access Requirement widget without a target subject will result in notifications sent to the user will not have the context (Project/Folder/File associated with the restriction).
```
jay-hodgson, line 62: This diff is terrible.  Basically, this component now always queries for the AccessRequirement (used to only do this if the subject was not given).  The view title is now based on the Access Requirement name (which did not exist when this presenter was written!).  Also, we can't assume that the subject array has at least one element (with the new annotation based subjects).

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/CreateAccessRequirementButton.java
```diff
@@ -126,11 +127,17 @@ private void useSrcWizard() {
...
-    CreateOrUpdateAccessRequirementWizardProps.OnComplete onComplete = () -> {
-      wizard.setOpen(false);
-      refreshCallback.invoke();
-      view.clearWidgets();
-    };
+    CreateOrUpdateAccessRequirementWizardProps.OnComplete onComplete =
+      accessRequirementID -> {
+        wizard.setOpen(false);
+        view.clearWidgets();
+        AccessRequirementPlace target = new AccessRequirementPlace("");
```
jay-hodgson, line 134: Instead of refreshing the page, go to the AR place (pointing directly to this specific AR ID).

jay-hodgson:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/910

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ACTAccessRequirementWidget.java
```diff
@@ -149,6 +149,7 @@ public void onSuccess(String rootWikiId) {
     convertACTAccessRequirementButton.configure(ar, refreshCallback);
     view.setAccessRequirementName(ar.getName());
     view.setAccessRequirementID(ar.getId().toString());
+    view.setSubjectsDefinedByAnnotations(ar.getSubjectsDefinedByAnnotations());
```
jay-hodgson, line 152: These changes are x4 (one for ACTAccessRequirement (legacy type), ManagedACTAccessRequirement, SelfSignAccessRequirement, and TermsOfUseAccessRequirement (legacy type)

jay-hodgson: Very little shared code between these widgets :/

jay-hodgson:
Moved to point at 498, the plan will be to patch staging next week

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ACTAccessRequirementWidget.java
hallieswan: The other widgets (ManagedACT AR, SelfSign AR, TermsOfUse AR) only `setSubjectsDefinedByAnnotations` if `isACT`. Just curious, why does the ACT AR widget handle setting the `subjectsDefinedByAnnotations` differently than the other AR widgets?

hallieswan (approved):
Looks good to me! Just one question

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ACTAccessRequirementWidget.java
jay-hodgson: Good question.  I think because ACT Access Requirements should no longer be used, so the UI is even lazier (users trying to gain access for this type of AR need to go out-of-band and will see some details in this renderer that may be confusing).  There's a migration path for ACT to move these to the Managed ACT AR type, which Kimyen added many years ago

PR #5378, opened 2024-05-02
SWC-6761: small amount of progress with dependencies

jay-hodgson:


nickgros (approved):
LGTM! Do you know what we use jquery-ui for?

jay-hodgson:
> LGTM! Do you know what we use jquery-ui for?

Unknown, but my guess was the old provenance widget.  Next step could be an attempt to remove that dependency and try to find what breaks!

PR #5381, opened 2024-05-10
Bump synapse-react-client to v3.2.16

nickgros:
Note: target branch is staging. 

If we merge this and patch, then we will need to update the fix-version for the following tickets:
* SWC-6777 
* SWC-6827
* SWC-6775


jay-hodgson (approved):

nickgros:
We decided it is not worth the risk to patch v498. These issues are not especially severe.

PR #5383, opened 2024-05-10
SWC-6765: animate the progress bar and make it striped

jay-hodgson:
Google upload can stall, so update the UI to indicate that things are happening


jay-hodgson:
Merging to kick off the build

PR #5385, opened 2024-05-13
SWC-6838: update Synapse backend artifact, and use AuthenticationController.cle…

jay-hodgson:
…arLocalCache when it needs to be cleared


hallieswan (approved):
Looks good to me!

PR #5386, opened 2024-05-15
SWC-6839

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -644,7 +643,7 @@ public static String getSynapseHistoryToken(
     String entityId,
     Long versionNumber
   ) {
-    return "#" + getSynapseHistoryTokenNoHash(entityId, versionNumber);
+    return getSynapseHistoryTokenNoHash(entityId, versionNumber);
   }
 
   public static String getSynapseHistoryTokenNoHash(
```
nickgros: Since we're completely removing the hash, I would expect each of the `getSynapseHistoryTokenNoHash` to be either renamed or deleted so we just have `getSynapseHistoryToken` methods (unless these are somehow useful to redirect traffic from `#!` URLs)

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -669,7 +668,7 @@ public static String getSynapseHistoryTokenNoHash(
     String areaToken
   ) {
     Synapse place = new Synapse(entityId, versionNumber, area, areaToken);
-    return "!" + getPlaceString(Synapse.class) + ":" + place.toToken();
+    return "/" + getPlaceString(Synapse.class) + ":" + place.toToken();
```
nickgros, line 655: Surprised we need to add `"/"` here since it wasn't there before

src/main/java/org/sagebionetworks/web/client/place/ACTDataAccessSubmissionDashboardPlace.java
```diff
@@ -16,7 +16,7 @@ public String toToken() {
     return token;
   }
 
-  @Prefix("!ACTDataAccessSubmissionDashboard")
+  @Prefix("ACTDataAccessSubmissionDashboard")
```
nickgros, line 19: This dashboard uses a combination of GWT router + React Router (which I expect should have better compatibility with SWC without the hashbang), so we should use extra care to validate functionality in this place

src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/ButtonLinkWidget.java
```diff
@@ -105,7 +96,7 @@ public void configure(
   public boolean isOpenInNewWindow(String url) {
     return (
       url != null &&
-      !(url.startsWith("#!") || url.startsWith(gwt.getHostPrefix()))
+      !(!url.startsWith("http") || url.startsWith(gwt.getHostPrefix()))
```
nickgros, line 99: seems reasonable

src/main/java/org/sagebionetworks/web/server/servlet/filter/HtmlInjectionFilter.java
```diff
@@ -0,0 +1,187 @@
...
+    return targetPath.contains(":");
+    //      Class<? extends PlaceTokenizer<?>>[] placeClasses =
+    //          AppPlaceHistoryMapper.class.getAnnotation(WithTokenizers.class).value();
+    //        for (Class<? extends PlaceTokenizer<?>> c : placeClasses) {
+    //          String simpleName = c.getEnclosingClass().getSimpleName();
+    //          if (targetPath.startsWith(simpleName+":"))
+    //              return true;
+    //        }
+    //        return false;
+  }
```
nickgros: I think this is fine, I think the only place we use `:` is in query params

src/main/java/org/sagebionetworks/web/server/servlet/filter/PlacesRedirectFilter.java
```diff
@@ -1,63 +0,0 @@
...
-import javax.servlet.ServletRequest;
-import javax.servlet.ServletResponse;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
-/**
- * This filter redirects traffic heading to /Place:Token to /#!Place:Token
- *
- */
-public class PlacesRedirectFilter implements Filter {
```
nickgros, line 18: Should this do the reverse operation now? Do we not need that?

src/main/webapp/Portal.html
```diff
@@ -268,7 +259,7 @@ <h2 class="center" style="color: white; padding-top: 180px">
           Your web browser must have JavaScript enabled in order for this
           application to display correctly.<br />
           If you are an automated web crawler from a search engine, follow this
-          <a href="https://synapse.prod.sagebase.org/Portal.html#!Home:0"
+          <a href="https://synapse.prod.sagebase.org/Home:0"
             >AJAX application crawl link</a
           >
```
nickgros, line 261: This link doesn't work as far as I can tell

src/main/java/org/sagebionetworks/web/server/servlet/filter/HtmlInjectionFilter.java
```diff
@@ -0,0 +1,187 @@
...
+import org.sagebionetworks.web.server.servlet.SynapseClientImpl;
+import org.sagebionetworks.web.server.servlet.UserDataProvider;
+import org.sagebionetworks.web.shared.WikiPageKey;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+/**
+ * This filter modifies the base Portal.html file to inject with GWT place specific information, like Open Graph metadata.
+ * Will attempt to ignore bot traffic (handled by CrawlFilter)
+ */
+public class HtmlInjectionFilter extends OncePerRequestFilter {
```
nickgros, line 50: I really like the injection thru templating!

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
jay-hodgson: Path didn't matter before, since it was a hash fragment (it would maintain the same path).  Now the path is changing.

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
jay-hodgson: Good call

src/main/java/org/sagebionetworks/web/server/servlet/filter/PlacesRedirectFilter.java
jay-hodgson: We need the opposite, but the Servlet will never be sent the hash fragment.  Sooo, we need to detect this on the client-side (during app initialization).  See Portal.java onModuleLoad()

src/main/java/org/sagebionetworks/web/client/DisplayUtils.java
```diff
@@ -717,22 +700,6 @@ private static String getTeamPlaceString(Class<Team> place) {
     return getPlaceString(place.getName());
   }
 
-  private static String getTeamSearchPlaceString(Class<TeamSearch> place) {
```
jay-hodgson, line 720: unused, removed

src/main/java/org/sagebionetworks/web/client/GWTWrapperImpl.java
```diff
@@ -163,18 +163,18 @@ public boolean isWhitespace(String text) {
   @Override
   public void newItem(String historyToken, boolean issueEvent) {
     SynapseJSNIUtilsImpl._setIsInnerProgrammaticHistoryChange();
-    History.newItem(historyToken, issueEvent);
```
jay-hodgson, line 166: GWT History assumes hash fragment used for routing, use our new historian

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -190,17 +190,19 @@ public PlaceChanger getPlaceChanger() {
...
           @Override
           public void goTo(Place place) {
-            synapseJSNIUtils.setIsInnerProgrammaticHistoryChange();
-            // If we are not already on this page, go there.
-            if (!placeController.getWhere().equals(place)) {
-              try {
-                placeController.goTo(place);
-              } catch (Exception e) {
-                synapseJSNIUtils.consoleError(e.getMessage());
+            if (place != null) {
```
jay-hodgson, line 193: Not really necessary after the bug fixes, but there were cases where the place token didn't resolve to a Place, so it was passed in as null.  This resulted in an uncaught js exception, so I check for null here.

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -47,13 +47,14 @@ public static PortalGinInjector getInjector() {
...
+    // The GWT History class assumes tokens begin with '#', and we used to include the '!' as the prefix for all GWT Places.
     String initToken = History.getToken();
-    if (initToken.length() > 0 && !initToken.startsWith("!")) {
+    if (initToken.length() > 0 && initToken.startsWith("!")) {
       String fullUrl = Window.Location.getHref();
-      fullUrl = fullUrl.replace("#" + initToken, "#!" + initToken);
-      Window.Location.replace(fullUrl);
-      Window.Location.reload();
+      fullUrl = fullUrl.replace("#!", "");
+      Window.Location.assign(fullUrl);
```
jay-hodgson, line 57: This is important - this is the client-side check for #! during app initialization.  If detected, will remove the #! and reload the browser.

src/main/java/org/sagebionetworks/web/client/Html5Historian.java
```diff
@@ -0,0 +1,90 @@
+package org.sagebionetworks.web.client;
```
jay-hodgson, line 1: Based on https://gist.github.com/tbroyer/1883821?permalink_comment_id=2352599
Added replaceItem (uses replaceState())

src/main/java/org/sagebionetworks/web/client/widget/entity/MarkdownWidgetViewImpl.java
```diff
@@ -41,28 +42,39 @@ public interface Binder extends UiBinder<Widget, MarkdownWidgetViewImpl> {}
...
     if (Event.ONCLICK == event.getTypeInt()) {
       AnchorElement el = (AnchorElement) event.getCurrentTarget();
       String href = el.getHref();
-      String placeToken = href.substring(href.indexOf('!'));
-      AppPlaceHistoryMapper appPlaceHistoryMapper =
-        globalAppState.getAppPlaceHistoryMapper();
-      Place newPlace = appPlaceHistoryMapper.getPlace(placeToken);
-      globalAppState.getPlaceChanger().goTo(newPlace);
+      String placeToken = null;
+      // Detect old #!, and direct to the right place
```
jay-hodgson, line 53: Here's another general fix.  For markdown that have used relative links that include '#!', the URL that it points to is wrong (so if you hover over it and right click and copy the URL, it will be invalid).  However, if you click it, this code will run, which fixes the link on the fly.

src/main/java/org/sagebionetworks/web/client/widget/entity/MarkdownWidgetViewImpl.java
```diff
@@ -41,28 +42,39 @@ public interface Binder extends UiBinder<Widget, MarkdownWidgetViewImpl> {}
...
+      if (href.contains("#!")) {
+        placeToken = href.substring(href.indexOf('!') + 1);
+      } else {
+        placeToken = StringUtils.getGWTPlaceTokenFromURL(href);
+      }
+      if (placeToken != null) {
+        AppPlaceHistoryMapper appPlaceHistoryMapper =
+          globalAppState.getAppPlaceHistoryMapper();
+        Place newPlace = appPlaceHistoryMapper.getPlace(placeToken);
+        globalAppState.getPlaceChanger().goTo(newPlace);
```
jay-hodgson, line 63: And if you included a relative link without the '#!', it is optimized (internal SPA page change, not a page reload).

src/main/java/org/sagebionetworks/web/client/widget/entity/MarkdownWidgetViewImpl.java
```diff
@@ -88,10 +100,10 @@ private void addPlaceChangerEventHandlerToAnchors() {
     NodeList<Element> anchors = contentPanel
       .getElement()
       .getElementsByTagName("a");
-    String hostPageURL = GWT.getHostPageBaseURL();
+    String hostPrefix = gwt.getHostPrefix();
```
jay-hodgson, line 103: This is code that detects relative hrefs.  If the hostPrefix matches, then it's pointing to a local resource.

src/main/java/org/sagebionetworks/web/server/PortalServletModule.java
```diff
@@ -193,17 +187,6 @@ protected void configureServlets() {
     serveRegex("^\\/\\w+$").with(ProjectAliasServlet.class);
   }
 
-  public void handleGWTPlaces() {
-    bind(PlacesRedirectFilter.class).in(Singleton.class);
```
jay-hodgson, line 197: This was the redirector to detect when user was going to a GWT place, and it would redirect to the #! version of the page.  We still need to worry about the other direction, but not this way!

src/main/java/org/sagebionetworks/web/server/servlet/filter/BotHtml.java
```diff
@@ -0,0 +1,30 @@
+package org.sagebionetworks.web.server.servlet.filter;
```
jay-hodgson, line 1: Small helper class to organize html that should be included in pages sent back to bot traffic.  This is generally expensive to calculate, so we don't include this for normal visitors.  It typically includes a lot of extra links and basic information, rather than the skeleton page that the client fills in (usually with the same or similar information)

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -6,10 +6,9 @@
 import com.google.gwt.thirdparty.guava.common.base.Suppliers;
```
jay-hodgson, line 6: Change the CrawlFilter so that it isn't an independent filter.  It is used by the HtmlInjectionFilter, when a bot has been detected

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -470,8 +359,12 @@ private String getEntityHtml(String entityId)
       request.setNextPageToken(childList.getNextPageToken());
       i++;
     } while (i < MAX_CHILD_PAGES && childList.getNextPageToken() != null);
-    html.append("</body></html>");
-    return html.toString();
+    response.setBody(html.toString());
+
+    // SWC-6609: removed for now
+    //    response.head = getDatasetScriptElement(bundle, plainTextWiki);
```
jay-hodgson, line 362: Still not returning the dataset json ld script data.  We could turn this on, but I'd rather wait for a backend solution, and I don't want us to become very popular in Google Dataset Search before we implement managed plans

src/main/resources/org/sagebionetworks/web/client/view/SearchViewImpl.ui.xml
```diff
@@ -57,7 +57,7 @@
         <bh:Div addStyleNames="col-md-3 col-md-pull-9">
           <bh:Italic>Searching for a user profile? Try our</bh:Italic>
           <b:Anchor
-            href="#!PeopleSearch:"
+            href="PeopleSearch:"
```
hallieswan: Based on some local testing, I think that the `href` needs a forward slash for the path to be relative to the current URL -- 

```suggestion
            href="/PeopleSearch:"
```

without slash | with slash
:---:|:---:
image | image

src/main/resources/org/sagebionetworks/web/client/view/ACTDataAccessSubmissionsViewImpl.ui.xml
```diff
@@ -18,7 +18,7 @@
     <b:NavPills addStyleNames="margin-left-35 margin-right-10 margin-top-5">
       <b:AnchorListItem
         text="ACT data access dashboard"
-        href="#!ACTDataAccessSubmissionDashboard:default"
+        href="ACTDataAccessSubmissionDashboard:default"
```
hallieswan:
```suggestion
        href="/ACTDataAccessSubmissionDashboard:default"
```

src/main/resources/org/sagebionetworks/web/client/view/ACTAccessApprovalsViewImpl.ui.xml
```diff
@@ -18,7 +18,7 @@
     <b:NavPills addStyleNames="margin-left-35 margin-right-10 margin-top-5">
       <b:AnchorListItem
         text="ACT data access dashboard"
-        href="#!ACTDataAccessSubmissionDashboard:default"
+        href="ACTDataAccessSubmissionDashboard:default"
```
hallieswan:
```suggestion
        href="/ACTDataAccessSubmissionDashboard:default"
```

src/main/resources/org/sagebionetworks/web/client/view/users/PasswordResetViewImpl.ui.xml
```diff
@@ -99,7 +99,7 @@
           <b:Anchor
             text="&lt; Back to Sign In"
             addStyleNames="synapse-green"
-            href="#!LoginPlace:0"
+            href="LoginPlace:0"
```
hallieswan:
```suggestion
            href="/LoginPlace:0"
```

src/main/java/org/sagebionetworks/web/server/servlet/filter/HtmlInjectionFilter.java
```diff
@@ -240,10 +240,13 @@ protected void doFilterInternal(
                 Collections.singletonList(searchQueryRawValue)
               );
             }
-            dataModel.put(
-              PAGE_TITLE_KEY,
-              "Searching for: " + query.getQueryTerm().get(0)
-            );
+            String queryTerm = query.getQueryTerm().get(0);
+            if (queryTerm != null && queryTerm.trim().length() > 0) {
```
jay-hodgson, line 255: Special case.  If the query term is undefined, then let's not report the "Searching for" status.

src/main/resources/org/sagebionetworks/web/client/view/ChallengeOverviewViewImpl.ui.xml
```diff
@@ -19,7 +19,7 @@
     <div class="container">
       <ul class="breadcrumbs">
         <li>
-          <a href="#">Home</a>
```
jay-hodgson, line 22: Going to # no longer results in loading the root page

src/main/resources/org/sagebionetworks/web/client/widget/entity/download/CertificateWidgetViewImpl.ui.xml
```diff
@@ -9,7 +9,7 @@
       ui:field="alert"
       alertType="SUCCESS"
       primaryCTAText="View your dashboard"
-      primaryCTAHref="#!Profile:v/projects"
+      primaryCTAHref="Profile:v/projects"
```
hallieswan:
```suggestion
      primaryCTAHref="/Profile:v/projects"
```

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -830,7 +830,7 @@ public void onFailure(Throwable caught) {
                   public void onSuccess(
                     AddBatchOfFilesToDownloadListResponse result
                   ) {
-                    String href = "#!DownloadCart:0";
+                    String href = "DownloadCart:0";
```
hallieswan:
```suggestion
                    String href = "/DownloadCart:0";
```

src/main/java/org/sagebionetworks/web/client/widget/entity/WikiPageWidgetViewImpl.java
```diff
@@ -159,7 +159,7 @@ public void onClick(ClickEvent event) {
     wikiHistoryButton.setIcon(IconType.CARET_SQUARE_O_RIGHT);
     wikiCompareButton.addClickHandler(event -> {
       WikiDiff place = new WikiDiff(key);
-      DisplayUtils.newWindow("#!WikiDiff:" + place.toToken(), "_blank", "");
+      DisplayUtils.newWindow("WikiDiff:" + place.toToken(), "_blank", "");
```
hallieswan:
```suggestion
      DisplayUtils.newWindow("/WikiDiff:" + place.toToken(), "_blank", "");
```

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadHandlerWidget.java
```diff
@@ -34,8 +34,8 @@ public class FileDownloadHandlerWidget
   implements FileDownloadMenuItemView.Presenter {
 
   public static final String ACCESS_REQUIREMENTS_LINK =
-    "#!AccessRequirements:ID=";
-  public static final String LOGIN_PLACE_LINK = "#!LoginPlace:0";
```
hallieswan, line 38: Do these also need a preceding `/`?

src/main/java/org/sagebionetworks/web/server/servlet/filter/HtmlInjectionFilter.java
```diff
@@ -0,0 +1,373 @@
...
+                BOT_BODY_HTML_KEY,
+                crawlFilter.getAllProjectsHtml(query)
+              );
+            }
+          } else if (path.startsWith("/TeamSearch")) {
+            TeamSearch place = new TeamSearch(placeToken);
+            dataModel.put(
+              PAGE_TITLE_KEY,
+              SEARCHING_FOR_TEAM + place.getSearchTerm()
+            );
```
hallieswan: Will there always be team search term or should this also omit "Searching for Team: " if there is no search term?

hallieswan (approved):
Looks good to me! Just a couple small questions

src/main/java/org/sagebionetworks/web/server/servlet/filter/HtmlInjectionFilter.java
jay-hodgson: will fix

src/main/java/org/sagebionetworks/web/client/widget/entity/file/FileDownloadHandlerWidget.java
jay-hodgson: they sure do!

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -660,4 +659,19 @@ public EntityChildrenRequest createGetChildrenQuery(String parentId) {
...
     return newQuery;
   }
+
+  /**
+   * Helper to convert from the non-gwt compatible PaginatedResults to the compatible type.
+   *
+   * @param in
+   * @return
+   */
+  public <T extends JSONEntity> PaginatedResults<T> convertPaginated(
```
jay-hodgson, line 669: Brought this over from DiscussionForumClient to eliminate dependency but reduce the changes to this class

PR #5387, opened 2024-05-16
SWC-6769: configure appconfig servlet to poll and cache configurations

jinjunoh:


src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -72,6 +73,28 @@ public void onSuccess() {
...
+                ginjector
+                  .getSynapseJavascriptClient()
+                  .getFeatureFlagConfig(
+                    new AsyncCallback<String>() {
+                      @Override
+                      public void onSuccess(String config) {
+                        ginjector
+                          .getCookieProvider()
+                          .setCookie("PORTAL_FEATURE_FLAG", config);
+                      }
```
jinjunoh: Only store the configurations inside cookies if portals is under experimental mode

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -72,6 +73,28 @@ public void onSuccess() {
               // make sure jsni utils code is available to the client
               ginjector.getSynapseJSNIUtils();
 
+              if (DisplayUtils.isInTestWebsite(ginjector.getCookieProvider())) {
```
nickgros: You can go ahead and remove this condition, it's ok to store regardless of experimental mode, since it won't affect the user's experience either way.

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -72,6 +73,28 @@ public void onSuccess() {
...
+                ginjector
+                  .getSynapseJavascriptClient()
+                  .getFeatureFlagConfig(
+                    new AsyncCallback<String>() {
+                      @Override
+                      public void onSuccess(String config) {
+                        ginjector
+                          .getCookieProvider()
+                          .setCookie("PORTAL_FEATURE_FLAG", config);
```
nickgros: Can you make `"PORTAL_FEATURE_FLAG"` a constant (maybe in WebConstants.java)?

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -72,6 +73,28 @@ public void onSuccess() {
...
+                      }
+
+                      @Override
+                      public void onFailure(Throwable reason) {
+                        _consoleError(
+                          "Error getting configurations:" + reason.getMessage()
+                        );
```
nickgros: Just being picky about making the error message clear. We'll have to work through how we actually handle failure cases when we retrieve a specific flag, but what you have is good enough for now.
```suggestion
                        _consoleError(
                          "Error getting feature flag configuration:" + reason.getMessage()
                        );
```

src/main/java/org/sagebionetworks/web/shared/WebConstants.java
```diff
@@ -241,6 +241,7 @@ public class WebConstants {
   public static final String ALIAS_REDIRECTOR_SERVLET = "aliasredirector";
   public static final String ALIAS_PARAM_KEY = "alias";
   public static final String SLACK_SERVLET = "slack";
+  public static final String APPCONFIG_SERVLET = "appconfig";
```
nickgros: Naming a servlet endpoint is less important compared to adding a backend service, but I do wonder if this is the best name for this. IMO this endpoint serves a feature flag configuration, so the endpoint name should reflect that. That we use AppConfig to serve this configuration is an implementation detail, so it's a less appropriate name.

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -0,0 +1,155 @@
...
+            "-" +
+            stackConfiguration.getStackInstance() +
+            "-portal-configurations"
+          );
+      StartConfigurationSessionResult sessionResponse =
+        appConfigDataClient.startConfigurationSession(sessionRequest);
+      configurationToken = sessionResponse.getInitialConfigurationToken();
+    } catch (Exception e) {
+      configurationToken = null;
+    }
```
nickgros, line 86: Please log (or rethrow) the exception, a "swallowed" exception becomes incredibly hard to debug if something goes wrong.

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -0,0 +1,155 @@
...
+      return defaultConfigValue;
+    }
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+    throws ServletException, IOException {
+    try {
+      String configValue = configSupplier.get();
+      response.setContentType("text/plain");
```
nickgros: We are always serving JSON, right?
```suggestion
      response.setContentType("application/json");
```

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -0,0 +1,155 @@
...
+import org.sagebionetworks.aws.AwsClientFactory;
+
+public class AppConfigServlet extends HttpServlet {
+
+  public AWSAppConfigData appConfigDataClient;
+  public Supplier<String> configSupplier;
+  public String configurationToken;
+  private StackConfiguration stackConfiguration;
+  private final String defaultConfigValue =
+    "{\"default configuration\":\"true\"}";
```
nickgros: rename to reflect that it's a constant. What do you think about making the value an empty JSON object, since the actual config should always be non-empty?
```suggestion
  private final String DEFAULT_CONFIG_VALUE = "{}";
```

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -0,0 +1,155 @@
...
+    } catch (Exception e) {
+      return defaultConfigValue;
+    }
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
+    throws ServletException, IOException {
+    try {
+      String configValue = configSupplier.get();
```
nickgros: You will probably want to set the `Cache-Control` header so browsers don't cache the config, so they are always fetching the latest version (within our servers' 5min cache window)


See from `VersionsServlet.java` 
```
response.setHeader(
      WebConstants.CACHE_CONTROL_KEY,
      WebConstants.CACHE_CONTROL_VALUE_NO_CACHE
    ); 
```

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -0,0 +1,155 @@
...
+      return defaultConfigValue;
+    }
+    try {
+      GetLatestConfigurationRequest latestConfigRequest =
+        new GetLatestConfigurationRequest()
+          .withConfigurationToken(configurationToken);
+      GetLatestConfigurationResult latestConfigResponse =
+        appConfigDataClient.getLatestConfiguration(latestConfigRequest);
+      configurationToken = latestConfigResponse.getNextPollConfigurationToken();
+      ByteBuffer configData = latestConfigResponse.getConfiguration();
```
nickgros: From [the docs](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/appconfigdata/model/GetLatestConfigurationResult.html#getConfiguration--): 
> It is recommended to call ByteBuffer.duplicate() or ByteBuffer.asReadOnlyBuffer() before using or reading from the buffer. This behavior will be changed in a future major version of the SDK. 

```suggestion
      ByteBuffer configData = latestConfigResponse.getConfiguration().asReadOnlyBuffer();
```

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -0,0 +1,155 @@
...
+        java.nio.charset.StandardCharsets.UTF_8
+      ); // This may be empty if the client already has the latest version of the configuration.
+      if (!newConfigValue.isEmpty()) {
+        lastConfigValue = newConfigValue;
+      }
+      return lastConfigValue;
+    } catch (Exception e) {
+      return defaultConfigValue;
+    }
+  }
```
nickgros: I am worried about this case where the config is malformed JSON. I think the servlet should parse the JSON and handle the case where the JSON is malformed, and _never_ pass malformed JSON to the client.

WDYT of something like this, using `JSONObjectAdapter` to store data, therefore ensuring it is valid JSON?

```suggestion
  public JSONObjectAdapter getLatestConfiguration() {
    try {
      if (configurationToken == null) {
        return new JSONObjectAdapterImpl(defaultConfigValue);
      }
      GetLatestConfigurationRequest latestConfigRequest =
        new GetLatestConfigurationRequest()
          .withConfigurationToken(configurationToken);
      GetLatestConfigurationResult latestConfigResponse =
        appConfigDataClient.getLatestConfiguration(latestConfigRequest);
      configurationToken = latestConfigResponse.getNextPollConfigurationToken();
      ByteBuffer configData = latestConfigResponse
        .getConfiguration()
        .asReadOnlyBuffer();
      String newConfigString = new String(
        configData.array(),
        java.nio.charset.StandardCharsets.UTF_8
      );
      // This String may be empty if the client already has the latest version of the configuration.
      if (!newConfigString.isEmpty()) {
        lastConfigValue = new JSONObjectAdapterImpl(newConfigString);
      }
    } catch (Exception e) {
      logger.log(
        Level.SEVERE,
        "Failed to get or parse latest configuration",
        e
      );
    }
    return lastConfigValue;
  }
```

src/test/java/org/sagebionetworks/web/unitserver/servlet/AppConfigServletTest.java
```diff
@@ -0,0 +1,146 @@
...
+    // If an exception is thrown, the configuration token should remain null
+    assertEquals(null, servlet.configurationToken);
+  }
+
+  @Test
+  public void testDoGetMock() throws Exception {
+    servlet.doGet(mockRequest, mockResponse);
+
+    // Verify the response contains the expected configuration value
+    assertEquals("test configuration", responseWriter.toString());
```
nickgros: should also check the headers, like content-type

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -97,38 +93,61 @@ public void initializeConfigSupplier() {
...
-      ByteBuffer configData = latestConfigResponse.getConfiguration();
-      String newConfigValue = new String(
-        configData.array(),
+      ByteBuffer readOnlyConfigData = latestConfigResponse
+        .getConfiguration()
+        .asReadOnlyBuffer();
+      byte[] bytes = new byte[readOnlyConfigData.remaining()];
+      readOnlyConfigData.get(bytes);
+      String newConfigString = new String(
+        bytes,
```
nickgros: You can avoid instantiating and filling an array if you use a CharBuffer, though it's probably doing the exact same thing that you're doing here under-the-hood. I would just prefer to let the Java IO abstractions handle that

https://stackoverflow.com/a/23552156

https://www.mindprod.com/jgloss/bytebuffer.html

```suggestion
       CharBuffer charBuffer = StandardCharsets.UTF_8.decode(readOnlyConfigData);
       String newConfigString = charBuffer.toString();
```

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -97,38 +93,61 @@ public void initializeConfigSupplier() {
...
         java.nio.charset.StandardCharsets.UTF_8
-      ); // This may be empty if the client already has the latest version of the configuration.
-      if (!newConfigValue.isEmpty()) {
-        lastConfigValue = newConfigValue;
+      );
```
nickgros:
```suggestion
```

nickgros: goes with above suggestion

nickgros (approved):
Looks great! Just one suggestion

PR #5393, opened 2024-05-23
SWC-6770: migrate Provenance v2 visualization to feature flag

jinjunoh:


src/main/java/org/sagebionetworks/web/client/FeatureFlagConfigFactory.java
```diff
@@ -0,0 +1,29 @@
+package org.sagebionetworks.web.client;
+
+import com.google.gwt.json.client.JSONObject;
+import com.google.gwt.json.client.JSONParser;
+import com.google.gwt.json.client.JSONValue;
+import com.google.inject.Inject;
+
+public class FeatureFlagConfigFactory {
```
nickgros:
```suggestion
public class FeatureFlagConfig {
```

src/main/java/org/sagebionetworks/web/client/FeatureFlagConfigFactory.java
```diff
@@ -0,0 +1,29 @@
...
+    JSONValue parsed = JSONParser.parseStrict(json);
+    config = parsed.isObject();
+  }
+
+  public boolean isFeatureEnabled(String featureName) {
+    return (
+      DisplayUtils.isInTestWebsite(ginInjector.getCookieProvider()) ||
+      config.get(featureName).isBoolean().booleanValue()
+    );
+  }
```
nickgros: If you want to keep the Factory pattern (so you can dynamically create the class using a config), split this into two classes:

`FeatureFlagConfigFactory` has (at least) one method, `FeatureFlagConfig create(String jsonString)`.

`FeatureFlagConfig` has (at least) one method, `boolean isFeatureEnabled(String featureName)`
and will also have the constructor `FeatureFlagConfig(JSONObject config)`

nickgros: `FeatureFlagConfig(String jsonObject)`


Is it possible to get the injected cookie provider in this provides method?
```java
@Provides 
FeatureFlagConfig provideFeatureFlagConfig() {
  cookieProvider.getCookie(WebConstants.PORTAL_FEATURE_FLAG)
}

```

src/main/java/org/sagebionetworks/web/client/FeatureFlagConfig.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public boolean isFeatureEnabled(String featureName) {
+    try {
+      return (
+        DisplayUtils.isInTestWebsite(cookieProvider) ||
+        config.get(featureName).isBoolean().booleanValue()
+      );
+    } catch (Exception e) {
+      return DisplayUtils.isInTestWebsite(cookieProvider);
```
nickgros, line 35: Log error instead of swallowing exception
```suggestion
      SynapseJSNIUtilsImpl._consoleError(e)
      return DisplayUtils.isInTestWebsite(cookieProvider);
```

src/main/java/org/sagebionetworks/web/client/widget/entity/registration/WidgetRegistrarImpl.java
```diff
@@ -326,7 +329,7 @@ public WidgetRendererPresenter getWidgetRendererForWidgetDescriptorAfterLazyLoad
     } else if (contentTypeKey.equals(REFERENCE_CONTENT_TYPE)) {
       presenter = ginInjector.getReferenceRenderer();
     } else if (contentTypeKey.equals(PROVENANCE_CONTENT_TYPE)) {
-      if (DisplayUtils.isInTestWebsite(ginInjector.getCookieProvider())) {
+      if (featureFlagConfig.isFeatureEnabled("Provenance v2 visualization")) {
```
nickgros: Can you create an enum of feature flag keys that includes `"Provenance v2 visualization"` and use it in all of the places you have this string?

You could even change the signature of `isFeatureEnabled` so that the argument must be your feature flag key enum

src/test/java/org/sagebionetworks/web/client/FeatureFlagConfigTest.java
```diff
@@ -0,0 +1,49 @@
...
+  @Test
+  public void testFeatureDisabled() {
+    when(mockJsonObject.get(FEATURE_NAME))
+      .thenReturn(JSONBoolean.getInstance(false));
+    featureFlagConfig =
+      new FeatureFlagConfig(mockJsonObject, mockCookieProvider);
+
+    assertFalse(featureFlagConfig.isFeatureEnabled(FEATURE_NAME));
+  }
+}
```
nickgros, line 73: Need at least 2 more tests:
* Verify that `true` is returned if the feature is disabled but experimental mode is enabled
* Verify that the experimental mode value is returned if an exception is thrown when `mockJsonObject.get(FEATURE_NAME)` is called

nickgros (approved):

PR #5394, opened 2024-05-24
SWC-6844 - Redirect PasswordReset, ChangePassword to OneSage

nickgros:
See SWC-6844

PR #5395, opened 2024-05-28
SWC-6844 - Remove Account Settings functionality and redirect users to OneSage

nickgros:
Subsumes #5394 to completely remove Account Settings functionality from SWC and redirect users to OneSage

Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/956


src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
```diff
@@ -80,15 +79,6 @@ public static class SynapseComponents {
...
     > CreatedByModifiedBy;
-    public static ReactComponentType<
-      TwoFactorAuthSettingsPanelProps
-    > TwoFactorAuthSettingsPanel;
-    public static ReactComponentType<
-      TwoFactorBackupCodesProps
-    > TwoFactorBackupCodes;
-    public static ReactComponentType<
-      TwoFactorEnrollmentFormProps
-    > TwoFactorEnrollmentForm;
```
nickgros, line 91: Later we can remove these to trim down the UMD bundle

src/main/java/org/sagebionetworks/web/client/presenter/AccountPresenter.java
```diff
@@ -79,9 +79,7 @@ public void onSuccess(Void result) {
           view.showInfo(DisplayConstants.EMAIL_SUCCESS);
           globalAppState
             .getPlaceChanger()
-            .goTo(
-              new Profile(Profile.VIEW_PROFILE_TOKEN, ProfileArea.SETTINGS)
-            );
+            .goTo(new Profile(Profile.VIEW_PROFILE_TOKEN, ProfileArea.PROFILE));
```
nickgros, line 82: This should only happen if your EmailValidationSignedToken link points to www.synapse.org, which (I assume) would only happen in a previous version of the stack. I am not sure how long those tokens last, so this may never work for users.

The profile page shows your email addresses if you log in, so this should be good enough.

src/main/java/org/sagebionetworks/web/client/presenter/EmailInvitationPresenter.java
```diff
@@ -148,18 +146,19 @@ public void onFailure(Throwable t) {
...
-              // SWC-4741: invitation not associated to the current user, send user to the Settings page to add
-              // the new email address
+              // SWC-4741: invitation not associated to the current user, send user to the Profile page, where they can
+              // click 'Edit Profile', then 'Change email' to get to OneSage and add the new email address
+              // NOTE: We don't send the user directly to OneSage so that they have a chance to read the message first
               placeChanger.goTo(
                 new Profile(
                   authController.getCurrentUserPrincipalId(),
-                  ProfileArea.SETTINGS
+                  ProfileArea.PROFILE
```
nickgros: Keep showing an error message for the edge case where you try to join a team using an emailed link, but the email address is not tied to your account. 

From the profile page, the user can follow a few steps to change their email address, so this should be good enough for this uncommon scenario.

src/main/java/org/sagebionetworks/web/client/presenter/PersonalAccessTokensPresenter.java
```diff
@@ -3,38 +3,35 @@
...
-  }
+  public void setPlace(PersonalAccessTokenPlace place) {}
 
   @Override
   public void start(AcceptsOneWidget panel, EventBus eventBus) {
-    this.view.render();
-    panel.setWidget(view.asWidget());
+    Window.Location.replace(
+      WebConstants.ONESAGE_PRODUCTION_URL +
+      "/authenticated/personalaccesstokens"
```
nickgros: For PATs, redirect to OneSage

src/main/java/org/sagebionetworks/web/client/presenter/ProfilePresenter.java
```diff
@@ -978,7 +953,7 @@ private void refreshData(ProfileArea tab) {
         refreshTeams();
         break;
       case SETTINGS:
-        getSettingsPresenter().configure();
+        Window.Location.replace(ONESAGE_ACCOUNT_SETTINGS_URL);
```
nickgros, line 944: If in the settings area, redirect to OneSage

src/main/java/org/sagebionetworks/web/client/presenter/TwoFactorAuthPresenter.java
```diff
@@ -2,79 +2,24 @@
...
   public void start(AcceptsOneWidget panel, EventBus eventBus) {
-    panel.setWidget(this.view.asWidget());
-  }
-
-  @Override
-  public void setPlace(final TwoFactorAuthPlace place) {
-    view.setPresenter(this);
-    view.clear();
-    configureView(place);
+    Window.Location.replace(WebConstants.ONESAGE_ACCOUNT_SETTINGS_URL);
```
nickgros, line 20: 2FA place will redirect to OneSage

src/main/java/org/sagebionetworks/web/client/presenter/users/PasswordResetPresenter.java
```diff
@@ -2,180 +2,27 @@
...
-          public void onFailure(Throwable caught) {
-            view.showExpiredRequest();
-          }
-        }
-      );
-    } else {
-      view.showRequestForm();
-    }
+  public void start(AcceptsOneWidget acceptsOneWidget, EventBus eventBus) {
+    Window.Location.replace(ONE_SAGE_RESET_PASSWORD_URL);
```
nickgros, line 26: PasswordReset place redirects to OneSage

src/main/java/org/sagebionetworks/web/client/presenter/users/RegisterAccountPresenter.java
```diff
@@ -1,112 +1,36 @@
...
+    if (token != null && token.contains("@")) {
+      // is likely an email address
+      Window.Location.replace(
+        WebConstants.ONESAGE_PRODUCTION_URL + "/register1?email=" + token
+      );
+    } else {
+      Window.Location.replace(
+        WebConstants.ONESAGE_PRODUCTION_URL + "/register1"
+      );
     }
```
nickgros: Redirect. If a token representing an email address was included, forward the token to OneSage

jay-hodgson: This is always URL decoded, right?  (We don't need to look for "%40" as well?)

nickgros: The existing SWC feature assumes it is decoded. I tested with an encoded string and it did not automatically decode it.

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/shared/WebConstants.java
```diff
@@ -25,6 +25,11 @@ public class WebConstants {
 
   public static final String INVALID_EMAIL_MESSAGE = "Invalid email address";
 
+  public static final String ONESAGE_PRODUCTION_URL =
+    "https://accounts.sagebionetworks.synapse.org";
+  public static final String ONESAGE_ACCOUNT_SETTINGS_URL =
+    ONESAGE_PRODUCTION_URL + "/authenticated/myaccount";
```
jay-hodgson: This should set the appId to "synapse.org" (in case it has been customized for some other portal experience)

jay-hodgson (approved):

PR #5398, opened 2024-05-30
SWC-6829

jay-hodgson:


jay-hodgson:
Provide an easy way for users to immediately access the file that was most recently uploaded

hallieswan (approved):
Looks good to me!

PR #5399, opened 2024-05-30
SWC-6849

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/googlemap/GoogleMapViewImpl.java
```diff
@@ -171,7 +171,7 @@ private static native void _addMarker(
   ) /*-{
 
 		var image = {
-			url : 'images/synapse-map-marker.png',
+			url : '/images/synapse-map-marker.png',
```
jay-hodgson, line 174: Verified that the map works with or without this change

src/main/java/org/sagebionetworks/web/client/AppLoadingView.java
```diff
@@ -21,7 +21,7 @@ public AppLoadingView() {
      * DO NOT ADD THIS TO THE MAIN IMAGE BUNGLE! This is shown while the image bundle is being download
      * to the browser so putting it in the bundle defeats the whole purpose of this image.
      */
-    final Image ajaxImage = new Image("images/main-page-load.gif");
+    final Image ajaxImage = new Image("/images/main-page-load.gif");
```
jay-hodgson, line 24: Not sure when this is used, but this was the only other reference that I could find in the codebase with the broken pattern

xschildw (approved):

PR #5402, opened 2024-05-31
SWC-6854: store/read query in the hash fragment

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -57,6 +57,13 @@ public void onModuleLoad() {
     if (initToken.length() > 0 && initToken.startsWith("!")) {
       String fullUrl = Window.Location.getHref();
       fullUrl = fullUrl.replace("#!", "");
+      // SWC-6854: Also detect a tables query in the hash fragment.  If exist, move just the query to the hash fragment
+      if (
+        initToken.startsWith("!Synapse") && initToken.contains("/tables/query/")
+      ) {
+        fullUrl = fullUrl.replace("/tables/query/", "/tables/#query/");
```
jay-hodgson, line 64: For old links, actively move the query portion to the hash fragment during client-side app init.

src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
```diff
@@ -145,4 +145,8 @@ void showJiraIssueCollector(
   long getLastModified(JavaScriptObject blob);
 
   void setIsInnerProgrammaticHistoryChange();
+
+  String setHash(String hash);
+
+  String getHash();
```
jay-hodgson, line 151: Provide way to get and set the hash fragment

PR #5421, opened 2024-06-11
SWC-6869

jay-hodgson:


jay-hodgson:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/999

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -570,19 +570,32 @@ public void onPreviewNativeEvent(NativePreviewEvent event) {
...
+                  String href = targetElement.getAttribute("href");
+                  boolean handled = handleRelativePathClick(href);
+                  if (handled) {
+                    event.cancel();
+                  }
+                }
+              }
+            } catch (Exception e) {
+              synapseJSNIUtils.consoleError(e.getMessage());
+            }
```
nickgros, line 598: Does this work with the tabs in the `DataAccessManagement` place (ACT/AR reviewer dashboard), where the subpath is controlled by React Router?

nickgros (approved):
Had one question, but I see no reason to not try this without lots of automated e2e tests that we don't have today.

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
jay-hodgson: Yes, verified that this works as expected, thanks for asking.
There is an uncaught exception whenever I switch tabs in this place, but that's true on production today as well (unexpected null).
Screen Shot 2024-06-12 at 2 33 34 PM

PR #5429, opened 2024-06-13
SWC-6862: Set page description based on entity type, and include Syna…

jay-hodgson:
…pse platform description.


nickgros (approved):
Awesome!

PR #5431, opened 2024-06-18
SWC-6761 - bump Bootstrap 3 to v3.4.1 and load through Portal.html

nickgros:
Bootstrap 3.3.7 is loaded by GWTBootstrap3, but this will not occur if we load our own version of Bootstrap first.

The loaded version of Bootstrap can be determined by running in the console: `$.fn.tooltip.Constructor.VERSION`

image


src/main/webapp/Portal.html
```diff
@@ -167,6 +167,10 @@
       loadCss(cdnEndpoint + 'generated/swc.css')
       loadJs(cdnEndpoint + 'generated/jquery.min.js')
       loadJs(cdnEndpoint + 'generated/jquery-ui.min.js')
+      // Load latest version of Bootstrap 3 to override the version bundled by GWTBootstrap3
+      // The Bootstrap 3 custom theme CSS is loaded through swc.scss
```
nickgros, line 171: The old CSS still looks ok after clicking around a bit after updating the JS package

src/main/webapp/Portal.html
```diff
@@ -167,6 +167,10 @@
...
+      // Note: we cannot source this from NPM because we rely on Bootstrap 4 to be in node_modules/bootstrap to compile our scss.
+      loadJs(cdnEndpoint + 'js/bootstrap-3.4.1-dist/js/bootstrap.min.js')
```
jay-hodgson: Is it possible to pull bootstrap 3.4.1 from npm in package.json and copy over the assets into the generated folder?

jay-hodgson (approved):

src/main/webapp/Portal.html
nickgros: I originally did not because because when we compile our CSS with Sass, it also uses scss source files from synapse-react-client (so scss variables can be overridden). synapse-react-client's scss relies on Bootstrap 4's scss to be present, so ultimately Bootstrap 4's scss must be located in `SynapseWebClient/node_modules/bootstrap`. 

We can work around this by using an [alias](https://classic.yarnpkg.com/lang/en/docs/cli/add/#toc-yarn-add-alias)

src/main/webapp/Portal.html
```diff
@@ -167,6 +167,10 @@
       loadCss(cdnEndpoint + 'generated/swc.css')
       loadJs(cdnEndpoint + 'generated/jquery.min.js')
       loadJs(cdnEndpoint + 'generated/jquery-ui.min.js')
+      // Load latest version of Bootstrap 3 to override the version bundled by GWTBootstrap3
+      // The Bootstrap 3 custom theme CSS is loaded through swc.scss
+      // Note: we alias to bootstrap3 because we rely on Bootstrap 4 to be in node_modules/bootstrap to compile our scss (via synapse-react-client).
+      loadJs(cdnEndpoint + 'node_modules/bootstrap3/dist/js/bootstrap.min.js')
```
jay-hodgson: Is node_modules available in the webapp output directory?
If it was, I would expect this asset (for example) to load properly: https://cdn-www.synapse.org/node_modules/plotly.js/dist/plotly.min.js

nickgros: You're right, I skipped the copy step

PR #5432, opened 2024-06-18
SWC-6882 - ReactComponent widget that can accept child widgets

nickgros:
- Create ReactComponent widget to replace ReactComponentDiv, ReactComponentSpan. The constructor can be used to pick a root element tag.
- ReactComponent widget extended to support child widgets. If all child widgets are ReactComponent widgets, the ReactNodes of each widget are passed as React children. Otherwise, the widget elements are injected into the component's ref element
- Added abstract class `HasStyle` for ReactComponent widgets that have a `style` prop. The `style` prop will be used to manipulate the component's visibility when the `setVisible` widget method is called.

This is part 1. Part 2 will use this new code to replace GWTBootstrap3 `Row`/`Column` with [MUI's Grid v2](https://mui.com/material-ui/react-grid2/#fluid-grids)


src/main/java/org/sagebionetworks/web/client/jsinterop/react/HasStyle.java
```diff
@@ -0,0 +1,54 @@
...
+import org.sagebionetworks.web.client.jsinterop.JsObject;
+import org.sagebionetworks.web.client.jsinterop.PropsWithStyle;
+import org.sagebionetworks.web.client.widget.ReactComponent;
+
+/**
+ * Abstract class for React component widgets that have a style prop. The style prop for the component may be manipulated
+ * to show/hide the component based on the current state of the widget.
+ * @param <T> the prop type.
+ */
+public abstract class HasStyle<T extends PropsWithStyle>
```
nickgros, line 13: With ReactComponent widgets, we need special logic to manipulate the visibility. When we inject ReactComponent widgets as children of other ReactComponents, the child components do not have the root element where GWT would typically add `display: none`, so the superclass' `setVisible` implementation does not work by itself.

We can make this work at least for components that have a `style` prop by appending `display: none` to `style`, and updating the props of the component.

devdocs/ReactIntegration.md
```diff
@@ -17,7 +17,7 @@ Using JsInterop to call your React component can be summarized in these basic st
 
 1. Create a class to represent your components' props
 1. Add your component to the SRC JsInterop type
-1. Add a ReactComponentDiv to your View
+1. Add a ReactComponent to your View
```
nickgros, line 20: Renamed `ReactComponentDiv` to `ReactComponent`. `ReactComponentSpan` was removed; the root element of the `ReactComponent` can be chosen when calling the widget constructor.

src/main/java/org/sagebionetworks/web/client/jsinterop/ReactComponentProps.java
```diff
@@ -1,7 +1,46 @@
...
+import com.google.gwt.dom.client.Element;
+import jsinterop.annotations.JsConstructor;
+import jsinterop.annotations.JsFunction;
+import jsinterop.annotations.JsOverlay;
 import jsinterop.annotations.JsPackage;
 import jsinterop.annotations.JsType;
 
 @JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
-public class ReactComponentProps {}
+public class ReactComponentProps {
```
nickgros, line 11: Extend `ReactComponentProps` to add general support for passing a `ref` or `children`. If a component doesn't accept these props, it will just ignore them.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,225 @@
...
+import org.sagebionetworks.web.client.jsinterop.React;
+import org.sagebionetworks.web.client.jsinterop.ReactComponentProps;
+import org.sagebionetworks.web.client.jsinterop.ReactComponentType;
+import org.sagebionetworks.web.client.jsinterop.ReactDOMRoot;
+import org.sagebionetworks.web.client.jsinterop.ReactNode;
+
+/**
+ * Automatically unmounts the ReactComponent (if any) inside this div when this container is detached/unloaded.
+ */
+public class ReactComponent<T extends ReactComponentProps>
```
nickgros, line 35: Contains all of the logic from `ReactComponentDiv`/`ReactComponentSpan` plus lots of new logic to handle adding widget children.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,225 @@
...
+
+  private ReactNode<?> reactElement;
+
+  public ReactComponent() {
+    this(DivElement.TAG);
+  }
+
+  public ReactComponent(String tag) {
+    setElement(Document.get().createElement(tag));
+  }
```
nickgros, line 51: Use `div` by default, but a caller could use `span` (or any other tag)

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,225 @@
...
+        if (node != null) {
+          // Once the DOM node is defined, inject each child
+          getChildren().forEach(w -> node.appendChild(w.getElement()));
+        }
+      };
+
+      if (this.props == null) {
+        this.props = (T) JsPropertyMap.of();
+      }
+      this.props.ref = refCallback;
```
nickgros: This block handles adding children (special case if all are React children, otherwise append generic widgets)

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    // This component may be a React child of another component. If so, re-render the entire tree
+    // This can resolve issues where e.g. a component was set change visibility
+    ReactComponent<?> componentToRender = getRootReactComponentWidget();
+    if (componentToRender == this) {
+      root.render(this.reactElement);
+    } else {
+      componentToRender.rerender();
+    }
```
nickgros: Many bugs went away when I started re-rendering from the root node in cases where we had nested React components.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentLifecycleUtils.java
nickgros, line 1: This class just held shared code between ReactComponentDiv and ReactComponentSpan. Now ReactComponent contains all of the logic.

jay-hodgson:
I think it should be a rare and intentional action to do a force push to the repository since there is a risk of losing commits in this process.  The only reason I can think of would be to clean up the commit history, which I don't think we are generally concerned about (for increased transparency of the dev process).
Going forward, could you restrain from force pushing, or explain why it's needed?

src/main/java/org/sagebionetworks/web/client/jsinterop/react/HasStyle.java
```diff
@@ -0,0 +1,54 @@
...
+  extends ReactComponent<T> {
+
+  public HasStyle() {
+    super();
+  }
+
+  public void setStyle(JsPropertyMap<String> style) {
+    this.props.style =
+      (JsPropertyMap<String>) JsObject.assign(new JsObject(), (JsObject) style);
```
jay-hodgson, line 23: Is this essentially cloning the style object?  In any case, could you add a comment (because I was confused)?

jay-hodgson: And maybe make a helper setter method since this code is repeated a few times in this file?

src/main/java/org/sagebionetworks/web/client/jsinterop/react/HasStyle.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Override
+  public void setVisible(boolean visible) {
+    if (visible) {
+      if (this.props != null && this.props.style != null) {
+        this.props.style.delete("display");
+        this.props.style =
```
jay-hodgson: Use setStyle()

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,243 @@
...
+
+  public ReactComponent() {
+    this(DivElement.TAG);
+  }
+
+  public ReactComponent(String tag) {
+    setElement(Document.get().createElement(tag));
+  }
+
+  boolean allChildrenAreReactComponents() {
```
jay-hodgson: Many of these methods are package-private, is that intentional?  Can they be private?

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,243 @@
...
+        break;
+      }
+    }
+    return allChildrenAreReactComponents;
+  }
+
+  boolean isRenderedAsReactComponentChild() {
+    return (
+      getParent() instanceof ReactComponent &&
+      ((ReactComponent<?>) getParent()).allChildrenAreReactComponents()
```
jay-hodgson, line 67: So if a sibling is not a react component, then isRenderedAsReactComponentChild is false?

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,243 @@
...
+      List<ReactComponent<?>> childWidgets = new ArrayList<>();
+      getChildren().forEach(w -> childWidgets.add(((ReactComponent<?>) w)));
+
+      ReactNode<?>[] childReactElements = childWidgets
+        .stream()
+        .map(ReactComponent::getReactElement)
+        .toArray(ReactNode<?>[]::new);
+
+      this.reactElement =
+        React.cloneElement(reactElement, this.props, childReactElements);
```
jay-hodgson, line 119: nifty

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -0,0 +1,243 @@
...
+    } else if (getChildren().size() > 0) {
+      // Create a callback ref that will allow us to inject the GWT children into the DOM
+      ReactComponentProps.CallbackRef refCallback = (Element node) -> {
+        if (node != null) {
+          // Once the DOM node is defined, inject each child
+          getChildren().forEach(w -> node.appendChild(w.getElement()));
```
jay-hodgson, line 125: also nifty :D

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
nickgros: Correct! The behavior if we mixed React component children and injected widgets would be unusual from a dev perspective (and potentially undefined depending on the component's behavior). I think in most cases, your GWT widgets would end up injected in the DOM after the children. So if you had widget XML like

```xml
<WrapperThatIsAReactComponentWidget>
  <NonReactWidget1 />
  <ReactWidget />
  <NonReactWidget2 />
</WrapperThatIsAReactComponent>
```
 
You would end up with HTML in a different order, like:

```xml
<wrapper>
  <!-- React children of wrapper are rendered here -->
  <reactWidget />
  <!-- non-React widgets are injected afterwards -->
  <nonReactWidget1 />
  <nonReactWidget2 />
</wrapper>
```

That being said, this approach has the disadvantage that child ReactComponent widgets that are injected will retain their wrapper (e.g. `div` or `span`) that they would not otherwise have, which itself can cause layout issues.

I opted for extra wrappers over incorrect ordering since I think it would generally be easier to work around extra wrappers using CSS (if necessary, it's often not). And if we didn't want to use CSS to fix issues with the extra wrapper, the alternative is migrating the non-React sibling widgets to become React components.

nickgros:
> Going forward, could you restrain from force pushing, or explain why it's needed?

I only ever force-push to private branches on my own fork, so it should never cause problems with shared history. I prefer rebasing over merging as I find the linear commit history to be easier to review, bisect, revert.

I personally often make many commits that are incomplete/non-atomic that I often want to later split into multiple PRs to simplify the review process (and in the case of risky changes, easily revert). IMO, re-writing private history is the more natural way to do that instead of linearly removing and cherry-picking code.

I've been doing this for a while, but [this recent blog post](https://rednafi.com/misc/on_rebasing/) does a good job of capturing my opinions and work style.

src/main/java/org/sagebionetworks/web/client/jsinterop/react/HasStyle.java
nickgros: Yes, will do

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
nickgros: I think at one point I had other classes call these methods, but cleaned it up to consolidate the logic in ReactComponent.java. I will make these private.

src/main/java/org/sagebionetworks/web/client/jsinterop/react/HasStyle.java
```diff
@@ -27,25 +27,27 @@ public void setVisible(boolean visible) {
     if (visible) {
       if (this.props != null && this.props.style != null) {
         this.props.style.delete("display");
-        this.props.style =
+        // Clone the style object so React treats it as a new prop and properly re-renders.
+        setStyle(
```
jay-hodgson: Could this not be simplified to just
setStyle(this.props.style)
(since setStyle does the clone)?  I'm probably missing something

nickgros: No you're right, I guess I'm just a bit slow today 🤦

PR #5438, opened 2024-06-24
SWC-6909 (release-503)

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/renderer/WikiSubpageOrderEditorTreeViewImpl.java
```diff
@@ -163,16 +163,6 @@ public void selectTreeItem(SubpageOrderEditorTreeNode toSelect) {
   private Widget makeListItem(final SubpageOrderEditorTreeNode node) {
     final Anchor l = new Anchor(node.getText());
     l.addStyleName("link");
-    l.setHref(
-      "/Synapse:" +
```
jay-hodgson, line 167: Our smart relative link click handler will handle this link if we assign the GWT path!  Let the custom click handler do the work (in the addClickHandler call below)

xschildw (approved):

PR #5444, opened 2024-07-03
SWC-5266 - Show entity type icon in deletion prompt

nickgros:
Also show additional warning from project deletion

image

image

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/PortalGinInjector.java
+++ b/src/main/java/org/sagebionetworks/web/client/PortalGinInjector.java
@@ -53,6 +53,7 @@ import org.sagebionetworks.web.client.security.AuthenticationController;
 import org.sagebionetworks.web.client.view.DivView;
 import org.sagebionetworks.web.client.widget.CommaSeparatedValuesParser;
 import org.sagebionetworks.web.client.widget.CopyTextModal;
+import org.sagebionetworks.web.client.widget.EntityTypeIcon;
 import org.sagebionetworks.web.client.widget.FileHandleWidget;
 import org.sagebionetworks.web.client.widget.LoadMoreWidgetContainer;
 import org.sagebionetworks.web.client.widget.QuarantinedEmailModal;
@@ -862,4 +863,6 @@ public interface PortalGinInjector extends Ginjector {
   FollowingPagePresenter getFollowingPagePresenter();
 
   ColumnModelsEditorWidget getColumnModelsEditorWidget();
+
+  EntityTypeIcon getEntityTypeIcon();
 }
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
@@ -17,4 +17,8 @@ public class ReactDOM {
   public static native ReactDOMRoot createRoot(Element container);
 
   public static native boolean unmountComponentAtNode(Element container);
+
+  public static native void flushSync();
+
+  public static native void flushSync(Callback callback);
 }
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
@@ -18,7 +18,17 @@ public class ReactDOM {
 
   public static native boolean unmountComponentAtNode(Element container);
 
+  /**
+   * flushSync lets you force React to flush any updates inside the provided callback synchronously. This ensures that the DOM is updated immediately.
+   * Using flushSync is uncommon and can hurt the performance of your app. Most of the time, flushSync can be avoided. Use flushSync as last resort.
+   * <a href="https://react.dev/reference/react-dom/flushSync">See docs</a>
+   */
   public static native void flushSync();
 
+  /**
+   * flushSync lets you force React to flush any updates inside the provided callback synchronously. This ensures that the DOM is updated immediately.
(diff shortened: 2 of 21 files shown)


src/main/java/org/sagebionetworks/web/client/PortalGinInjector.java
```diff
@@ -862,4 +863,6 @@ public interface PortalGinInjector extends Ginjector {
   FollowingPagePresenter getFollowingPagePresenter();
 
   ColumnModelsEditorWidget getColumnModelsEditorWidget();
+
+  EntityTypeIcon getEntityTypeIcon();
```
nickgros, line 867: `EntityTypeIcon` can be retrieved through PortalGinInjector (and thus may be easily mocked in test)

src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
```diff
@@ -17,4 +17,8 @@ public interface Callback {
   public static native ReactDOMRoot createRoot(Element container);
 
   public static native boolean unmountComponentAtNode(Element container);
+
+  public static native void flushSync();
+
+  public static native void flushSync(Callback callback);
```
nickgros: [Register this ReactDOM function](https://react.dev/reference/react-dom/flushSync)

src/main/java/org/sagebionetworks/web/client/widget/EntityTypeIcon.java
```diff
@@ -1,52 +1,11 @@
...
-   * @param type
-   */
-  public void setType(String type) {
-    EntityType enumValue = EntityType.file;
-    try {
-      enumValue = EntityType.valueOf(type);
-    } catch (IllegalArgumentException e) {}
-    setType(enumValue);
-  }
+public interface EntityTypeIcon extends HasVisibility, IsWidget {
```
nickgros, line 8: Extract interface to make it easier to mock

src/main/java/org/sagebionetworks/web/client/widget/EntityTypeIconImpl.java
```diff
@@ -0,0 +1,56 @@
+package org.sagebionetworks.web.client.widget;
+
+import com.google.gwt.dom.client.SpanElement;
+import org.sagebionetworks.repo.model.EntityType;
+import org.sagebionetworks.web.client.jsinterop.EntityTypeIconProps;
+import org.sagebionetworks.web.client.jsinterop.React;
+import org.sagebionetworks.web.client.jsinterop.ReactNode;
+import org.sagebionetworks.web.client.jsinterop.SRC;
+
+public class EntityTypeIconImpl
```
nickgros, line 10: Implementation, one tiny change (called out)

src/main/java/org/sagebionetworks/web/client/widget/EntityTypeIconImpl.java
```diff
@@ -0,0 +1,56 @@
...
+  extends ReactComponent
+  implements EntityTypeIcon {
+
+  public EntityTypeIconImpl() {
+    super(SpanElement.TAG);
+  }
+
+  public EntityTypeIconImpl(EntityType type) {
+    super(SpanElement.TAG);
```
nickgros, line 19: We were missing this `super` call to make this a `span` by default

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2776,18 +2781,32 @@ public void onDeleteWiki() {
...
   }
 
+  private String getIconHTML() {
+    EntityTypeIcon icon = ginInjector.getEntityTypeIcon();
+    icon.configure(entityBundle.getEntityType());
+    // Pull the HTML out of the icon React component and inline it
+    // It's possible ReactDOM has not injected HTML to the widget yet, so use flushSync to force it to be written
+    ReactDOM.flushSync();
+    return icon.getElement().getInnerHTML();
+  }
```
nickgros, line 2791: Super hacky, but easier & safer than reimplementing the dialog and wiring it into EntityActionController. We can already pass HTML to `showConfirmDeleteDialog`. So create a React component with the icon, pull out the HTML, and stick the HTML into the message. 

I chose this route because I'd prefer to wait for substantial design changes to this component (or substantially more logic from EntityActionControllerImpl moved to synapse-react-client) before reimplementing it.

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2776,18 +2781,32 @@ public void onDeleteWiki() {
...
-      "\"?";
-    if (this.entity instanceof Folder) {
-      display += DELETE_FOLDER_EXPLANATION;
+      "</strong>?";
+    if (this.entity instanceof Folder || this.entity instanceof Project) {
+      display +=
+      DELETE_CONTAINER_EXPLANATION_START +
+      this.entityTypeDisplay +
+      DELETE_CONTAINER_EXPLANATION_END;
     }
```
nickgros, line 2810: Also display container warning for projects

src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
jay-hodgson: Could you add a comment indicating that people should probably not need to use this?
something like: "flushSync is used to force React to flush any pending work and update the DOM synchronously, and  can significantly hurt performance. Use sparingly.

jay-hodgson (approved):

PR #5451, opened 2024-07-12
SWC-6930: replace synapse footer and homepage

jay-hodgson:
Must update SRC to the latest before merging


src/main/java/org/sagebionetworks/web/client/widget/footer/FooterViewImpl.java
```diff
@@ -1,163 +1,50 @@
...
-      sageLogo1.addClickHandler(sageLogoClickHandler);
-      sageLogo2.addClickHandler(sageLogoClickHandler);
+      wrapper.add(container);
       refresh();
     };
-    gwt.scheduleExecution(constructViewCallback, 2500);
-    this.cookies = cookies;
-    this.globalAppState = globalAppState;
-    this.hideACTActionsButton = hideACTActionsButton;
+    gwt.scheduleExecution(constructViewCallback, 2000);
```
nickgros, line 42: just curious, why do we delay initializing the footer?

nickgros (approved):
looks great!

src/main/java/org/sagebionetworks/web/client/widget/footer/FooterViewImpl.java
jay-hodgson: I think it was introduced to delay working on something that should be off screen, and because some page content doesn't push the footer down until loaded (so the footer would initially appear towards the top of the page, and eventually be pushed to the bottom)

PR #5459, opened 2024-07-18
SWC-6946 - Add filter to rewrite JS Content-Type; add script to use tomcat in dev

nickgros:


jay-hodgson (approved):
dev:tomcat and dev:codeserver scripts are not working for me, but the new filter lgtm

PR #5461, opened 2024-07-19
Move dev script to a bash script, allow compiling for Firefox in

nickgros:
PortalDebug module, move sass compilation to process-resources to ensure it is reliably copied in the compile step

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -370,7 +370,7 @@
                 src/main/webapp/sass:src/main/webapp/generated
               </arguments>
             </configuration>
-            <phase>compile</phase>
+            <phase>process-resources</phase>
           </execution>
         </executions>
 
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -370,7 +370,7 @@
                 src/main/webapp/sass:src/main/webapp/generated
               </arguments>
             </configuration>
-            <phase>compile</phase>
+            <phase>process-resources</phase>
```
nickgros, line 373: Move `sass` compilation to `process-resources` to ensure the static resources created by this job are copied, which happens in the `compile` step

src/main/resources/org/sagebionetworks/web/PortalDebug.gwt.xml
```diff
@@ -2,5 +2,5 @@
 <module rename-to='Portal'>
   <inherits name="org.sagebionetworks.web.Portal" />
 
-  <set-property name="user.agent" value="safari" />
+  <set-property name="user.agent" value="safari,gecko1_8" />
```
nickgros, line 5: allow using firefox with PortalDebug module. If you have precompile off, it won't compile the gecko1_8 permutation unless you actually load the app in devmode with the expected user agent

PR #5463, opened 2024-07-22
SWC-6895 - Bump synapse-react-client to v3.3.3

nickgros:
Fixes SWC-6895. Note that this will also include SWC-6929


nickgros:
Note - patching production

xschildw (approved):

PR #5466, opened 2024-07-25
SWC-6947

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/jsinterop/AccessRequirementListProps.java
```diff
@@ -18,18 +19,21 @@ public interface Callback {
...
     Callback onHide,
     List<AccessRequirement> accessRequirements,
-    String entityId
+    String subjectId,
+    RestrictableObjectType subjectType
   ) {
     AccessRequirementListProps props = new AccessRequirementListProps();
     props.renderAsModal = true;
     props.onHide = onHide;
-    props.entityId = entityId;
```
jay-hodgson, line 32: Use subject Id and RestrictableObjectType enumeration value in AccessRequirementList instead of deprecated entityId prop value.

src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/ManagedACTAccessRequirementWidgetViewImpl.java
```diff
@@ -419,8 +419,4 @@ public void setSubjectsDefinedByAnnotations(
     subjectsDefinedByAnnotationsUI.setVisible(v);
     subjectsDefinedInAccessRequirementUI.setVisible(!v);
   }
-
-  public void hideRequestAccessModal() {
-    requestDataAccessWidget.clear();
```
jay-hodgson, line 424: Clearing widgets did not unmount the react component, which is what we need to do in this case (since it will display the modal if it is mounted and configured to show itself as a modal).  So instead wrap the ReactComponent in a GWT SimplePanel (div that allows a single child), and add the ReactComponent when it should be shown (and remove when it should be hidden)

nickgros: That works

nickgros (approved):

PR #5467, opened 2024-07-25
SWC-6939

jay-hodgson:


jay-hodgson:
The SRC HelpPopover buttons don't seem to be utilizing the MUI theme, but it is functional and I think looks ok:
Screen Shot 2024-07-25 at 4 41 53 PM

src/main/java/org/sagebionetworks/web/client/widget/HelpWidget.java
```diff
@@ -45,190 +30,73 @@
...
-        DisplayUtils.getESCKeyDownHandler(event -> {
-          hidePopover();
-        }),
-        KeyDownEvent.getType()
-      );
-  }
-
-  public void setWidth(String width) {
-    widget.setWidth(width);
+    ReactNode component = React.createElement(
```
nickgros: Does this fix the theme issue?

```suggestion
    ReactNode component = React.createElementWithThemeContext(
```

nickgros: Also, I think it would be a good idea to leave a comment that we should update this to use `createElementWithSynapseContext`. I am guessing that this cannot be easily done without tons of changes because it's a pain to do dependency injection with Gin in widgets that use UiBinder

nickgros: We should investigate moving the data that we pass to 'SynapseContext' to a global data store that exists outside of the component tree that would update the component trees when the store state changes (e.g. redux, Jotai's [createStore](https://jotai.org/docs/core/store#createstore)).

If we put the store in a global object, then we could just pass the global store to the component tree when we mount it instead of relying on Gin to provide our context data. In the context of creating these components, we aren't taking advantage of the benefits of dependency injection, so I can't think of a reason we shouldn't do this.

nickgros (approved):

jay-hodgson:
Theme issue has been fixed in this PR (thanks to Nick)

PR #5470, opened 2024-07-26
SWC-6943 - Use sessionStorage to store feature flag information

nickgros:


jay-hodgson (approved):
Successful transition to using session storage instead of a cookie for the portal feature flag json object.

jay-hodgson:
Plans to remove the legacy "experimental mode" single value, and only rely on the feature flag info?

nickgros:
@jay-hodgson we need to add a mechanism to override the config, which is tracked by https://sagebionetworks.jira.com/browse/SWC-6847

PR #5471, opened 2024-07-26
SWC-7000: fix logical error

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/TeamPresenter.java
```diff
@@ -61,7 +61,6 @@ public class TeamPresenter
   private IsACTMemberAsyncHandler isACTMemberAsyncHandler;
   private TeamProjectsModalWidget teamProjectsModalWidget;
   private PortalGinInjector ginInjector;
-  private FeatureFlagConfig featureFlagConfig;
```
jay-hodgson, line 64: Clean up member variable, which is not used (config is only referenced during construction)

src/main/java/org/sagebionetworks/web/client/presenter/TeamPresenter.java
```diff
@@ -229,6 +227,7 @@ public void onSuccess(TeamBundle result) {
           view.setTeam(team, teamMembershipStatus);
           Boolean canRequestMembership = team.getCanRequestMembership();
           boolean isLockedDown =
+            !canPublicJoin &&
```
jay-hodgson, line 230: Team is not logged down if public can join!

src/main/java/org/sagebionetworks/web/client/widget/team/controller/TeamEditModalWidget.java
```diff
@@ -159,18 +159,23 @@ private void show() {
     view.clear();
     view.configure(team);
     //SWC-6259
-    Boolean canPublicJoin = team.getCanPublicJoin();
+    //SWC-7000
+    Boolean canPublicJoin = team.getCanPublicJoin() == null
```
jay-hodgson, line 163: Match logic to TeamPresenter.  canPublicJoin is true iff team.getCanPublicJoin().
Is locked down only if !canPublicJoin and team.getCanRequestMembership()===false

src/test/java/org/sagebionetworks/web/unitclient/widget/entity/team/controller/TeamEditModalWidgetTest.java
```diff
@@ -431,6 +431,16 @@ public void testConfigurePublicCanJoin() {
     verify(mockTeam).setCanRequestMembership(false);
   }
 
+  @Test
+  public void testConfigurePublicCanJoinCanRequestMembershipFalse() {
```
jay-hodgson, line 435: All existing tests still pass.  Add test for failure case, when canRequestMembership == false (not null).  This failed before the change, and passes now.

src/main/java/org/sagebionetworks/web/client/widget/team/controller/TeamEditModalWidget.java
nickgros, line 163: These should never be null since https://sagebionetworks.jira.com/browse/PLFM-8304 was done, are you seeing a regression?

nickgros (approved):

PR #5472, opened 2024-07-26
SWC-6800 - Integrate new Entity ACL Editor

nickgros:
Requires release of a new SRC


src/main/java/org/sagebionetworks/web/client/FeatureFlagConfig.java
```diff
@@ -26,11 +25,11 @@ public FeatureFlagConfig(JSONObject config, CookieProvider cookieProvider) {
     this.cookieProvider = cookieProvider;
   }
 
-  public boolean isFeatureEnabled(String featureName) {
+  public boolean isFeatureEnabled(FeatureFlagKey feature) {
```
nickgros, line 28: updated signature to be more type-safe by requiring the enum object

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -2923,12 +2933,18 @@ public Place createDeletePlace() {
...
+    ) {
+      getEntityAccessControlListModalWidget()
+        .configure(entity.getId(), this::fireEntityUpdatedEvent);
+      getEntityAccessControlListModalWidget().setOpen(true);
+    } else {
+      getAccessControlListModalWidget()
+        .configure(entity, permissions.getCanChangePermissions());
+      this.getAccessControlListModalWidget()
+        .showSharing(this::fireEntityUpdatedEvent);
+    }
```
nickgros, line 2947: Use new interface/impl when the feature flag is enabled.

jay-hodgson (approved):

PR #5478, opened 2024-07-30
SWC-6945

jay-hodgson:


nickgros (approved):

src/main/java/org/sagebionetworks/web/client/MarkdownItImpl.java
```diff
@@ -30,7 +30,7 @@ private static final native String _markdown2Html(
 						$wnd.markdownitSynapseHeading,
```
jay-hodgson, line 30: I just realized that this is a little hacky, since the parameters no longer line up for init_markdown_it().  It works because all that is done with markdownitEmphasisAlt and markdownitInlineComments is md.use(pluginName) (no special handling for these plugins, or markdownitBr).  Once we pull in a new SRC, all the parameters will align again.  No harm with the current state though.

PR #5481, opened 2024-08-01
SWC-7011

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/view/HomeViewImpl.java
```diff
@@ -26,34 +31,54 @@ public interface HomeViewImplUiBinder
...
+        globalAppState.handleRelativePathClick(href);
+      });
+      component =
+        React.createElementWithSynapseContext(
+          SRC.SynapseComponents.SynapseHomepageV2,
+          props,
+          propsProvider.getJsInteropContextProps()
+        );
+    } else {
+      //TODO: SWC-6999: Once V2 is released, delete this conditional
```
nickgros, line 72: This should be easy to catch when we remove the feature flag 😁

nickgros (approved):

PR #5485, opened 2024-08-12
SWC-6800 - Remove `onViewSharingSettingsClicked` integrations 

nickgros:
depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1122


nickgros:
Should also be able to remove the feature flag and old code, since the change this depends on also expects the feature flag to be removed.

PR #5486, opened 2024-08-12
SWC-7021

jay-hodgson:


jay-hodgson:
Clean-up, requires the SRC >= 3.3.7 in order to be in sync

nickgros (approved):

PR #5487, opened 2024-08-12
SWC-7019

jay-hodgson:
Rework Tables list html and css


jay-hodgson:
After changes

Screen Shot 2024-08-12 at 3 07 52 PM

Screen Shot 2024-08-12 at 3 08 08 PM

jay-hodgson:
Instead of each row being a Table, make each row a table row (tr) in the table body (tbody).  And fix css so overflow is handled properly (text overflow is ellipsis) and things line up.

jay-hodgson:
Note the border was added after the screenshots were taken (between the table header and body rows)

nickgros (approved):

PR #5491, opened 2024-08-15
SWC-7030

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -82,18 +81,17 @@ public void onSuccess() {
...
                     public void onSuccess(JSONObjectAdapter config) {
                       ginjector
-                        .getSessionStorage()
-                        .setItem(
-                          WebConstants.PORTAL_FEATURE_FLAG_SESSION_STORAGE_KEY,
-                          config.toString()
-                        );
+                        .getFeatureFlagConfig()
+                        .setJson(config.toString());
+                      continueInit();
```
jay-hodgson, line 94: Terrible diff below, but all of the initialization code below (that used to run in parallel) has been moved into continueInit() so that it runs serially.  This is to fix the race condition

src/main/java/org/sagebionetworks/web/client/FeatureFlagConfig.java
```diff
@@ -3,18 +3,25 @@
...
 
   private final CookieProvider cookieProvider;
-  private final JSONObject config;
+  private JSONObject config;
 
-  public FeatureFlagConfig(String json, CookieProvider cookieProvider) {
-    JSONValue parsed = JSONParser.parseStrict(json);
-    config = parsed.isObject();
+  @Inject
+  public FeatureFlagConfig(CookieProvider cookieProvider) {
```
jay-hodgson, line 15: Removed the FeatureFlagConfigProvider and made the FeatureFlagConfig a Singleton (from Guice's dependency injection perspective)

jay-hodgson: Also removed the use of session storage.  The feature flags are initialized when the app is loaded.  One can imagine updating the FeatureFlagConfig with another value, that would disappear as soon as you reload the app.  Maybe this isn't the right behavior, but we don't have this UI right now to validate with users.

src/main/java/org/sagebionetworks/web/client/Portal.java
jay-hodgson: I'd like to convert these into Futures, and wait for them to process in parallel, but I felt like this would be too high risk to patch the staging site on Thursday/Friday, especially when we have this simple (but less optimal) solution.

nickgros: Is there another ticket for this follow-up work?

nickgros (approved):
Thank you for catching and fixing this!

src/main/java/org/sagebionetworks/web/client/Portal.java
jay-hodgson: I'll open one up

PR #5494, opened 2024-08-15
SWC-7032

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -161,50 +167,57 @@ private void continueInit() {
...
+                                // clears out the web storage (cache) if this is the case.
+                                ginjector
+                                  .getWebStorageMaxSizeDetector()
+                                  .start();
+                              }
+                            }
+                          );
+                        } catch (Exception e1) {
+                          // TODO Auto-generated catch block
+                          e1.printStackTrace();
```
nickgros: Is this meant to just allow the feature flag call to fail (since we don't have configs for local dev environments)? Should this `try`/`catch` only wrap the feature flag-related logic?

nickgros (commented):
Future refactoring looks good, but one question about exception handling

src/main/java/org/sagebionetworks/web/client/Portal.java
jay-hodgson: Yes and yes.  Updated the code

nickgros (approved):

PR #5497, opened 2024-08-22
SWC-7038 - fix e2e tests

nickgros:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1149

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
@@ -7,6 +7,7 @@ import com.google.gwt.event.dom.client.ClickEvent;
 import com.google.gwt.event.dom.client.ClickHandler;
 import com.google.gwt.event.dom.client.HasClickHandlers;
 import com.google.gwt.event.shared.HandlerRegistration;
+import com.google.gwt.user.client.Timer;
 import com.google.gwt.user.client.ui.ComplexPanel;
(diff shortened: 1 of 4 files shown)


src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -175,9 +176,17 @@ protected void onLoad() {
...
+      // Asynchronously schedule unmounting the root to allow React to finish the current render cycle.
+      // https://github.com/facebook/react/issues/25675
+      Timer t = new Timer() {
+        @Override
+        public void run() {
+          root.unmount();
+          root = null;
+        }
+      };
+      t.schedule(0);
```
nickgros, line 188: Fixes an issue where changing the type of a table column would crash the app.

I think what was happening was the `EntityUpdatedEvent` was triggered by a React callback, which triggered a place reload, and this unload method ended up trying to remove the root DOM node at the same time that React was still using it.

By scheduling `unmount` instead of executing it synchronously, React gets a chance to finish the current render cycle, and now there's no longer a race condition.

jay-hodgson: Sounds good.  And thank you for fixing all of the breaking changes that I've made (experimental mode, cookie banner, etc)!

jay-hodgson (approved):

PR #5500, opened 2024-08-28
SWC-7040 - Bump Java version to 11

nickgros:
Also bump Guice to 6, Mockito to 2, and lock guice-assistedinject to 6 to override Gin.

Migrate many tests to be compatible with Mockito 2. The main reason tests had to change was because `anyString()` and `any(Foo.class)` matchers do not allow `null` as a valid match in v2, but they do in v1. In these cases, just use `any()`

Also fix the dev scripts so now `yarn dev` should launch the GWT Codeserver and a Tomcat 9 container that will re-compile on refresh if there's a source code change.

We should be able to upgrade to Java 17 at a later date just by upgrading the JDK and compile target.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "@types/uuid": "^9.0.7",
     "@typescript-eslint/eslint-plugin": "^6.4.0",
     "@typescript-eslint/parser": "^6.17.0",
+    "concurrently": "^8.2.2",
     "dotenv": "^16.3.1",
     "eslint": "^8.0.1",
     "husky": ">=6",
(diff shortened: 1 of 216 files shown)


package.json
```diff
@@ -54,16 +55,17 @@
...
     "*.{*}": "prettier --write"
   },
   "scripts": {
     "build": "mvn -B package --file pom.xml",
-    "dev": "npm-run-all --parallel dev:tomcat dev:codeserver",
-    "dev:tomcat": "docker pull tomcat:9.0; docker run --name swc-dev --rm -p 8888:8080 -v \"/$(pwd)/target/portal-develop-SNAPSHOT/:/usr/local/tomcat/webapps/ROOT/\" -v \"/$HOME/.m2/settings.xml\":/root/.m2/settings.xml tomcat:9.0",
-    "dev:codeserver": "mvn gwt:run-codeserver -Dgwt.compiler.skip=true -Dnoserver=true -Dwar=/$(pwd)/target/portal-develop-SNAPSHOT -Dstyle=PRETTY",
+    "dev": "concurrently -k -n \"CODESERVER,TOMCAT\" -c \"auto,auto\" \"yarn dev:codeserver\" \"yarn dev:tomcat\"",
+    "dev:codeserver": "mvn clean gwt:run-codeserver",
+    "dev:tomcat": "wait-on tcp:9876 && docker pull tomcat:9.0; docker run --name swc-dev --rm -p 8888:8080 -v \"/$(pwd)/target/portal-develop-SNAPSHOT/:/usr/local/tomcat/webapps/ROOT/\" -v \"/$HOME/.m2/settings.xml\":/root/.m2/settings.xml tomcat:9.0",
```
nickgros, line 68: `wait-on` will prevent Tomcat from launching until the code server is ready

src/test/java/org/sagebionetworks/web/client/SynapseJavascriptClientTest.java
```diff
@@ -138,7 +138,7 @@
 import org.sagebionetworks.web.shared.exceptions.UnauthorizedException;
 import org.sagebionetworks.web.shared.exceptions.UnknownErrorException;
 
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(MockitoJUnitRunner.Silent.class)
```
nickgros, line 141: Without `Silent`, mockito fails tests that have mocks that are not called. Which is great, but a ton of unnecessary work to fix.

src/test/java/org/sagebionetworks/web/client/SynapseJavascriptClientTest.java
```diff
@@ -225,9 +225,7 @@ public void before() throws RequestException {
     when(mockJsniUtils.getAccessTokenCookieUrl())
       .thenReturn(SESSION_COOKIE_URL);
 
-    when(
-      mockRequestBuilder.sendRequest(anyString(), any(RequestCallback.class))
-    )
+    when(mockRequestBuilder.sendRequest(any(), any(RequestCallback.class)))
```
nickgros, line 228: For nearly all of these test changes, the mock was called where the matcher received `null`. In Mockito v1, `anyString()` matched null. In Mockito v2, it does not. Replace it with `any()`.

jay-hodgson: Ah, now I understand why my newly checked in test with anyLong() may fail :)

jay-hodgson (approved):

PR #5502, opened 2024-08-28
SWC-6763: integrate reworked SRC error page

jay-hodgson:


src/test/java/org/sagebionetworks/web/unitclient/presenter/EntityPresenterTest.java
```diff
@@ -361,7 +361,7 @@ public void testClear() {
   public void testShow403() {
     entityPresenter.setEntityId("123");
     entityPresenter.show403();
-    verify(mockSynAlert).show403(anyString());
+    verify(mockSynAlert).show403(anyString(), anyLong());
```
nickgros, line 364: I hope these matchers don't break when we go to Mockito 2!

nickgros (approved):

PR #5507, opened 2024-08-29
SWC-7047

jay-hodgson:
Add TrustCenter to Synapse that pulls content from GitHub Governance markdown documents and renders using the existing SynapseStandaloneWikiView


jay-hodgson:
Screen Shot 2024-08-29 at 10 57 55 AM

Screen Shot 2024-08-29 at 11 03 25 AM

src/main/resources/org/sagebionetworks/web/client/view/SynapseStandaloneWikiViewImpl.ui.xml
```diff
@@ -10,16 +10,16 @@
     type='org.sagebionetworks.web.client.SageImageBundle'
   />
 
-  <bh:Div>
```
jay-hodgson, line 13: Noticed that there was no space above or below wiki content, and the synapse alert was previously shown below the 400px height blank markdown container

src/main/java/org/sagebionetworks/web/client/presenter/TrustCenterPresenter.java
```diff
@@ -0,0 +1,126 @@
...
+      showFileContent(documentUrl);
+    }
+  }
+
+  public void showFileContent(String url) {
+    requestBuilder.configure(RequestBuilder.GET, url);
+    requestBuilder.setHeader(
+      WebConstants.CONTENT_TYPE,
+      WebConstants.TEXT_PLAIN_CHARSET_UTF8
+    );
```
nickgros, line 87: I doubt that we will have many users viewing this page, so it's unlikely to cause problems, but I would consider moving this to the servlet and caching it.

nickgros (approved):
Looks great! One optional suggestion

src/main/java/org/sagebionetworks/web/client/presenter/TrustCenterPresenter.java
jay-hodgson: An older (about 3 years old) question about this topic suggests that they allow may impose a limit of 5000 requests per hour, per IP address (which would be the client IP).  
https://stackoverflow.com/questions/66522261/does-github-rate-limit-access-to-public-raw-files

PR #5508, opened 2024-08-29
Upgrade to SRC 3.3.10, fix e2e tests (again)

nickgros:


src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -143,15 +143,22 @@ private void injectChildWidgetsIntoComponent() {
...
+      // Asynchronously schedule creating a root in case React is still rendering and may unmount the current root
+      Timer t = new Timer() {
+        @Override
+        public void run() {
+          maybeCreateRoot();
+          // Resynchronize with the DOM
+          root.render(reactElement);
+        }
+      };
+      t.schedule(0);
```
nickgros, line 161: Follow up from #5497, the call to `render` also needs to be scheduled or else the old root might still exist.

src/main/webapp/Portal.html
```diff
@@ -196,7 +193,7 @@
       loadJs(cdnEndpoint + 'generated/react-router-dom.min.js')
       loadJs(cdnEndpoint + 'generated/universalCookie.min.js')
       loadJs(cdnEndpoint + 'generated/react-bootstrap.min.js')
-      loadJs(srcPath) // We don't load SRC's scss because we include it in our own compiled scss
+      loadJs(cdnEndpoint + 'generated/synapse-react-client.production.min.js') // We don't load SRC's scss because we include it in our own compiled scss
```
nickgros, line 196: dev build was removed in latest SRC

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
nickgros: Fixes an e2e test

jay-hodgson (approved):

PR #5509, opened 2024-08-30
SWC-7049 - Fix remounting/cleanup issue with React components

nickgros:
Pulling #5497 and https://github.com/Sage-Bionetworks/SynapseWebClient/commit/d8211e6cfeaaf423d3149abac3a33b92542f73eb from #5508 into release-511 to fix SWC-7049

### DIFF
--- a/e2e/files.spec.ts
+++ b/e2e/files.spec.ts
@@ -356,7 +356,10 @@ testAuth.describe('Files', () => {
       await expect(
         userPage.getByRole('heading', { name: userProject.name }),
       ).toBeVisible()
-      await userPage.getByRole('link', { name: 'Files', exact: true }).click()
+      await userPage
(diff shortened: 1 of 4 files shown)


e2e/files.spec.ts
```diff
@@ -356,7 +356,10 @@ testAuth.describe('Files', () => {
       await expect(
         userPage.getByRole('heading', { name: userProject.name }),
       ).toBeVisible()
-      await userPage.getByRole('link', { name: 'Files', exact: true }).click()
+      await userPage
+        .getByRole('link', { name: 'Files', exact: true })
+        .first()
+        .click()
```
nickgros, line 362: E2E test changes are from #5497, which also had one of the React component fixes in the same commit, so it ended up also being pulled in here.

jay-hodgson (approved):

PR #5513, opened 2024-09-06
SWC-6663: build and publish SWC artifact from GH

xschildw:
This PR should enable building SWC and publishing the artifact from Github Workflow

- action.yml is modified to accept an mvn goal and an optional pom version (to version the artifact)
- build-test-e2e.yml now calls the action with 'package' goal (as it was hardcoded in action previously)
- main.yaml
  - on pull request to dev or release branch, call build with 'test' goal
  - on push request to dev or release branch, version accordingly and call build with 'deploy' goal ('package' first to test)
- release.yaml releases a branch to production 
  - tag branch with v.1 version (staging is .0)
  - create a new v+1 branch and tags it with v+1.0
  - starts a build off v+1 branch
- maven.yml is replaced by main.yaml


.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+          next_release_number=$((release_number + 1))
+          new_branch="release-${next_release_number}"
+          git checkout -b $new_branch
+
+          new_tag="${next_release_number}.0"
+          git tag -a $new_tag -m stack-${next_release_number}-staging
+          git push origin $new_branch
+          echo "pomversion=$new_tag" >> $GITHUB_ENV
+
+      - id: build_new_release_branch
```
xschildw, line 52: The original idea was to just push in the step above and have the trigger do the build but the trigger does not get fired if the push is coming from an action (unless it's done with a PAT), so starting a build of the new branch here.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+        uses: ./.github/workflows/build
+        with:
+          mvn_goal: package
```
xschildw, line 55: package first to test

.github/workflows/main.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Set POM version for dev branch
+        if: ${{ github.ref == 'refs/heads/develop' }}
+        run: echo "pomversion=$(date +%Y-%m-%d)-$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+      - name: Set POM version for release branch
+        if: ${{ startsWith(github.ref, 'refs/heads/release-') }}
+        run: echo "pomversion=$(date +%Y-%m-%d)-$(git describe --tags)" >> $GITHUB_ENV
+
+      - uses: ./.github/workflows/build
+        with:
+          mvn_goal: package
```
xschildw, line 28: package first to test

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+      - id: git_config
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "actions@github.com"
+
+      - id: push_prod_tag
+        run: |
+          tag_name="${release_number}.1"
+          git tag -a $tag_name -m stack-${release_number}-prod
```
nickgros, line 37: I feel like I'm just missing some context:
* How do we build + tag patch versions today? Will this be compatible with that process (or is that an intended future change)?
* What's the purpose of these new `-prod`/`-staging` tags?

.github/workflows/release.yaml
nickgros: The commit should already be tested, so this seems fine. When would we run this workflow?

.github/workflows/release.yaml
xschildw: Now I manually run a shell script that takes the repo, the current and next release number (e.g. 'create-branch ~/bin/create-branch.sh SynapseWebClient 512 513') to tag and create the branches each week. It clones the repo and does essentially the same steps as here (simpler since I give it the versions...). I figured it'd be nice to just click a button right in GH since I'm there re-checking that everything is merged anyway.

The '-prod' and 'staging' are just added to the comment on the git tag (to remind that .0 lives while that version is on staging and .1  while it's in prod).

To version the artifacts, we use 2 scripts (one for dev branch, one for release branches) that modify the version in all the pom.xml files, I would like to replace them by the optional 'mvn versions:set' step in build/actions.yml.

nickgros (approved):

PR #5515, opened 2024-09-11
PORTALS-3204: Use SynapseChat from SRC (when available)

jay-hodgson:
Notes: Chat place is not in the open places set, so will prompt for login.
Chat place uses a parameterized token, so we can extend to support other agent IDs later if we want (and chatbot names).
Since these services are not currently available, this component will never load properly (we only get it to load by using mock service responses).  With this code, hopefully we will not need to update SWC for this again (other than updating the SRC version!)

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/place/ChatPlace.java
+++ b/src/main/java/org/sagebionetworks/web/client/place/ChatPlace.java
@@ -1,0 +1,29 @@
+package org.sagebionetworks.web.client.place;
+
+import com.google.gwt.place.shared.PlaceTokenizer;
+import com.google.gwt.place.shared.Prefix;
+
+public class ChatPlace extends ParameterizedPlace {
+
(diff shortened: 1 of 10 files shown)


nickgros (approved):

src/main/java/org/sagebionetworks/web/client/place/ChatPlace.java
```diff
@@ -0,0 +1,29 @@
+package org.sagebionetworks.web.client.place;
+
+import com.google.gwt.place.shared.PlaceTokenizer;
+import com.google.gwt.place.shared.Prefix;
+
+public class ChatPlace extends ParameterizedPlace {
+
+  public static final String INITIAL_MESSAGE = "initalMessage";
```
nickgros:
```suggestion
  public static final String INITIAL_MESSAGE = "initialMessage";
```

PR #5520, opened 2024-09-16
SWC-6923 - Use React implementation for verification rejection dialog

nickgros:


.husky/pre-commit
```diff
@@ -1,4 +1 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
```
nickgros, line 3:
```
husky - DEPRECATED

Please remove the following two lines from .husky/pre-commit:

#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"

They WILL FAIL in v10.0.0
```

src/main/resources/portal.properties
```diff
@@ -2,8 +2,10 @@ org.sagebionetworks.portal.challenge_synapse_id = syn1968135
 org.sagebionetworks.portal.challenge_writeup_synapse_id = syn2027095
 org.sagebionetworks.portal.default_thread_id = 234
 
-org.sagebionetworks.portal.ACT_profile_validation_rejection_reasons_json = syn24977268
+org.sagebionetworks.portal.ACT_profile_validation_rejection_reasons_json = syn24977268
```
nickgros:
```suggestion
org.sagebionetworks.portal.ACT_profile_validation_rejection_reasons_json = syn24977268
```

src/main/java/org/sagebionetworks/web/client/widget/verification/VerificationSubmissionWidget.java
```diff
@@ -425,22 +434,14 @@ public void suspendVerification() {
   }
 
   private void rejectSuspendVerification() {
-    RejectReasonWidget promptModal = ginInjector.getRejectReasonWidget();
```
jay-hodgson, line 428: Can RejectReasonWidget be deleted?

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/verification/VerificationSubmissionWidget.java
nickgros: It's still used by some data access request widgets, though those may also be deletable

PR #5525, opened 2024-09-19
SWC-7081: pass the invitation token to One Sage

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/place/users/RegisterAccount.java
```diff
@@ -1,19 +1,17 @@
 package org.sagebionetworks.web.client.place.users;
 
-import com.google.gwt.place.shared.Place;
 import com.google.gwt.place.shared.PlaceTokenizer;
 import com.google.gwt.place.shared.Prefix;
+import org.sagebionetworks.web.client.place.ParameterizedPlace;
 
-public class RegisterAccount extends Place {
+public class RegisterAccount extends ParameterizedPlace {
```
jay-hodgson, line 7: Change to a parameterized place so we can pass both the email and the signed token to the register account page (which will be passed to One Sage)

src/main/java/org/sagebionetworks/web/client/place/users/RegisterAccount.java
```diff
@@ -1,19 +1,17 @@
...
 
-  private String token;
+  public static final String EMAIL_QUERY_PARAM = "email";
+  public static final String MEMBERSHIP_INVTN_QUERY_PARAM =
+    "membershipInvtnSignedToken";
 
   public RegisterAccount(String token) {
```
jay-hodgson, line 13: Note that there are cases in the codebase that point to RegisterAccount with a default token ("0").  I tested that the presenter just redirects to the OneSage registration page without any additional query params (besides the appId used to brand the One Sage site)

src/main/java/org/sagebionetworks/web/client/presenter/users/RegisterAccountPresenter.java
```diff
@@ -21,22 +21,21 @@ public void start(AcceptsOneWidget panel, EventBus eventBus) {}
...
-      Window.Location.replace(
-        WebConstants.ONESAGE_PRODUCTION_URL +
-        "/register1?" +
-        WebConstants.ONESAGE_SYNAPSE_APPID_QUERY_PARAM
-      );
+    String emailInvitationToken = place.getParam(
+      RegisterAccount.MEMBERSHIP_INVTN_QUERY_PARAM
+    );
+    String email = place.getParam(RegisterAccount.EMAIL_QUERY_PARAM);
+    StringBuilder targetUrl = new StringBuilder();
```
jay-hodgson, line 28: Build up the target URL based on what parameters we have

src/main/java/org/sagebionetworks/web/client/presenter/users/RegisterAccountPresenter.java
```diff
@@ -21,22 +21,21 @@ public void start(AcceptsOneWidget panel, EventBus eventBus) {}
...
+    targetUrl.append(WebConstants.ONESAGE_PRODUCTION_URL);
+    targetUrl.append("/register1?");
+    targetUrl.append(WebConstants.ONESAGE_SYNAPSE_APPID_QUERY_PARAM);
+
+    if (emailInvitationToken != null) {
+      targetUrl.append("&signedToken=" + emailInvitationToken);
```
jay-hodgson, line 34: This is where the magic happens.   One Sage already has logic to read a signed token from the query parameters and store it in Local Storage.  Then on the Account Created page, it already has logic to read this token in, check if it's of this type, and even automatically joins the target team (as well as changing the link to say something like "Take me to the Team"!).  This workflow functions perfectly (I just tested it), once we send over this signed token!

nickgros (approved):

PR #5528, opened 2024-09-23
SWC-6776 - Show ACL editor after successful upload

nickgros:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1229

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
@@ -12,6 +12,7 @@ import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.Widget;
 import com.google.inject.Inject;
 import java.util.List;
+import org.sagebionetworks.repo.model.AccessControlList;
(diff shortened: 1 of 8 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
```diff
@@ -790,7 +791,22 @@ public void onSuccess(Entity result) {
         entity = result;
         view.showInfo(DisplayConstants.TEXT_LINK_SUCCESS);
         if (successHandler != null) {
-          successHandler.onSuccessfulUpload();
+          synapseClient.getEntityBenefactorAcl(
```
jay-hodgson, line 794: ideally this would be a jsClient call instead of a gwt rpc

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
nickgros: Agreed. I think we'll eventually be re-implementing this widget based on the other open tickets, so I went with the fastest option.

PR #5537, opened 2024-10-01
SWC-7100: add a new TrustCenterView, and rely on a new SRC component …

jay-hodgson:
…for GitHub MD resources (which uses the latest tag)


jay-hodgson:
Showing the local client integration test that uses the new SRC component (showing network activity to GitHub):
Screenshot 2024-10-01 at 4 40 27 PM

jay-hodgson:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1247

nickgros (approved):

PR #5539, opened 2024-10-03
SWC-6533

jay-hodgson:
update LoginPresenter to often go to the One Sage root page to handle login instead of showing it here

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/GWTWrapperImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/GWTWrapperImpl.java
@@ -20,6 +20,8 @@ import com.google.gwt.user.client.ui.HTMLPanel;
 import com.google.gwt.user.datepicker.client.CalendarUtil;
 import com.google.gwt.xhr.client.XMLHttpRequest;
 import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
 import org.sagebionetworks.web.client.utils.Callback;
 
 public class GWTWrapperImpl implements GWTWrapper {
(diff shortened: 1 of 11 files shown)


src/main/java/org/sagebionetworks/web/client/GWTWrapperImpl.java
```diff
@@ -241,6 +243,36 @@ public String getFriendlySize(double size, boolean abbreviatedUnits) {
...
+  private static Map<String, String> hostName2OneSageSite = new HashMap<>();
+
+  static {
+    hostName2OneSageSite.put(
+      "staging.synapse.org",
+      "https://staging.accounts.synapse.org/?appId=staging.synapse.org"
+    );
+    hostName2OneSageSite.put(
+      "portal-dev.dev.sagebase.org",
+      "https://accounts-dev.dev.sagebase.org/?appId=dev.synapse.org"
```
jay-hodgson, line 255: Does not yet exist, see IT-3848 and PORTALS-3203

src/main/java/org/sagebionetworks/web/client/GWTWrapperImpl.java
```diff
@@ -241,6 +243,36 @@ public String getFriendlySize(double size, boolean abbreviatedUnits) {
...
+    );
+    hostName2OneSageSite.put(
+      "localhost",
+      "http://localhost:3000/?appId=localhost"
+    );
+    hostName2OneSageSite.put(
+      "127.0.0.1",
+      "http://127.0.0.1:3000/?appId=localhost"
+    );
```
nickgros, line 264: Good idea to handle both

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -612,6 +613,14 @@ public void onPreviewNativeEvent(NativePreviewEvent event) {
     }
   }
 
+  @Override
+  public void gotoLoginPage() {
+    //tell One Sage to return to the current url
+    cookieProvider.setCookie(ONESAGE_REDIRECT_COOKIE_KEY, gwt.getCurrentURL());
```
nickgros: ditto feedback from https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1252 , consider expiring

src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
```diff
@@ -46,79 +46,22 @@ public class LoginPresenter
...
     this.authenticationController = authenticationController;
     this.globalApplicationState = globalApplicationState;
     this.synAlert = synAlert;
-    this.popupUtils = popupUtils;
     view.setSynAlert(synAlert);
     view.setPresenter(this);
   }
 
-  @Override
-  public void onAcceptTermsOfUse() {
```
nickgros, line 69: So if I sign in to OneSage and have not accepted the terms of use, that app already handles prompting today (before new ToS implementation)?

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/presenter/LoginPresenter.java
jay-hodgson: It should.  The ApplicationSessionManager attempts to initialize the session (used in OneSage AppWrapper).  In doing so, it tries to get your User Profile, and looks for the "Terms of use have not been signed" error.  It sets the ApplicationSessionContext acceptsTermsOfUse value.  If this value is false, then in AppInitializer, it will redirect you to /authenticated/signTermsOfUse, showing you the TermsOfUsePage

jay-hodgson:
In stack 518, send to One Sage for login!

PR #5540, opened 2024-10-03
SWC-6776, SWC-7096 - fix issue where parentId was null

nickgros:
Also fix a race condition

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
+++ b/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
@@ -3295,4 +3295,33 @@ public class SynapseJavascriptClient {
     String url = getAuthServiceUrl() + SESSION_ACCESS_TOKEN;
     return getFuture(cb -> doDelete(url, cb));
   }
+
+  public FluentFuture<AccessControlList> getEntityBenefactorAcl(
+    String entityId
+  ) {
+    // Retrieving the benefactor ACL is always permitted regardless of permissions, so only retrieve that part of the bundle.
+    EntityBundleRequest request = new EntityBundleRequest();
+    request.setIncludeBenefactorACL(true);
+    String url = getRepoServiceUrl() + ENTITY + "/" + entityId + BUNDLE2;
+
+    return getFuture(cb ->
+      doPost(
+        url,
+        request,
+        OBJECT_TYPE.EntityBundle,
+        true,
+        new AsyncCallback<EntityBundle>() {
+          @Override
+          public void onFailure(Throwable caught) {
+            cb.onFailure(caught);
+          }
+
+          @Override
+          public void onSuccess(EntityBundle result) {
+            cb.onSuccess(result.getBenefactorAcl());
+          }
+        }
+      )
+    );
+  }
 }
--- a/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
+++ b/src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
@@ -3296,32 +3296,31 @@ public class SynapseJavascriptClient {
     return getFuture(cb -> doDelete(url, cb));
   }
 
-  public FluentFuture<AccessControlList> getEntityBenefactorAcl(
-    String entityId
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -3295,4 +3295,33 @@ public FluentFuture<Void> deleteSessionAccessToken() {
     String url = getAuthServiceUrl() + SESSION_ACCESS_TOKEN;
     return getFuture(cb -> doDelete(url, cb));
   }
+
+  public FluentFuture<AccessControlList> getEntityBenefactorAcl(
```
nickgros: moved HTTP call to run in client instead of RPC to simplify debugging

src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
```diff
@@ -791,25 +791,27 @@ public void onSuccess(Entity result) {
...
+                  // Upload was still a success, benefactor ID is not required to continue
+                  successHandler.onSuccessfulUpload(null);
+                  entityUpdated();
+                }
+              },
+              directExecutor()
+            );
         }
-
-        entityUpdated();
```
nickgros, line 812: Moved `entityUpdated` call inside the callback, otherwise there was a race condition where the uploader would not be cleaned up

src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
```diff
@@ -1025,24 +1027,27 @@ private void uploadSuccess() {
     view.resetToInitialState();
     resetUploadProgress();
     if (successHandler != null) {
-      synapseClient.getEntityBenefactorAcl(
-        parentEntityId,
```
nickgros, line 1029: For some reason, I tried to use `parentEntityId`, which is null sometimes (maybe I assumed `entityId` might be null?). We can just use the current entity ID. This is what caused the 500 error on the RPC

src/main/java/org/sagebionetworks/web/client/SynapseJavascriptClient.java
```diff
@@ -3295,4 +3295,32 @@ public FluentFuture<Void> deleteSessionAccessToken() {
     String url = getAuthServiceUrl() + SESSION_ACCESS_TOKEN;
     return getFuture(cb -> doDelete(url, cb));
   }
+
+  public void getEntityBenefactorAcl(
```
nickgros, line 3299: Moved call from RPC to direct API call to simplify debugging

src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
```diff
@@ -791,25 +791,30 @@ public void onSuccess(Entity result) {
...
-          synapseClient.getEntityBenefactorAcl(
+          jsClient.getEntityBenefactorAcl(
             result.getId(),
             new AsyncCallback<AccessControlList>() {
               @Override
               public void onSuccess(AccessControlList benefactorAcl) {
-                successHandler.onSuccessfulUpload(benefactorAcl.getId());
+                if (benefactorAcl.getId().equals(entity.getId())) {
+                  // Don't show the ACL modal if the entity is its own benefactor
+                  successHandler.onSuccessfulUpload(null);
```
nickgros, line 801: Handle new edge case

src/main/java/org/sagebionetworks/web/client/widget/entity/download/Uploader.java
```diff
@@ -791,25 +791,30 @@ public void onSuccess(Entity result) {
...
+                } else {
+                  successHandler.onSuccessfulUpload(benefactorAcl.getId());
+                }
+                entityUpdated();
```
nickgros, line 805: Move `entityUpdated` call inside callback to fix race condition issue where the Uploader would be partially torn down before `successHandler` is called

jay-hodgson (approved):

PR #5542, opened 2024-10-04
SWC-6776 - Put feature in experimental mode under feature flag

nickgros:
note: not planning on adding the feature flag key to the stack builder

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/FeatureFlagKey.java
+++ b/src/main/java/org/sagebionetworks/web/client/FeatureFlagKey.java
@@ -37,6 +37,9 @@ public enum FeatureFlagKey {
   // If enabled, use the re-implemented ACL Editor for entities
   REACT_ENTITY_ACL_EDITOR("REACT_ENTITY_ACL_EDITOR"),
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/sagebionetworks/web/client/FeatureFlagKey.java
```diff
@@ -37,6 +37,9 @@ public enum FeatureFlagKey {
   // If enabled, use the re-implemented ACL Editor for entities
   REACT_ENTITY_ACL_EDITOR("REACT_ENTITY_ACL_EDITOR"),
 
+  // If enabled, sharing settings will appear in a dialog immediately after uploading one or more files.
+  SHOW_SHARING_SETTINGS_AFTER_UPLOAD("SHOW_SHARING_SETTINGS_AFTER_UPLOAD"),
```
nickgros, line 41: I'm not planning on adding this to the stack builder for now, since we don't need to dynamically toggle it. This is basically in experimental mode only, but it's easy to find + remove all of the toggles for this feature when we need to.

jay-hodgson (approved):

PR #5547, opened 2024-10-14
PORTALS-3195 - Use bundled react-router, add props to ReviewerDashboard

nickgros:


pom.xml
```diff
@@ -214,7 +214,6 @@
         <!-- Plugin configuration. There are many available options, see gwt-maven-plugin 
 					documentation at codehaus.org -->
         <configuration>
-          <workDir>${project.build.directory}/gwt/codeserver</workDir>
```
nickgros, line 217: sometimes I run into issues where this dir doesn't exist, and my plugin throws an error until I manually create it. I added this line in August, so I think I can safely remove it.

package.json
```diff
@@ -30,7 +30,6 @@
     "react-dom": "18.2.0",
     "react-measure": "2.5.2",
     "react-plotly.js": "^2.6.0",
-    "react-router-dom": "^5.3.4",
```
nickgros, line 33: react-router is now  bundled in SRC

jay-hodgson (approved):

PR #5551, opened 2024-10-17
SWC-7102, SWC-7131 - Fixes for ReactComponent widgets

nickgros:


src/main/java/org/sagebionetworks/web/client/SynapseJSNIUtils.java
```diff
@@ -14,8 +14,6 @@ public interface SynapseJSNIUtils {
 
   public void highlightCodeBlocks();
 
-  void loadSummaryDetailsShim();
```
nickgros, line 18: details-shim was removed in https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5346

I missed these references to the module, so removing them here will make an error message in the console go away. This does not fix SWC-7102

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentV2.java
```diff
@@ -94,8 +94,17 @@ private void createRoot() {
 
   private void destroyRoot() {
```
nickgros, line 95: Add the same logic to ReactComponentV2.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentV2.java
```diff
@@ -94,8 +94,17 @@ private void createRoot() {
...
+      this.root = null;
+      // Asynchronously schedule unmounting the root to allow React to finish the current render cycle.
+      // https://github.com/facebook/react/issues/25675
+      Timer t = new Timer() {
+        @Override
+        public void run() {
+          oldRoot.unmount();
+        }
+      };
+      t.schedule(0);
```
nickgros, line 107: Move the timer logic to the helper method

nickgros:
This approach might not work. You can declare two roots (root1, root2) on one DOM node. If you render into root2, unmounting root1 destroys the rendered content in root2

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -50,37 +69,32 @@ public void run() {
     t.schedule(0);
   }
 
+  public void render(ReactElement<?, ?> reactElement) {
+    this.reactElement = reactElement;
+    createRootAndRender();
+  }
```
nickgros, line 75: a bit of refactoring to unify the calls to `Timer.run`/`Timer.schedule` (`setTimeout`) and make it more clear why it's being scheduled. No logical change

src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java
```diff
@@ -50,37 +69,32 @@ public void run() {
...
   }
 
   @Override
   public void clear() {
-    // clear doesn't typically call onUnload, but we want to for this element.
-    this.onUnload();
-    super.clear();
+    if (root != null) {
+      root.render(React.createElement(React.Fragment));
+    }
```
nickgros, line 97: `onUnload` was inappropriate here. This widget can only contain a React component tree (no widget children), so it's much safer to just render a fragment. This prevents bugs where the root was unmounted and destroyed when it didn't need to be.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentV2.java
```diff
@@ -92,11 +92,40 @@ private void createRoot() {
     }
   }
 
+  /**
+   * Asynchronously (in the task queue, via setTimeout) unmounts the root and sets it to null.
+   */
   private void destroyRoot() {
```
nickgros, line 98: Just as in ReactComponent (v1), move the asynchronous logic inside of `destroyRoot`

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentV2.java
```diff
@@ -181,24 +210,13 @@ public void render() {
...
+    }
 
-        // Create a fresh ReactElement tree and render it
-        componentToRender.createRoot();
-        componentToRender.root.render(componentToRender.createReactElement());
-      }
-    };
-    t.schedule(0);
+    // Schedule rendering
+    createRootAndRender();
```
nickgros, line 219: Note that even through the timers were changed so `destroyRoot` and `createRootAndRender` now trigger two timers, they end up on the same queue, so they will execute in order.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentV2.java
```diff
@@ -216,20 +234,10 @@ protected void onLoad() {
 
   @Override
   protected void onUnload() {
-    super.onUnload();
```
nickgros, line 219: The `super` method is an empty implementation, so this call is not necessary

jay-hodgson (approved):

PR #5554, opened 2024-10-21
SWC-7124 - Dynamically set OneSage appId

nickgros:


src/main/java/org/sagebionetworks/web/client/GWTWrapperImpl.java
```diff
@@ -243,36 +241,6 @@ public String getFriendlySize(double size, boolean abbreviatedUnits) {
     return DisplayUtils.getFriendlySize(size, abbreviatedUnits);
   }
 
-  private static Map<String, String> hostName2OneSageSite = new HashMap<>();
```
nickgros, line 246: Move logic to another utility

src/main/java/org/sagebionetworks/web/client/OneSageUtils.java
```diff
@@ -0,0 +1,62 @@
...
+        return "https://staging.accounts.synapse.org";
+      case "portal-dev.dev.sagebase.org":
+        return "https://accounts-dev.dev.sagebase.org";
+      case "localhost":
+      case "127.0.0.1":
+        return "http://" + Window.Location.getHostName() + ":3000";
+      default:
+        return "https://accounts.synapse.org";
+    }
+  }
```
nickgros, line 23: Separated `host` and `appId` logic so we can inject a path between them as necessary

PR #5560, opened 2024-10-23
SWC-7126 - Fix bootstrapping redirect bug 

nickgros:
SWC-7126

Fix bug where authenticationController might try to use placeController before it has been set

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/Portal.java
+++ b/src/main/java/org/sagebionetworks/web/client/Portal.java
@@ -18,6 +18,7 @@ import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.ui.RootPanel;
 import com.google.gwt.user.client.ui.SimplePanel;
 import java.util.HashMap;
+import java.util.function.Consumer;
 import org.sagebionetworks.schema.adapter.JSONObjectAdapter;
 import org.sagebionetworks.web.client.mvp.AppActivityMapper;
 import org.sagebionetworks.web.client.mvp.AppPlaceHistoryMapper;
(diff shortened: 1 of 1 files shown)


src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -92,16 +93,12 @@ public void onSuccess() {
...
-              FluentFuture<Void> checkForUserChangeFuture = ginjector
-                .getAuthenticationController()
-                .getCheckForUserChangeFuture();
 
               FluentFuture
                 .from(
                   whenAllComplete(
                     featureFlagConfigFuture,
-                    synapsePropertiesFuture,
-                    checkForUserChangeFuture
```
nickgros, line 104: We have a dependency cycle where `AuthenticationControllerImpl.checkForUserChangeFuture` may try to call `GlobalApplicationStateImpl.goTo`, but `GlobalApplicationStateImpl.goTo` relies on the `PlaceController`, which is not ready; it is initialized and set in this future's callback (on line 167 (old) / 164 (new).

We'll run the first two futures, but defer running `checkForUserChangeFuture` until the `placeController` is set

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -170,38 +167,67 @@ public void onSuccess() {
...
-                              // initialize the view default columns so that they're ready when we need them (do this by constructing that singleton object)
-                              ginjector.getViewDefaultColumns();
+                        FluentFuture
+                          .from(
+                            whenAllComplete(
+                              // Checking the session may result in a redirect, so this must be invoked after `setPlaceController`
+                              ginjector
+                                .getAuthenticationController()
+                                .getCheckForUserChangeFuture()
+                            )
```
nickgros, line 178: using this `whenAllComplete.call` pattern just the like the outer future so we can call the wrapping `onFailure` method if this fails.

src/main/java/org/sagebionetworks/web/client/Portal.java
```diff
@@ -170,38 +167,67 @@ public void onSuccess() {
...
+                              .call(
+                                () -> {
+                                  globalApplicationState.init(
```
nickgros, line 181: Finally, call `globalApplicationState.init`

jay-hodgson (approved):

PR #5587, opened 2024-11-22
SWC-7044 - Handle error when fetching feature flag configuration

nickgros:
attempt to recover on a BadRequestException by re-initializing the client

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
@@ -1,8 +1,10 @@
 package org.sagebionetworks.web.server.servlet;
 
 import com.amazonaws.services.appconfigdata.AWSAppConfigData;
+import com.amazonaws.services.appconfigdata.model.BadRequestException;
 import com.amazonaws.services.appconfigdata.model.GetLatestConfigurationRequest;
 import com.amazonaws.services.appconfigdata.model.GetLatestConfigurationResult;
+import com.amazonaws.services.appconfigdata.model.InternalServerException;
 import com.amazonaws.services.appconfigdata.model.StartConfigurationSessionRequest;
 import com.amazonaws.services.appconfigdata.model.StartConfigurationSessionResult;
 import com.google.gwt.thirdparty.guava.common.base.Supplier;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -95,11 +97,31 @@ public void initializeConfigSupplier() {
...
+    }
+    try {
+      return new JSONObjectAdapterImpl(DEFAULT_CONFIG_VALUE);
+    } catch (JSONObjectAdapterException e) {
+      logger.log(
+        Level.SEVERE,
+        "JSONObjectAdapterException occurred in default configuration",
+        e
+      );
+      throw new RuntimeException(e);
```
nickgros, line 112: If we can't serialize a constant string `"{}"`, we have bigger problems

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -95,11 +97,31 @@ public void initializeConfigSupplier() {
...
       if (configurationToken == null) {
-        logger.log(Level.SEVERE, "returning default config");
-        return new JSONObjectAdapterImpl(DEFAULT_CONFIG_VALUE);
+        logger.log(
+          Level.SEVERE,
+          "The configuration token is null, the last config value will be returned." +
+          " This usually means that the initial call to initialize the configuration session failed."
+        );
+        return getLastConfigValueOrDefault();
       }
```
nickgros, line 125: It seems like this error case will only happen when initialization failed, so I updated the log message to give context.

src/main/java/org/sagebionetworks/web/server/servlet/AppConfigServlet.java
```diff
@@ -116,21 +138,22 @@ public JSONObjectAdapter getLatestConfiguration() {
...
+    } catch (BadRequestException | InternalServerException e) {
+      // Invalid token or server error, re-initialize the session to try to recover.
+      logger.log(
+        Level.SEVERE,
+        "Failed to get latest configuration, returning last or default configuration and attempting to re-initialize the session.",
+        e
+      );
+      initializeAppConfigClient();
+      return getLastConfigValueOrDefault();
```
nickgros, line 149: In the logs, a `BadRequestException` was thrown with message `"Token not valid"`. The tokens are one-time-use and my hypothesis is that we probably used a token more than once (maybe some network failure occurred and we didn't get the response containing a new token). Attempt to recover by re-initializing the session and getting a fresh token.

jay-hodgson (approved):

PR #5589, opened 2024-11-25
SWC-7194

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -352,10 +352,7 @@ public void onError(Throwable caught) {
     headerWidget.configure();
     if (caught instanceof NotFoundException) {
       show404();
-    } else if (
-      caught instanceof ForbiddenException &&
-      authenticationController.isLoggedIn()
```
jay-hodgson, line 357: Previously, if anonymous it would call StuAlert handleException() which passed in null for the entity ID (and version).  Now, we always call show403(entityId, versionNumber), but only show the open invitations if you are logged in (see show403() in this class).

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
```diff
@@ -364,35 +361,35 @@ public void onError(Throwable caught) {
   }
 
   public void show403() {
-    if (entityId != null) {
-      synAlert.show403(entityId, versionNumber);
-    }
+    synAlert.show403(entityId, versionNumber);
```
nickgros, line 364: does this work ok if the entity ID is nonsense, like if the place is `Synapse:notASynId`? Just wondering if the null check is relevant to parsing the place token

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/presenter/EntityPresenter.java
jay-hodgson: Works with an invalid synapse ID place token (we do attempt to get the DOI information, which results in a 400, but this does not impact the UI).

PR #5590, opened 2024-11-26
SWC-7185

jay-hodgson:


jay-hodgson:
https://github.com/user-attachments/assets/7e9b0474-a0ad-4f5e-9170-f3c2f167f857

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/FilesBrowserViewImpl.java
```diff
@@ -23,6 +26,8 @@ public interface FilesBrowserViewImplUiBinder
 
   private EntityTreeBrowser entityTreeBrowser;
   private Widget widget;
+  SynapseReactClientFullContextPropsProvider propsProvider;
```
nickgros, line 29: Consistency nit
```suggestion
  private SynapseReactClientFullContextPropsProvider propsProvider;
```

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/FilesBrowserViewImpl.java
```diff
@@ -53,15 +67,46 @@ public FilesBrowserViewImpl(
...
+    } else {
+      entityTreeBrowser.configure(entityId);
+    }
+  }
+
+  public void rerenderFileBrowser(String parentContainerId) {
+    // TODO: Should not need to remount, but currently the parentContainer is not shown if the existing react component is updated
+    // TODO: Bug: The tree is not always auto-expanding (to the path to the parent container Id)
+    reactFileBrowserContainer.clear();
+    ReactComponent componentContainer = new ReactComponent();
```
nickgros, line 87: You shouldn't need to create a new ReactComponent instance, you should just be able to create one instance and call `render` on it with the new ReactElement. But it's possible you tried that and it resulted in another bug.

nickgros (approved):
Looking forward to trying this!

src/main/java/org/sagebionetworks/web/client/widget/entity/browse/FilesBrowserViewImpl.java
jay-hodgson: Yes, tried.  Noted a couple of bugs in TODO comments that would need to be addressed if we're moving forward with this solution

PR #5591, opened 2024-11-27
SWC-7196 - Use correct GlobalApplicationState singleton instance

nickgros:
Note: targeting staging (523)


src/main/java/org/sagebionetworks/web/client/widget/entity/download/UploadDialogWidgetV2.java
```diff
@@ -27,7 +26,7 @@ public class UploadDialogWidgetV2 extends Widget {
 
   @Inject
   public UploadDialogWidgetV2(
-    GlobalApplicationStateImpl globalApplicationState,
```
nickgros, line 30: This erroneous injection registered the uploader's drag & drop handler on a new instance of GlobalApplicationStateImpl instead of the singleton.

jay-hodgson (approved):

PR #5597, opened 2024-12-03
PORTALS-3333

jay-hodgson:
Prebuilt PDF.js version, with one change.  In src/main/webapp/pdf.js/web/viewer.html, added the following CSP

```
<meta
      content="frame-ancestors 'self' https://*.synapse.org localhost;"
    />
```


jay-hodgson:
Sanity check - locally able to render PDFs using the updated PDF.js:
Screenshot 2024-12-03 at 11 51 00 AM

nickgros (approved):

PR #5604, opened 2024-12-10
SWC-7203: change to a project search query object

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -288,6 +298,26 @@ public ProfileViewImpl(
...
+    JSONObjectAdapter adapter = jsonObjectAdapter.createNew();
+    try {
+      SearchQuery query = SearchQueryUtils.getDefaultSearchQuery();
+      KeyValue projectsOnly = new KeyValue();
+      projectsOnly.setKey("node_type");
+      projectsOnly.setValue("project");
+      query.getBooleanQuery().add(projectsOnly);
+      query.setQueryTerm(
+        Collections.singletonList(projectSearchTextBox.getValue())
+      );
```
nickgros: should we split up the text into multiple query terms? I am indifferent and prefer whatever yields "better" results, but if we don't, otherwise you get this sequence where search doesn't "feel" idempotent:

1. Search from the project dashboard "my search terms"
2. Search place opens with query "my search terms" + "project" filter
3. Click "Search" again, and the results populate with the query ["my", "search", "terms"] + "project filter".

nickgros (approved):

nickgros (approved):

PR #5611, opened 2025-01-08
SWC-7067

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/jsinterop/CardConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+import jsinterop.base.Js;
+import jsinterop.base.JsPropertyMap;
+
+@JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
+public class CardConfiguration {
+
+  String type;
+  double secondaryLabelLimit;
```
nickgros, line 15:
```suggestion
  long secondaryLabelLimit;
```

nickgros (approved):
Small primitive type change but overall looks good

src/main/java/org/sagebionetworks/web/client/jsinterop/CardConfiguration.java
jay-hodgson: Did something change, I thought long values were disallowed?  Or perhaps I'm misunderstanding.
From https://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html

> The Java long type cannot be represented in JavaScript as a numeric type, so GWT emulates it using an opaque data structure. This means that JSNI methods cannot process a long as a numeric type. The compiler therefore disallows, by default, directly accessing a long from JSNI: JSNI methods cannot have long as a parameter type or a return type, and they cannot access a long using a [JSNI reference](https://www.gwtproject.org/doc/latest/DevGuideCodingBasics.html#DevGuideJavaScriptNativeInterface).

nickgros: This is JsInterop, not JSNI. However, it seems like `double` is the [least problematic primitive type for JsInterop](https://stackoverflow.com/a/62650203). In JsInterop, `int` and `long` do work; we're using those for properties in other JsInterop types. I cannot find an authoritative document that describes exactly what Java primitives and classes JsInterop supports.

Keeping double is probably fine!

jay-hodgson: Right!  I'm still thinking in JSNI.  I guess I'll leave it, for the small potential performance gain

PR #5612, opened 2025-01-09
SWC-4693

nickgros:
- Update UploadDialogWidgetV2 to configure drag-and-drop on render, and to completely re-render when the modal is closed.

EntityActionControllerImpl
- Configure the uploader when EntityActionControllerImpl is configured to prepare drag-and-drop
- Use a single instance of UploadDialogWidgetV2 to ensure the same component instance used to configure drag-and-drop is used when the modal is opened.

- Update Drag-and-Drop initialization code to use JsInterop instead of JSNI
- Remove feature flag checks for UploadDialogWidgetV2
- Delete UploadDialogWidget (v1)


src/main/java/org/sagebionetworks/web/client/FeatureFlagKey.java
```diff
@@ -40,9 +40,6 @@ public enum FeatureFlagKey {
   // If enabled, sharing settings will appear in a dialog immediately after uploading one or more files.
   SHOW_SHARING_SETTINGS_AFTER_UPLOAD("SHOW_SHARING_SETTINGS_AFTER_UPLOAD"),
 
-  // If enabled, uses the v2 uploader (react implementation) for file entity uploads.
-  UPLOADER_V2("UPLOADER_V2"),
-
```
nickgros, line 45: This has been enabled for weeks, let's remove it!

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -483,61 +487,61 @@ public void initializeToastContainer() {
   public void initializeDropZone() {
```
nickgros, line 487: Refactored to JsInterop for easier debugging

src/main/java/org/sagebionetworks/web/client/GlobalApplicationStateImpl.java
```diff
@@ -483,61 +487,61 @@ public void initializeToastContainer() {
...
+          globalAppState.onDrop(((DragEvent) e).dataTransfer.files);
+        }
+      };
+
+      dropZone.addEventListener("dragenter", allowDrag);
+      dropZone.addEventListener("dragover", allowDrag);
+
+      dropZone.addEventListener("drop", handleDrop);
+      dropZone.addEventListener("dragend", e -> hideDropZone.run());
+      dropZone.addEventListener("dragleave", e -> hideDropZone.run());
```
nickgros, line 535: Add dragend/dragleave handlers to fix bug where if you dragged into the window, then dragged out of the window, the dropzone wouldn't go away.

src/main/java/org/sagebionetworks/web/client/jsinterop/ReactComponentProps.java
```diff
@@ -17,6 +18,10 @@ public interface CallbackRef {
     void run(Element element);
   }
 
+  @JsNullable
+  public String key;
+
```
nickgros, line 23: All React components can take a `key` prop

src/main/java/org/sagebionetworks/web/client/jsinterop/EntityUploadModalProps.java
```diff
@@ -20,20 +21,26 @@ public interface Callback {
 
   public Callback onClose;
 
+  @JsNullable
+  public Callback onUploadReady;
```
nickgros, line 25: Use new callback prop

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -545,16 +544,12 @@ private EvaluationSubmitter getEvaluationSubmitter() {
...
-    view.setUploadDialogWidget(uploadDialogWidget.asWidget());
-    return uploadDialogWidget;
-  }
-
-  private UploadDialogWidgetV2 getNewUploadDialogWidgetV2() {
-    UploadDialogWidgetV2 uploadDialogWidgetV2 =
-      ginInjector.getUploadDialogWidgetV2();
-    view.setUploadDialogWidget(uploadDialogWidgetV2.asWidget());
+  private UploadDialogWidgetV2 getUploadDialogWidget() {
+    // Note: there must only be one UploadDialogWidget
```
nickgros:
```suggestion
```

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -545,16 +544,11 @@ private EvaluationSubmitter getEvaluationSubmitter() {
...
-
-  private UploadDialogWidgetV2 getNewUploadDialogWidgetV2() {
-    UploadDialogWidgetV2 uploadDialogWidgetV2 =
-      ginInjector.getUploadDialogWidgetV2();
-    view.setUploadDialogWidget(uploadDialogWidgetV2.asWidget());
+  private UploadDialogWidgetV2 getUploadDialogWidget() {
+    if (uploadDialogWidgetV2 == null) {
+      uploadDialogWidgetV2 = ginInjector.getUploadDialogWidget();
+      view.setUploadDialogWidget(uploadDialogWidgetV2.asWidget());
+    }
```
nickgros, line 551: Use a single instance of the widget to prevent bugs where the widget instance that was configured for drag-and-drop was not the same instance of the widget that is shown to the user. I don't think the present code would have this bug, but I noticed it in an intermediate iteration of this feature and it would otherwise be very easy to reintroduce in a later refactor

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1340,28 +1346,31 @@ private void configureTableCommands() {
...
-      actionMenu.setActionVisible(
-        Action.UPLOAD_FILE,
-        permissions.getCanCertifiedUserEdit()
-      );
-      actionMenu.setActionListener(Action.UPLOAD_FILE, this);
-    } else {
-      actionMenu.setActionVisible(Action.UPLOAD_FILE, false);
-    }
+    actionMenu.setActionVisible(Action.UPLOAD_FILE, canUploadFileToContainer());
+    actionMenu.setActionListener(Action.UPLOAD_FILE, this);
```
nickgros, line 1373: Refactor to share logic for these checks

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1830,6 +1839,7 @@ public void onAction(Action action, ReactMouseEvent event) {
         onSubmit();
         break;
       case UPLOAD_NEW_FILE:
+      case UPLOAD_FILE:
         onUploadFile();
```
nickgros, line 1843: These were different at one point, but today they both open the same widget with identical arguments.

src/main/java/org/sagebionetworks/web/client/widget/entity/download/UploadDialogWidgetV2.java
```diff
@@ -61,19 +69,41 @@ public void show() {
...
 
+  private void handleDrop(FileList fileList) {
+    if (this.ref != null && this.ref.current != null) {
+      this.ref.current.handleUploads(fileList);
+      // Show the uploader
+      this.show();
+    } else {
+      DomGlobal.console.error(
+        "EntityUploadHandle ref is null, aborting handleDrop"
+      );
```
nickgros, line 80: This should never happen, and user can recover by just opening the uploader, so I decided not to spend time on more sophisticated error handling.

src/main/java/org/sagebionetworks/web/client/widget/entity/download/UploadDialogWidgetV2.java
```diff
@@ -61,19 +69,41 @@ public void show() {
...
+    }
+  }
+
+  private void onUploadReady() {
+    globalApplicationState.setDropZoneHandler(this::handleDrop);
+  }
```
nickgros, line 86: Defer setting up drag-and-drop until the component tells us it's ready via callback

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
jay-hodgson: We should do some extra testing with component initialization with this change.  After uploading a file (or set of files), if you open the uploader to upload another set of files is the component state properly re-initialized?

jay-hodgson: (assume so with the call to configure())

src/main/java/org/sagebionetworks/web/client/widget/entity/download/UploadDialogWidgetV2.java
```diff
@@ -61,19 +69,41 @@ public void show() {
...
+
   private void onClose() {
     eventBus.fireEvent(new EntityUpdatedEvent(entityId));
+    // Re-render the component with a new key to reset the state so it no longer shows the previous set of uploads
+    keyCounter++;
```
jay-hodgson, line 91: Ah, there's the re-init.  Ignore my previous comment :D

jay-hodgson (approved):

PR #5614, opened 2025-01-09
SWC-6754 - Do not cache version check

nickgros:
SWC-6754

- Also add log to report instances where the servlet is not configured with a repo endpoint

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/StackEndpoints.java
+++ b/src/main/java/org/sagebionetworks/web/server/StackEndpoints.java
@@ -98,6 +98,15 @@ public class StackEndpoints {
         // None of these hosts match and no configuration was loaded from settings, fall back to prod
         endpointPrefix = "https://repo-prod.prod.sagebase.org";
       }
+
+      logger.warning(
+        "No configuration found for " +
+        REPO_ENDPOINT_KEY +
+        ", so servlet request will defer to the host header. Host: " +
+        host +
+        ", endpointPrefix: " +
+        endpointPrefix
+      );
     }
 
     return endpointPrefix;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/server/StackEndpoints.java
```diff
@@ -98,6 +98,15 @@ private static String getEndpointPrefix(String host) {
...
       }
+
+      logger.warning(
+        "No configuration found for " +
+        REPO_ENDPOINT_KEY +
+        ", so servlet request will defer to the host header. Host: " +
+        host +
+        ", endpointPrefix: " +
+        endpointPrefix
+      );
```
nickgros, line 109: IMO the main problem is that our servlet may use the request header to pick a backend endpoint instead of a configuration parameter, if the configuration parameter is not present. The servlet may cache a response that depends on a header that could vary (or be spoofed!) by different users.

WDYT about opening a follow-up ticket to make the `org.sagebionetworks.repositoryservice.endpoint` (REPO_ENDPOINT_KEY) required?

src/main/java/org/sagebionetworks/web/server/servlet/VersionsServlet.java
```diff
@@ -35,11 +31,14 @@ public class VersionsServlet extends HttpServlet {
       VersionsServlet.perThreadRequest.get()
     );
 
-  private final Supplier<SynapseVersionInfo> synapseVersionCache =
-    Suppliers.memoizeWithExpiration(versionSupplier(), 5, TimeUnit.MINUTES);
```
jay-hodgson, line 39: I think this was added because John identified the web client as performing a DOS attack with version checks.  Would expiring this cache every 10 seconds be sufficient?

src/main/java/org/sagebionetworks/web/server/StackEndpoints.java
jay-hodgson: Follow-up ticket to make this mandatory seems reasonable

nickgros: I no longer think that this approach makes sense, since staging and prod are configured the same way. I will update the PR to use the origin in the cache key (see SWC-6389 for historical context)

nickgros:
Closing in favor of #5634

PR #5618, opened 2025-01-24
SWC-7165 - Widen Show more results button on project mobile page

afwillia:
SWD-7165 Create a new scss class `btn-default-wide` to make the button span the width of the container on mobile. Apply this style to the LoadMoreWidgetContainerView

### DIFF
--- a/src/main/webapp/sass/_core.scss
+++ b/src/main/webapp/sass/_core.scss
@@ -422,6 +422,12 @@ end sticky footer
     color: $synapse-dark-text;
   }
 
+  @media (max-width: 500px) {
+    .btn-default-wide {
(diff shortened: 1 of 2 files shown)


src/main/webapp/sass/_core.scss
```diff
@@ -422,6 +422,12 @@ end sticky footer
     color: $synapse-dark-text;
   }
 
+  @media (max-width: 500px) {
```
nickgros: 768px is the common convention in the codebase for narrow screen widths
```suggestion
  @media (max-width: 768px) {
```

nickgros (approved):

PR #5621, opened 2025-01-24
Watch synapse-react-client & automatically copy resources

nickgros:


pom.xml
```diff
@@ -301,6 +301,29 @@
...
+        <!-- This configuration is used to watch the node_modules directory for changes and copy those to the static assets folder when they change -->
+        <!-- Then those assets are copied into the exploded WAR -->
+        <configuration>
+          <touchFile>target/classes/watcher_explode_war.txt</touchFile>
+          <watches>
+            <watch>
+              <directory>
+                ${project.basedir}/node_modules/synapse-react-client/dist/umd
+              </directory>
+              <!--              <directory>${project.basedir}/node_modules</directory>-->
```
nickgros:
```suggestion
```

pom.xml
```diff
@@ -301,8 +301,34 @@
...
+          <watches>
+            <watch>
+              <directory>
+                ${project.basedir}/node_modules/synapse-react-client/dist/umd
+              </directory>
+            </watch>
+            <watch>
+              <directory>${project.basedir}/src/main/webapp/css</directory>
+              <directory>${project.basedir}/src/main/webapp/sass</directory>
+            </watch>
```
nickgros, line 321: Watch source SCSS and CSS files for changes

pom.xml
```diff
@@ -301,8 +301,34 @@
...
+        <version>1.0.6</version>
+        <!-- This configuration is used to watch the node_modules directory for changes and copy those to the static assets folder when they change -->
+        <!-- Then those assets are copied into the exploded WAR -->
+        <configuration>
+          <touchFile>target/classes/watcher_explode_war.txt</touchFile>
+          <watches>
+            <watch>
+              <directory>
+                ${project.basedir}/node_modules/synapse-react-client/dist/umd
+              </directory>
```
nickgros, line 316: Watch synapse-react-client for changes. We can update this list as needed; I didn't want to start with watching the entire node_modules folder (not sure of the performance implications of watching 80k+ files)

pom.xml
```diff
@@ -316,7 +342,7 @@
           <execution>
             <id>make-assembly</id>
             <!-- this is used for inheritance merges -->
-            <phase>package</phase>
+            <phase>compile</phase>
```
nickgros, line 345: For some reason, my local instance of the GWT servlet would occasionally respond with esoteric failure messages on RPCs. The error in the RPC response would read "Type name elision in RPC payloads is only supported if the RPC whitelist file is used." and the client could not deserialize it. I have been having this issue for ages, with both Tomcat & Jetty, regardless of if I started the code server with the Maven plugin or the IntelliJ GWT extension.

Today I discovered that this was because my code server seemed to create incorrect (stale?) .rpc files for DevMode after running either `mvn clean compile` or `mvn gwt:run`. Running `mvn clean package` somehow fixed this. I noticed that the compiled .rpc file names generated after running `mvn package` differed from those generated by `mvn compile`, so I am assuming this assembly step accounts for the difference.

My working theory is that this assembly step is related and maybe updates some stale cache layer? I would appreciate more insight into this step if more context is available. I am hoping that moving the RPC assembly to the `compile` step will prevent this from happening in general.

jay-hodgson: I've also encountered stale rpc files that breaks GWT RPC (with the unable to deserialize client error).  I wish I had more context to give, but I don't see any issues moving this to the compile step

jay-hodgson (approved):

jay-hodgson:
Those force pushes are a bit scary (are you permanently deleting the correct commits?).  I assume you know what you're doing, since I don't remember losing any upstream work (so my assumption is that you are deleting your own local commits).  I feel it's our responsibility to raise the cautionary flag for this case, whether it's a real problem or not.

nickgros:
@jay-hodgson I'm only force-pushing over commits that are in my feature branches' history only, so shared history is untouched. Before my code has been reviewed, I'll often overwrite commits if I don't think that the sequence or commit messages provide useful context for anyone that might review the commit log (I especially do a lot of `git commit -m "work in progress"`). The habit also makes it easier to cherry-pick or split large PRs if I feel they have included too many separable changes for one review.

PR #5639, opened 2025-02-19
SWC-7188: SWC TablesTabViewImpl updates on mobile

kianamcc:
Jira tickets this PR covers: 1) https://sagebionetworks.jira.com/browse/SWC-7188 2) https://sagebionetworks.jira.com/browse/SWC-7190

Screenshot 2025-02-17 at 4 03 36 PM
Screenshot 2025-02-17 at 4 03 22 PM
Screenshot 2025-02-17 at 4 03 17 PM
Screenshot 2025-02-17 at 4 03 12 PM
Screenshot 2025-02-18 at 4 15 38 PM
Screenshot 2025-02-18 at 4 15 01 PM

### DIFF
--- a/src/main/resources/org/sagebionetworks/web/client/widget/discussion/ForumWidgetViewImpl.ui.xml
+++ b/src/main/resources/org/sagebionetworks/web/client/widget/discussion/ForumWidgetViewImpl.ui.xml
@@ -7,9 +7,7 @@
   xmlns:bh="urn:import:org.gwtbootstrap3.client.ui.html"
 >
   <bh:Div>
-    <bh:Div
-      addStyleNames="flexcontainer-row-unless-xs full-width-xs-only flexcontainer-xs-gap margin-bottom-20"
-    >
+    <bh:Div addStyleNames="flexcontainer-row-unless-xs margin-bottom-20">
       <bh:Div
         ui:field="headingFlexContainer"
         addStyleNames="flexcontainer-column flexcontainer-column-fill-width"
@@ -27,8 +25,6 @@
           Show All Threads
         </b:Button>
       </bh:Div>
-      <!--         addStyleNames="flexcontainer-column flexcontainer-align-items-flex-end min-width-70"
- -->
       <bh:Div
         ui:field="subscribersFlexContainer"
         addStyleNames="flexcontainer-column flexcontainer-xs-align-end"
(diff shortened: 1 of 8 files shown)


src/main/resources/org/sagebionetworks/web/client/widget/discussion/ForumWidgetViewImpl.ui.xml
```diff
@@ -27,37 +27,37 @@
...
         marginRight="10"
       >
         <b:Button ui:field="newThreadButton" type="PRIMARY">
           New Thread
         </b:Button>
       </bh:Div>
-
       <bh:Div
-        addStyleNames="flexcontainer-column flexcontainer-align-items-flex-end min-width-70"
+        addStyleNames="flexcontainer-column flexcontainer-xs-align-end full-width-button full-width-xs-only"
```
nickgros, line 60: Why add this `full-width-button` class? Should the menu component buttons always be full width when `xs`, or does it make sense to only apply this to this element?

src/main/webapp/sass/_core.scss
```diff
@@ -930,6 +930,17 @@ input[type='text'][readonly].border-none,
   vertical-align: bottom;
 }
 
+@media (max-width: 768px) {
+  .full-width-xs-only {
+    min-width: 100%;
+  }
+
+  .full-width-button .MuiButton-root,
+  .full-width-button .MuiBox-root {
```
nickgros, line 939: Not sure if we need the class, but if we do, suggest rename to `full-width-button-xs-only`

src/main/resources/org/sagebionetworks/web/client/widget/discussion/ForumWidgetViewImpl.ui.xml
kianamcc: The full-width-button class helps balance out this breakpoint:

Screenshot 2025-02-19 at 12 52 38 PM

Thought the Discussion Tools button should be full width here.

However, out of all of these devices, only the iPad mini and it's specific breakpoint has this imbalance issue:
Screenshot 2025-02-19 at 12 54 16 PM

Should I just remove this style because it seems to only be for a pretty specific breakpoint?

nickgros: Seems like the issue is that the MUI breakpoints and the Bootstrap breakpoints don't align--fixing that might be a good idea, but out of scope for this ticket, so I'm ok with this for now.

nickgros (approved):

PR #5654, opened 2025-03-12
SWC-7258

nickgros:
- Collect custom analytic events for entity search

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
@@ -509,10 +509,23 @@ public class SearchPresenter
   private <T extends SearchQueryEventData> T addSearchToAnalyticsEventData(
     T eventData
   ) {
-    eventData.query_term = join(currentSearch.getQueryTerm(), " ");
     eventData.search_context = SearchContext.synapse_entity.toString();
 
     JSONObjectAdapter adapter = this.jsonObjectAdapter.createNew();
+    try {
+      JSONArrayAdapter queryTermJSON = currentSearch
+        .writeToJSONObject(adapter)
+        .getJSONArray("queryTerm");
+      if (queryTermJSON != null) {
+        eventData.query_term = queryTermJSON.toJSONString();
+      }
+    } catch (JSONObjectAdapterException e) {
+      jsniUtils.consoleError(
+        "Error serializing queryTerm. It will be omitted from the analytics event."
+      );
+      jsniUtils.consoleError(e);
+    }
+
     if (
       currentSearch.getBooleanQuery() != null &&
       !currentSearch.getBooleanQuery().isEmpty()
--- a/src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
+++ b/src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
@@ -1,6 +1,7 @@
 package org.sagebionetworks.web.client.presenter;
 
 import com.google.gwt.activity.shared.AbstractActivity;
+import com.google.gwt.event.dom.client.ClickHandler;
 import com.google.gwt.event.shared.EventBus;
 import com.google.gwt.http.client.URL;
 import com.google.gwt.place.shared.Place;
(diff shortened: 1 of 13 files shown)


src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
```diff
@@ -414,4 +449,141 @@ private static String join(List<String> list, String delimiter) {
...
+   * Get event data for an individual search result
+   * @param hit
+   * @return
+   */
+  private SearchResultEventData getSearchResultEventData(Hit hit) {
+    SearchResultEventData resultEventData =
+      addSearchPageResultToAnalyticsEventData(new SearchResultEventData());
+    resultEventData.page_index =
+      Double.valueOf(1 + getIndexOfPageContainingHit(hit));
+    resultEventData.rank = Double.valueOf(1 + getIndexOfHit(hit));
```
jay-hodgson, line 491: There are opportunities to optimize, but this works.  Curious about the performance after clicking a handful of page results of data (say 15).
getSearchResultEventData() is called for every hit returned in the response.  Each call to getSearchResultEventData() calls getIndexOfPageContainingHit() , which goes through all search results that the client has retrieved to find the index of the page.  It also calls getIndexOfHit(), which goes through all search results that the client has retrieved to find the cumulativeIndex.
I think we could hold on to the currentPageIndex, and use that for the new SearchResultEventData (as well as the rank if we know the page size).

src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
```diff
@@ -414,4 +449,141 @@ private static String join(List<String> list, String delimiter) {
...
+  }
+
+  /**
+   * Adds data related to the latest search request that is not dependent on page
+   * @param eventData
+   */
+  private <T extends SearchQueryEventData> T addSearchToAnalyticsEventData(
+    T eventData
+  ) {
+    eventData.query_term = join(currentSearch.getQueryTerm(), " ");
```
jay-hodgson: Would this be more informative if we used a different separator to join the query terms?

src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
jay-hodgson: It's very possible that I'm missing something, so please correct me if I'm wrong!

src/test/java/org/sagebionetworks/web/unitclient/presenter/SearchPresenterTest.java
```diff
@@ -256,4 +557,14 @@ public void testEncodedQueryTerm() throws Exception {
     verify(mockJsClient)
       .getSearchResults(eq(expectedQuery), any(AsyncCallback.class));
   }
+
+  @Test
+  public void testGetSearchResultClickedHandler() throws Exception {
```
jay-hodgson: ?

jay-hodgson: You tested the click handler above, should this test be deleted?

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
nickgros: Probably so, as we could capture cases where uses use things like quotation marks, or perhaps future changes are made so we do not split on the space token. We could use some other separator but we would have to make sure to escape it when we serialize it. It's a bit frustrating that GA4 only allows string values for custom events. Maybe we just serialize the JSON array instead? We wouldn't be able to parse it in GA4 but we could in Snowflake or other analytical engine

jay-hodgson: It will probably depend on the viewer.  
But stringifying the JSON array seems like the best thing we can do to ensure they have all the necessary info.

src/main/java/org/sagebionetworks/web/client/presenter/SearchPresenter.java
nickgros: Great call out. I agree that this data should just be stored and referenced instead of iterating through the full result set $n^2$ times. When a page is fetched, each result is used to create a(n opaque) widget and passed to the view; we don't have a data structure right now that associates the Hit with the page it was on--we use the LoadMoreWidgetContainer which is extremely simple (to a fault). Creating something like that in the SearchPresenter was substantial refactoring that I didn't want to do (at some point, I'd rather just reimplement in React where mapping data to UI would be very straightforward).

IMO we can revisit if this causes a performance issue. I am not sure how many users ever get to page 15.

Slightly related, I found that [gtag supposedly does event batching](https://support.google.com/analytics/answer/9322688?hl=en#zippy=%2Crealtime-report%2Cdebugview-report) (though I didn't verify) so this shouldn't result in e.g. 20 API calls for a page of 20 items.

src/test/java/org/sagebionetworks/web/unitclient/presenter/SearchPresenterTest.java
nickgros: Yes it should, thanks

PR #5657, opened 2025-03-19
Release 536: update to SRC 3.4.7

jay-hodgson:


jay-hodgson:
Merging to kick off the build

PR #5662, opened 2025-04-02
SWC-7191: SWC SingleDiscussionThreadWidget updates on mobile

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7191

Screenshot 2025-04-02 at 10 44 14 AM

Screenshot 2025-04-02 at 10 45 20 AM


src/main/java/org/sagebionetworks/web/client/jsinterop/DiscussionThreadProps.java
```diff
@@ -0,0 +1,20 @@
...
+  public int limit;
+
+  @JsOverlay
+  public static DiscussionThreadProps create(String threadId, int limit) {
+    DiscussionThreadProps props = new DiscussionThreadProps();
+    props.threadId = threadId;
+    props.limit = limit;
+    return props;
+  }
+}
```
kianamcc, line 20: Prop class that represents the props from DiscussionThread in SRC. 'create' helper function is used to instantiate an instance of DiscussionThreadProps.

src/main/java/org/sagebionetworks/web/client/jsinterop/SRC.java
```diff
@@ -21,6 +21,7 @@ public class SRC {
   @JsType(isNative = true)
   public static class SynapseComponents {
 
+    public static ReactComponentType<DiscussionThreadProps> DiscussionThread;
```
kianamcc, line 24: Reference to DiscussionThread component in SRC which expects props that match the DiscussionThreadProps type.

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadViewImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import org.sagebionetworks.web.client.jsinterop.DiscussionThreadProps;
+import org.sagebionetworks.web.client.jsinterop.React;
+import org.sagebionetworks.web.client.jsinterop.ReactElement;
+import org.sagebionetworks.web.client.jsinterop.SRC;
+import org.sagebionetworks.web.client.widget.ReactComponent;
+
+public class DiscussionThreadViewImpl extends ReactComponent {
+
+  public DiscussionThreadViewImpl(
+    SynapseReactClientFullContextPropsProvider contextPropsProvider,
```
kianamcc, line 13: Initialize and render the DiscussionThread component, wrapped in SynapseContext. getJsInteropContextProps method most likely provides context data to ensure that the component has the right info to interact with Synapse.

src/main/java/org/sagebionetworks/web/client/widget/discussion/ForumWidgetViewImpl.java
```diff
@@ -31,12 +32,6 @@ public class ForumWidgetViewImpl implements ForumWidgetView {
   @UiField
   Button showAllThreadsButton;
 
-  @UiField
-  ButtonGroup repliesSortButtonGroup;
```
kianamcc, line 35: these fields were removed in the .ui.xml after replacing them with the SRC DiscussionThread.

kianamcc:
I think we would have to make a SRC release first or this could break things.

Since this is a React component wrapped in the Synapse context, it seems that we have all the necessary information, such as authentication, to use this component and perform actions like deleting, editing, replying, and such.

src/main/java/org/sagebionetworks/web/client/widget/discussion/DiscussionThreadViewImpl.java
```diff
@@ -0,0 +1,26 @@
+package org.sagebionetworks.web.client.widget.discussion;
+
+import com.google.gwt.core.client.GWT;
```
nickgros: can you make sure unused imports are removed?

src/main/java/org/sagebionetworks/web/client/widget/discussion/ForumWidget.java
```diff
@@ -374,9 +373,10 @@ public void invoke() {
             }
           );
 
-          view.setSingleThreadUIVisible(true);
+          view.configureDiscussionThread(threadId, 20);
+          view.setDiscussionThreadVisible(true);
```
nickgros: Can we pull this out of the `onSuccess` handler, since the component fetches the thread bundle internally? Do we still need this call to `jsClient.getThread` (seems like maybe we do since some of the other buttons depend on it)?

kianamcc: Correct me if I’m wrong but I believe jsClient.getThread might be needed for configuring things such as the action menu.

nickgros (approved):

PR #5683, opened 2025-04-24
SWC-7314

nickgros:
Fix dependency breaking build. Use same patch as in synapse-web-monorepo


patches/@jsdevtools__ono@7.1.3.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/esm/index.js b/esm/index.js
```
nickgros, line 1: This is the same patch for `@jsdevtools/ono` as we built for synapse-react-client (can be seen in `synapse-web-monorepo/patches`)

Other folks have experienced the same issue: https://github.com/JS-DevTools/ono/issues/19

jay-hodgson (approved):

jay-hodgson:
Object.assign(undefined, module.exports) is problematic, thanks for patching

PR #5693, opened 2025-05-09
Upgrade synapse-react-client & all outdated deps

nickgros:
Upgrade synapse-react-client and `pnpm update`

Also delete a CSS import that should have been removed in #5691 

A change to synapse-react-client breaks plots in dev, but not production mode (caused by some difference between esbuild and rollup). Opened https://sagebionetworks.jira.com/browse/SWC-7329 to track.


kianamcc (approved):
@nickgros will investigate and fix plots in dev on Monday.

PR #5704, opened 2025-05-23
SWC-7230

nickgros:
- Add docker-compose scripts for development & e2e
- Do not try to format all files, list file extensions that can be formatted.

### DIFF
--- a/compose-e2e.yml
+++ b/compose-e2e.yml
@@ -1,0 +1,14 @@
+# Compose file for SynapseWebClient E2E tests
+services:
+  swc-e2e-tomcat:
+    image: tomcat:9
+    ports:
+      - '8888:8080' # Adjust as needed
+    volumes:
+      - ./target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war
+      - ./e2e_workflow/settings.xml:/root/.m2/settings.xml:ro
+
+  sage-account-web:
+    image: ghcr.io/sage-bionetworks/sage-account-web:latest-development
+    ports:
+      - 3000:80
--- a/compose-link.yml
+++ b/compose-link.yml
@@ -1,0 +1,20 @@
+# Supplemental Docker Compose file to support linking synapse-web-monorepo packages with pnpm link in SynapseWebClient
+#
+# When using `pnpm link` to load one or more local instances of a package, the Docker container(s) that build SWC must
+#  be able to access those directories. This file adds synapse-web-monorepo as a volume to the gwt-codeserver and
+#  vite-codeserver services so that linked packages in the synapse-web-monorepo project can be accessed.
+# Note: this assumes that the synapse-web-monorepo project is located in the parent directory of the SynapseWebClient,
+#  i.e. your directory structure looks like this:
+#  .
+#  ├── synapse-web-monorepo
+#  └── SynapseWebClient
+#
+# This file can be used like `docker-compose -f docker-compose.yml -f compose-link.yml up`
+# See https://docs.docker.com/reference/cli/docker/compose/#specifying-multiple-compose-files
+services:
+  gwt-codeserver:
+    volumes:
+      - ../synapse-web-monorepo:/usr/src/synapse-web-monorepo
+  vite-codeserver:
+    volumes:
+      - ../synapse-web-monorepo:/synapse-web-monorepo
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,88 @@
+# Compose file for SynapseWebClient development server
+services:
+  # The GWT Codeserver compiles Java assets and serves an exploded WAR to be served by Tomcat
+  gwt-codeserver:
+    # Use corretto 17 as in production. We use the Amazon Linux 2023 variant because it has a recent enough version of glibc required to run Node
+    image: maven:3.9.9-amazoncorretto-17-al2023
+    environment:
+      - ORG_SAGEBIONETWORKS_REPOSITORYSERVICE_ENDPOINT=${ORG_SAGEBIONETWORKS_REPOSITORYSERVICE_ENDPOINT}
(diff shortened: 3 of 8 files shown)


compose-link.yml
```diff
@@ -0,0 +1,20 @@
...
+#  be able to access those directories. This file adds synapse-web-monorepo as a volume to the gwt-codeserver and
+#  vite-codeserver services so that linked packages in the synapse-web-monorepo project can be accessed.
+# Note: this assumes that the synapse-web-monorepo project is located in the parent directory of the SynapseWebClient,
+#  i.e. your directory structure looks like this:
+#  .
+#  ├── synapse-web-monorepo
+#  └── SynapseWebClient
+#
+# This file can be used like `docker-compose -f docker-compose.yml -f compose-link.yml up`
+# See https://docs.docker.com/reference/cli/docker/compose/#specifying-multiple-compose-files
```
nickgros, line 13: This is hacky, but I couldn't think of a better way to set this up. This PR doesn't remove/change the current process, so it's not required to use this to use the linked SRC, but it's probably the only way we can make it work in a docker container without merging SWC into the monorepo.

package.json
```diff
@@ -69,17 +69,20 @@
...
   "lint-staged": {
-    "*.{*}": "prettier --write"
+    "*.{java,tsx,ts,scss,xml,yml,json}": "prettier --write"
   },
   "scripts": {
     "build": "mvn -B package --file pom.xml",
+    "start:production": "docker compose --env-file .env.production up",
+    "start:staging": "docker compose --env-file .env.staging up",
+    "start:dev": "docker compose --env-file .env.development up",
     "dev": "pnpm concurrently -k -n \"CODESERVER,WATCHER,TOMCAT,VITE\" -c \"auto,auto,auto,auto\" \"pnpm dev:codeserver\" \"pnpm dev:watcher\" \"pnpm dev:tomcat\" \"pnpm dev:vite\"",
```
nickgros, line 79: Don't change `dev`, it still works the same

package.json
```diff
@@ -69,17 +69,20 @@
...
   },
   "lint-staged": {
-    "*.{*}": "prettier --write"
+    "*.{java,tsx,ts,scss,xml,yml,json}": "prettier --write"
   },
   "scripts": {
     "build": "mvn -B package --file pom.xml",
+    "start:production": "docker compose --env-file .env.production up",
+    "start:staging": "docker compose --env-file .env.staging up",
+    "start:dev": "docker compose --env-file .env.development up",
```
nickgros, line 78: Add scripts to start dev services on a stack of your choice.

package.json
```diff
@@ -69,17 +69,20 @@
...
     "dev": "pnpm concurrently -k -n \"CODESERVER,WATCHER,TOMCAT,VITE\" -c \"auto,auto,auto,auto\" \"pnpm dev:codeserver\" \"pnpm dev:watcher\" \"pnpm dev:tomcat\" \"pnpm dev:vite\"",
     "dev:codeserver": "mvn clean gwt:run-codeserver",
     "dev:watcher": "wait-on tcp:127.0.0.1:9876 && mvn fizzed-watcher:run",
     "dev:tomcat": "wait-on tcp:127.0.0.1:9876 && docker pull tomcat:9.0; docker run --name swc-dev --rm -p 8888:8080 -v \"/$(pwd)/target/portal-develop-SNAPSHOT/:/usr/local/tomcat/webapps/ROOT/\" -v \"/$HOME/.m2/settings.xml\":/root/.m2/settings.xml -e CATALINA_OPTS=\"-Dorg.sagebionetworks.web.client.dev.mode=true\" tomcat:9.0",
     "dev:vite": "vite dev",
-    "docker:start": "docker pull tomcat:9.0; docker run --name swc-tomcat -d --rm -p 8888:8080 -v \"/$(pwd)/target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war\" -v \"/$(pwd)/e2e_workflow/settings.xml\":/root/.m2/settings.xml tomcat:9.0",
-    "docker:stop": "docker stop swc-tomcat",
+    "docker:start": "docker compose -f compose-e2e.yml up -d",
+    "docker:stop": "docker compose -f compose-e2e.yml down",
```
nickgros, line 85: Migrate these to `docker compose`, where they will also start an instance of `SageAccountWeb`

compose-e2e.yml
```diff
@@ -0,0 +1,14 @@
...
+    ports:
+      - '8888:8080' # Adjust as needed
+    volumes:
+      - ./target/portal-develop-SNAPSHOT.war:/usr/local/tomcat/webapps/ROOT.war
+      - ./e2e_workflow/settings.xml:/root/.m2/settings.xml:ro
+
+  sage-account-web:
+    image: ghcr.io/sage-bionetworks/sage-account-web:latest-development
+    ports:
+      - 3000:80
```
nickgros, line 14: Migrated existing `docker-run` script and added another service to launch the accounts site

docker-compose.yml
```diff
@@ -0,0 +1,88 @@
...
+  # The sage-account-web service is a simple HTTP server that serves the SageAccountWeb static site as a convenience for local development
+  sage-account-web:
+    environment:
+      - SAGE_ACCOUNT_WEB_IMAGE_TAG=${SAGE_ACCOUNT_WEB_IMAGE_TAG}
+    # Parameterize the image using an environment variable
+    image: ghcr.io/sage-bionetworks/sage-account-web:${SAGE_ACCOUNT_WEB_IMAGE_TAG}
+    ports:
+      - '3000:80'
+
+volumes:
```
nickgros, line 70: Add a volume for each output/generated folder so the container user(s) don't interfere with the outer filesystem. We can't just copy everything at the start and work entirely in a virtual FS because we want the code server to recompile on changes

compose-link.yml
jay-hodgson: I don't see the benefits of merging SWC into the monorepo at this time, and can't think of a better solution.  So this works for me!

jay-hodgson (approved):

PR #5706, opened 2025-05-30
SWC-7345

nickgros:
Update styling on ManagedACTAccessRequirement and SelfSignAccessRequirement pages to remove blockquote, add some spacing for consistency across pages

Before:

image

After:

image


kianamcc (approved):
Looks better! Would it be nice to close up some of the spacing above the title, "Test AR"? There is some extra spacing for the bottom text between "...Synapse." and "If you...". I wonder if we should move the latter to a new line? Just some ideas/thoughts!

kianamcc (approved):

nickgros:
@kianamcc The space above the title is related to the wiki rendering, so it's kind of out of scope of this ticket. I fixed the mid-paragraph spacing issue though, nice catch!

PR #5716, opened 2025-06-13
SWC-7358: Add Cite As button in Synapse Entity Page

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7358?atlOrigin=eyJpIjoiMWJlNTg0OTJiNTFkNGI5ZmJmOGRkYmM4NGNiZDBmYjUiLCJwIjoiaiJ9

- Port synapse-react-client's EntityCitation and LinkWithIcon
- Include EntityCitation and LinkWithIcon in EntityPageTop
- Removing project help (LinkWithIcon replaces this)

Related PR (includes images): https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/1967


src/main/java/org/sagebionetworks/web/client/jsinterop/mui/ButtonProps.java
```diff
@@ -0,0 +1,46 @@
...
+  public boolean fullWidth;
+
+  @JsProperty
+  public boolean disabled;
+
+  @JsProperty
+  public String target;
+
+  @JsProperty
+  public String children;
```
nickgros, line 40: I think our props types already include children, so you may be able to get rid of this (but maybe yours is easier to use because it is easier to work with a `String` in Java).

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -615,6 +620,13 @@ public void updateEntityBundle(EntityBundle bundle, Long version) {
         dockerChanged(bundle);
       }
     }
+    String projectId = projectHeader != null ? projectHeader.getId() : null;
+    String entityId = entity != null ? entity.getId() : null;
+    Double versionNum = currentTargetVersionNumber != null
+      ? currentTargetVersionNumber.doubleValue()
+      : null;
+    entityCitation.configure(projectId, entityId, versionNum);
```
nickgros: Seems like this could be replaced with the utility you have below?
```suggestion
    configureEntityCitationForTab(bundle, currentTargetVersionNumber);
```

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -1024,6 +1053,7 @@ public void configureDockerTab() {
       dockerTab.configure(dockerEntityBundle, dockerAreaToken);
       dockerTab.asTab().setContentStale(false);
     }
+    entityCitation.configure(projectHeader.getId(), null, null);
```
nickgros:
```suggestion
    entityCitation.configure(projectHeader.getId(), dockerEntityBundle, null);
```

nickgros: docker repos are entities that can have DOIs

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1562,16 +1561,6 @@ private void configureEditProjectMetadataAction() {
     }
   }
 
-  private void configureProjectHelpAction() {
-    boolean isProjectMenu =
-      entityBundle.getEntity() instanceof Project && currentArea == null;
-    actionMenu.setActionVisible(Action.PROJECT_HELP, isProjectMenu);
```
nickgros, line 1568: Can you also remove the PROJECT_HELP enumerated value from `Action`?

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTopViewImpl.java
```diff
@@ -43,6 +50,8 @@ public interface Binder extends UiBinder<Widget, EntityPageTopViewImpl> {}
   @Inject
   public EntityPageTopViewImpl(Binder uiBinder) {
     initWidget(uiBinder.createAndBindUi(this));
+    button.setChildren("Project Support");
+    button.setStartIcon("helpChatBubble");
```
nickgros, line 54: Did this work well? Don't you need to use an IconSvg?

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
kianamcc: Do you mean something like this: `entityCitation.configure(projectHeader.getId(), dockerEntityBundle.getEntity().getId(), null);` instead of passing the entire entity bundle? The configure method expects props of type String (project ID), String (entity ID), and Number (version), so I assume I should extract the entity id from the bundle?

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTopViewImpl.java
kianamcc: It works for me! The setStartIcon method creates an IconSvg with that string (iconname).

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
kianamcc: I missed this, nice catch!

src/main/java/org/sagebionetworks/web/client/jsinterop/mui/ButtonProps.java
kianamcc: If it makes it easier I would like to keep it this way but if you'd like me to remove it, lmk!

nickgros: Is the reason because it's easier to pass text to a String type (vs creating an entire React node)? If so, that's fine. Just trying to understand/confirm

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
nickgros: My mistake, but really I think you can just use your utility

```
configureEntityCitationForTab(datasetsEntityBundle, null);
```

src/main/java/org/sagebionetworks/web/client/jsinterop/mui/Button.java
```diff
@@ -0,0 +1,66 @@
...
+    props.size = size;
+    this.render();
+  }
+
+  public void setStartIcon(String iconName) {
+    IconSvgProps iconProps = IconSvgProps.create(iconName, null);
+    props.startIcon =
+      React.createElement(SRC.SynapseComponents.IconSvg, iconProps);
+    this.render();
+  }
```
nickgros: Oh, I didn't realize you created the IconSvg here. I want this class to match the original interface as much as possible, but I like that you added this, as I think it is a really helpful utility to keep. What do you think of this refinement?

```suggestion
  /**
   * Convenience wrapper that uses synapse-react-client's IconSvg
   */
  public void setStartIcon(String iconSvgName) {
    IconSvgProps iconProps = IconSvgProps.create(iconSvgName, null);
    ReactNode icon = React.createElement(SRC.SynapseComponents.IconSvg, iconProps);
    this.setStartIcon(icon);
  }
  
  public void setStartIcon(ReactNode icon) {
    props.startIcon = icon
    this.render();
  }
```

src/main/java/org/sagebionetworks/web/client/widget/entity/EntityPageTop.java
```diff
@@ -1024,6 +1047,11 @@ public void configureDockerTab() {
       dockerTab.configure(dockerEntityBundle, dockerAreaToken);
       dockerTab.asTab().setContentStale(false);
     }
+    entityCitation.configure(
+      projectHeader.getId(),
+      dockerEntityBundle.getEntity().getId(),
+      null
+    );
```
nickgros: Does this work?
```suggestion
    configureEntityCitationForTab(dockerEntityBundle, null);
```

src/main/java/org/sagebionetworks/web/client/jsinterop/mui/Button.java
kianamcc: Makes sense! Though I don't see a ReactNode type.. can we use ReactElement?

src/main/java/org/sagebionetworks/web/client/jsinterop/mui/ButtonProps.java
kianamcc: Yes,, I think it would be easier to pass button labels directly as strings. However, it might not be ideal for complex nested children..

src/main/java/org/sagebionetworks/web/client/jsinterop/mui/Button.java
nickgros: My bad, `ReactElement` will work! `ReactNode` is only in TypeScript and wouldn't make sense to represent in Java

nickgros (approved):

PR #5723, opened 2025-06-24
SWC -7055 

nickgros:
- Fix global editing state bug in React modal components
- Reset DOI modal state after closing


src/main/java/org/sagebionetworks/web/client/jsinterop/CreateOrUpdateDoiModalProps.java
```diff
@@ -39,6 +39,10 @@ public static CreateOrUpdateDoiModalProps create(
     props.objectType = objectType.toString();
     props.objectId = objectId;
     props.defaultVersionNumber = defaultVersionNumber;
+
+    // Reset the modal state when it is closed
+    props.key = Boolean.toString(open);
```
nickgros, line 44: Without this, changes to the form fields in the DOI editor will persist after closing and re-opening, which is unexpected after telling the user the changes would be lost.

jay-hodgson (approved):

PR #5734, opened 2025-07-07
SWC-7374

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/EntityMetadataViewImpl.java
```diff
@@ -97,6 +114,16 @@ public void setDetailedMetadataVisible(boolean visible) {
...
+    SynapsePortalBannersProps props = SynapsePortalBannersProps.create(
+      entityId
+    );
+    ReactElement component = React.createElementWithSynapseContext(
+      SRC.SynapseComponents.SynapsePortalBanners,
+      props,
+      propsProvider.getJsInteropContextProps()
+    );
+
+    portalBannersReactContainer.render(component);
```
nickgros: May be cleaner to  create a `PortalBannersWidget` and move all this logic there. Here just call `portalBannersWidget.configure(entityId)`.

jay-hodgson: Refactored to PortalBannersWidget (a view only widget)

nickgros (approved):

PR #5735, opened 2025-07-09
SWC-7390 - Fix Vite dynamic imports, add dynamic CDN redirect

nickgros:
Fixes SWC-7390

- Add servlet endpoint to conditionally redirect to CDN
- Update Vite base path so dynamically loaded modules can be resolved through CDN servlet

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/CdnRedirectorServlet.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/CdnRedirectorServlet.java
@@ -75,9 +75,9 @@ public class CdnRedirectorServlet extends HttpServlet {
           pathToAsset
         );
       logger.warning(
-        "Not redirecting to CDN, server name \"" +
+        "Not redirecting to CDN: server name \"" +
         request.getServerName() +
-        "\" does not match known CDN server names. This is expected in development."
+        "\" does not match known server names with CDNs. This is expected in development."
       );
     }
 
--- a/src/main/java/org/sagebionetworks/web/server/servlet/CdnRedirectorServlet.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/CdnRedirectorServlet.java
@@ -12,7 +12,6 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import org.sagebionetworks.web.shared.WebConstants;
 
 /**
  * Handles file handler uploads.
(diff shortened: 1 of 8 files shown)


vite.config.ts
```diff
@@ -5,6 +5,8 @@ import { nodePolyfills } from 'vite-plugin-node-polyfills'
  * Vite config to generate the ESM & CJS bundles for Synapse React Client.
  */
 const config = defineConfig({
+  // Assets will be served from the `/Portal/cdn` servlet, which will redirect requests to the CDN when appropriate
+  base: '/Portal/cdn/generated/vite/',
```
nickgros, line 9: Crucially, dynamically imported assets will now be imported using this base path, fixing 404s.

src/main/java/org/sagebionetworks/web/server/servlet/CdnRedirectorServlet.java
```diff
@@ -0,0 +1,96 @@
...
+        );
+      logger.warning(
+        "Not redirecting to CDN: server name \"" +
+        request.getServerName() +
+        "\" does not match known server names with CDNs. This is expected in development."
+      );
+    }
+
+    // Redirect with 302 Found
+    response.sendRedirect(redirectUrl.toString());
```
nickgros, line 83: Not sure if we need to do anything special with cache headers. [This document](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/http-3xx-status-codes.html) describes how CloudFront treats 302s like any other status, but I don't know how we have configured it today.

The assets served by Vite include a hash, so aggressive caching would be fine for those files, but this endpoint could be used to serve other assets that do not bust the cache.

src/main/java/org/sagebionetworks/web/server/servlet/CdnRedirectorServlet.java
```diff
@@ -0,0 +1,96 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Handles file handler uploads.
+ *
+ * @author jay
+ *
+ */
```
nickgros:
```suggestion
/**
 * Redirects a request for an asset to the correct path, referencing the CDN if one exists.
 */
```

src/main/java/org/sagebionetworks/web/server/servlet/CdnRedirectorServlet.java
```diff
@@ -0,0 +1,93 @@
...
+    // Redirect with 302 Found
+    response.sendRedirect(redirectUrl.toString());
+  }
+
+  private static String getOriginalScheme(HttpServletRequest request) {
+    String forwardedProtoHeader = request.getHeader("X-Forwarded-Proto");
+    if (forwardedProtoHeader != null) {
+      return forwardedProtoHeader;
+    }
+    return request.getScheme();
```
nickgros, line 91: Another mystery to me -- does our load balancer connect to the app servers using HTTPS? My local reverse proxy connected to my local app server using HTTP, so I had to use the `X-Forwarded-Proto` header to tell the servlet that the browser was using HTTPS to connect to the RP, even if the RP was not using it to connect to the app server.

I dunno if any of this will even matter in production, since we should be redirecting HTTP to HTTPS.

jay-hodgson (approved):

PR #5740, opened 2025-07-16
SWC-7004 - Global store for context props

nickgros:
Move context to global store so it can be passed/retrieved outside of dependency injection.

The primary motivation for this is so we can instantiate via GWT UiBinder widgets that encapsulate React components with access to the MUI theme context. This is not possible with Guice dependency injection unless we explicitly provide the context in every widget that uses one of these components in their ui.xml, which is too much boilerplate to bear.

Here is an example commit of the types of changes that this enables: [`280d5db` (#5741)](https://github.com/Sage-Bionetworks/SynapseWebClient/pull/5741/commits/280d5db54cbe06af5eafdef8dd37a9cb4fd0c0ac#diff-297fbdadc2e58f0ec7bfb7575e7c3a9e250e61708071060e3df777a14f1d47dfR82-R90)

### DIFF
--- a/js/globalContext.js
+++ b/js/globalContext.js
@@ -1,0 +1,26 @@
+import { atom, createStore, Provider, useAtomValue } from 'jotai'
+import { createElement } from 'react'
+import { SynapseContext } from 'synapse-react-client'
+
+const contextStore = createStore()
+const contextAtom = atom({})
+
+function _SynapseContextProviderFromStore({ children }) {
+  const context = useAtomValue(contextAtom)
+
+  return createElement(SynapseContext.FullContextProvider, context, children)
+}
+
+function SynapseContextProviderFromStore(props) {
+  return createElement(
+    Provider,
+    { store: contextStore },
+    createElement(_SynapseContextProviderFromStore, props),
+  )
+}
+
+window.ContextUtils = {
+  setGlobalContext: ctx => contextStore.set(contextAtom, ctx),
+  contextStore: contextStore,
+  SynapseContextProviderFromStore: SynapseContextProviderFromStore,
+}
--- a/js/globalContext.js
+++ b/js/globalContext.js
@@ -2,25 +2,22 @@ import { atom, createStore, Provider, useAtomValue } from 'jotai'
 import { createElement } from 'react'
 import { SynapseContext } from 'synapse-react-client'
 
+/* A store that will be used across all React elements in the app */
 const contextStore = createStore()
-const contextAtom = atom({})
 
-function _SynapseContextProviderFromStore({ children }) {
-  const context = useAtomValue(contextAtom)
+/* Atom that can store the props for SynapseContextProvider */
+const contextProviderPropsAtom = atom({})
+
+/* Wraps children in a FullContextProvider, reading the global context store to configure the context */
+function SynapseContextProviderFromStore({ children }) {
+  const context = useAtomValue(contextProviderPropsAtom, {
+    store: contextStore,
+  })
 
   return createElement(SynapseContext.FullContextProvider, context, children)
 }
 
-function SynapseContextProviderFromStore(props) {
-  return createElement(
-    Provider,
-    { store: contextStore },
-    createElement(_SynapseContextProviderFromStore, props),
-  )
-}
-
 window.ContextUtils = {
-  setGlobalContext: ctx => contextStore.set(contextAtom, ctx),
-  contextStore: contextStore,
+  setGlobalContext: ctx => contextStore.set(contextProviderPropsAtom, ctx),
   SynapseContextProviderFromStore: SynapseContextProviderFromStore,
 }
--- a/src/main/java/org/sagebionetworks/web/client/SessionDetector.java
+++ b/src/main/java/org/sagebionetworks/web/client/SessionDetector.java
@@ -16,21 +16,18 @@ public class SessionDetector {
   public static final String SESSION_MARKER = "SESSION_MARKER";
   public static final int FORCE_CHECK_EVERY_X_ITERATIONS = 7;
   private int loopCount = 0;
-  private SynapseReactClientFullContextPropsProvider propsProvider;
 
   @Inject
   public SessionDetector(
     AuthenticationController authController,
     GlobalApplicationState globalAppState,
     GWTWrapper gwt,
-    ClientCache clientCache,
-    SynapseReactClientFullContextPropsProvider propsProvider
+    ClientCache clientCache
   ) {
     this.clientCache = clientCache;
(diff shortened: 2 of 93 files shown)


js/globalContext.js
```diff
@@ -0,0 +1,23 @@
...
+function SynapseContextProviderFromStore({ children }) {
+  const context = useAtomValue(contextProviderPropsAtom, {
+    store: contextStore,
+  })
+
+  return createElement(SynapseContext.FullContextProvider, context, children)
+}
+
+window.ContextUtils = {
+  setGlobalContext: ctx => contextStore.set(contextProviderPropsAtom, ctx),
```
nickgros, line 21: SWC will call this function to populate the atom with the context value.

js/globalContext.js
```diff
@@ -0,0 +1,23 @@
...
+const contextProviderPropsAtom = atom({})
+
+/* Wraps children in a FullContextProvider, reading the global context store to configure the context */
+function SynapseContextProviderFromStore({ children }) {
+  const context = useAtomValue(contextProviderPropsAtom, {
+    store: contextStore,
+  })
+
+  return createElement(SynapseContext.FullContextProvider, context, children)
+}
```
nickgros, line 18: Super simple context wrapper that will get the context props from a Jotai atom and update when the atom value changes.

src/main/java/org/sagebionetworks/web/client/SessionDetector.java
```diff
@@ -38,6 +42,8 @@ public void start() {
     // SWC-4947: check for user change immediately (don't wait 10 seconds to discover the existing
     // session token).
     authController.checkForUserChange();
+    // SWC-7004: Update global store with current React context
+    propsProvider.synchronizeContextWithGlobalStore();
```
nickgros: It seems to suffice to synchronize the global atom when `SessionDetector.start` runs (which is necessarily after authentication has completed)--but please suggest a better place if you can think of one.

src/main/java/org/sagebionetworks/web/client/context/SynapseReactClientFullContextPropsProviderImpl.java
```diff
@@ -76,4 +77,9 @@ public FullContextProviderPropsJSNIObject getJsniContextProps() {
     props.setQueryClient(QueryClientJSNIObject.getQueryClientSingleton());
     return props;
   }
+
+  @Override
+  public void synchronizeContextWithGlobalStore() {
+    ContextUtils.setGlobalContext(this.getJsInteropContextProps());
```
nickgros: Call the `setGlobalContext` function that we set up in the globalContext.js file.

src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
```diff
@@ -23,34 +23,27 @@ > ReactElement<T, P> createElement(
...
+   * simplifies creating the wrapper.
+   *
+   * @param <P>
+   * @param componentType
+   * @return
    */
   @JsOverlay
   public static <
     T extends ReactComponentType<P>, P extends ReactComponentProps
-  > ReactElement<?, ?> createElementWithThemeContext(
```
nickgros, line 32: `createElementWithThemeContext` was a hacky workaround for cases where we could not use dependency injection. Now that we do not need to use DI, we can remove this function and replace it with `createElementWithSynapseContext`

src/main/java/org/sagebionetworks/web/client/jsinterop/React.java
```diff
@@ -59,12 +52,16 @@ > ReactElement<T, P> createElement(
...
+      props,
+      children
+    );
     return createElement(
-      SRC.SynapseContext.FullContextProvider,
-      wrapperProps,
+      ContextUtils.SynapseContextProviderFromStore,
+      null,
       componentElement
     );
```
nickgros, line 66: Now, `createElementWithSynapseContext` uses the global `SynapseContextProviderFromStore` that we created in globalContext.js.

src/main/java/org/sagebionetworks/web/client/presenter/TrashPresenter.java
```diff
@@ -14,22 +13,17 @@ public class TrashPresenter
 
   private Trash place;
   private TrashView view;
-  private SynapseReactClientFullContextPropsProvider propsProvider;
```
nickgros, line 17: Many of the changes in this PR look like this, where we remove the injected `propsProvider`, which is now unnecessary. Core changes are called out with PR comments.

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentV2.java
```diff
@@ -56,7 +56,11 @@ public ReactComponentV2(T reactComponentType, P props) {
   public ReactComponentV2(T reactComponentType, P props, String tag) {
     this.reactComponentType = reactComponentType;
     this.props = props;
-    setElement(Document.get().createElement(tag));
+    Element el = Document.get().createElement(tag);
+    // For the element that wraps this React root, `display: contents;`, which causes an element's children to appear
+    // as if they were direct children of the element's parent
+    el.setAttribute("style", "display: contents;");
+    setElement(el);
```
nickgros, line 63: Minor improvement to all React components -- use `display: contents;` to ignore the wrapper element!

src/main/java/org/sagebionetworks/web/client/SessionDetector.java
jay-hodgson: The checkForUserChange() is asynchronous.  If you want this to run after the user check (where it looks for the access token cookie), then this line should be inside an AsyncCallback passed to checkForUserChange().  If it doesn't need to be after this initialization, then I would move this line somewhere else (perhaps in Portal.java after ginjector.getSessionDetector().start(), or maybe a new globalApplicationState. synchronizeReactContextWithGlobalStore() called from Portal.java)

src/main/java/org/sagebionetworks/web/client/widget/ReactComponentV2.java
jay-hodgson: Nice potential accessibility improvement

src/main/java/org/sagebionetworks/web/client/SessionDetector.java
nickgros: I'll move it into GlobalApplicationState and call it from Portal.java. Since we refresh if the user changes, I don't think we need to re-sync each time the session checker runs.

jay-hodgson (approved):

PR #5741, opened 2025-07-16
SWC-7004 - Proof of concept MUI button

nickgros:
image


src/main/resources/org/sagebionetworks/web/client/view/ProfileViewImpl.ui.xml
```diff
@@ -79,12 +79,14 @@
...
-                        icon="PLUS"
-                        pull="RIGHT"
-                        text="Create a New Project"
+                        variant="contained"
+                        color="primary"
+                        startIcon="add"
+                        children="Create a New Project"
+                        className="pull-right"
+                        size="small"
                       />
```
nickgros, line 90: This component uses the correct theme (Synapse theme, not default MUI theme) because the theme (and other context) is now available globally without DI.

image

jay-hodgson: Nice!  
Does this screenshot show a small size button, or was the code updated after the screenshot was taken?

nickgros: This is a small size button. As you probably noticed, it doesn't match--I put this into a separate PR so we can sort all of those concerns out separately :slightly_smiling_face:

PR #5754, opened 2025-07-29
SWC-6951: Inconsistent ways for users to ‘click to copy’ items

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=SWC-6951

Got the OK for this from Adam in Slack messaging: "I think that's better than what we have now, so I'd support that change. There may be a different way to do it, but we can worry about that later. I say go for it."

Screenshot 2025-07-29 at 10 53 17 AM
Screenshot 2025-07-28 at 5 06 47 PM

### DIFF
--- a/src/main/webapp/sass/_core.scss
+++ b/src/main/webapp/sass/_core.scss
@@ -1862,11 +1862,6 @@ h6 {
   text-decoration: none;
 }
 
-.no-underline {
-  text-decoration: none !important;
-  cursor: default !important;
-}
-
 /** SWC-4059: fix focus by changing tree items to block (so that top and left positions are set properly) **/
 .gwt-TreeItem {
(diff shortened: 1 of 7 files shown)


src/main/webapp/sass/_core.scss
```diff
@@ -1862,6 +1862,11 @@ h6 {
   text-decoration: none;
 }
 
+.no-underline {
+  text-decoration: none !important;
+  cursor: default !important;
+}
```
nickgros: seems like this isn't used anywhere?

nickgros (approved):

PR #5756, opened 2025-08-11
SWC-7420

nickgros:
- Upgrade pnpm
- Move `resolutions` to `pnpm.overrides`
- Delete unused patch file for removed dependency


package.json
```diff
@@ -74,6 +68,14 @@
     "vite-plugin-node-polyfills": "0.17.0",
     "wait-on": "^8.0.3"
   },
+  "pnpm": {
+    "overrides": {
```
nickgros, line 72: `pnpm.overrides` is the same as `resolutions` (which is preserved for compatibility with yarn). In pnpm 10, linking a package adds an entry to this overrides section, so we should use that instead of resolutions.

kianamcc (approved):

PR #5758, opened 2025-08-13
PORTALS-3709 Synapse Grid

kianamcc:
Related Jira: https://sagebionetworks.jira.com/browse/PORTALS-3709?atlOrigin=eyJpIjoiNGIyMjZiNjI3NWNjNDk5OTg0NGMwOWUxYTVmZWExMjciLCJwIjoiaiJ9

Related PR in SWM: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2087


src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
```diff
@@ -301,6 +302,20 @@ private void configureActions() {
...
+    this.actionMenu.setActionListener(
+        Action.SHOW_GRID,
+        (action, e) -> {
+          boolean isVisible = !view.isSynapseGridVisible();
+          view.setSynapseGridVisible(isVisible);
+          // Toggle QueryWrapperPlotNav visibility - hide when grid is shown, show when grid is hidden
+          view.setQueryWrapperPlotNavVisible(!isVisible);
+          if (isVisible && currentQuery != null) {
+            view.configureSynapseGrid(currentQuery.getSql());
+          }
```
nickgros: Not sure how this will behave as you toggle between them but I suppose we can refine that behavior later.

I think maybe an imperative call to instantiate the grid could go here to avoid the `useEffect` to instantiate the session in the SynapseGrid component (via React's `useImperativeHandle`)

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1293,6 +1293,10 @@ private void configureTableCommands() {
         "Upload Data to " + entityTypeDisplay
       );
       actionMenu.setActionVisible(Action.EDIT_TABLE_DATA, canEditResults);
+      actionMenu.setActionVisible(
+        Action.SHOW_GRID,
+        featureFlagConfig.isFeatureEnabled(FeatureFlagKey.SYNAPSE_GRID)
+      );
```
nickgros, line 1299: We may need to prevent showing this for specific kinds of tables (I don't know which ones work and which don't)

e.g. we would never need a grid for a SubmissionView

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
kianamcc: Not sure why I put it there at that time but removed for SubmissionView!

PR #5762, opened 2025-08-21
SWC-7436 - Fix redirect on unsigned ToS

nickgros:
- Use SynapseJavaScript client request instead of servlet proxy to get current user profile in `AuthenticationController`
- Fix incorrect ToS error checking (users will now be immediately redirected to sign the ToS instead of logged out)

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
@@ -60,7 +60,6 @@ public class AuthenticationControllerImpl implements AuthenticationController {
   private List<String> persistentLocalStorageKeys;
   private String currentUserAccessToken;
   private UserProfile currentUserProfile;
-  private UserAccountServiceAsync userAccountService;
   private ClientCache localStorage;
   private SessionStorage sessionStorage;
   private PortalGinInjector ginInjector;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -215,8 +211,11 @@ public void onSuccess(String accessToken) {
           } else if (forceResetQueryClientCache) {
             resetQueryClientCache();
           }
-          userAccountService.getMyProfile(
-            new AsyncCallback<UserProfile>() {
```
nickgros, line 219: Tangentially related, I discovered this issue when I realized in my local stack sometimes this call would hang (in development only)-- replacing the RPC call to the servlet with a request sent directly via the JS client fixes that intermittent issue.

src/main/java/org/sagebionetworks/web/client/security/AuthenticationControllerImpl.java
```diff
@@ -229,9 +228,7 @@ public void onFailure(Throwable t) {
                 currentUserProfile = null;
                 if (
                   t instanceof ForbiddenException &&
-                  ((ForbiddenException) t).getMessage()
-                    .toLowerCase()
-                    .contains("terms of use")
+                  t.getMessage().toLowerCase().contains("terms of service")
```
nickgros, line 231: The thrown exception changed at some point

jay-hodgson (approved):

PR #5764, opened 2025-08-22
SWC-7433: Dependabot: form-data uses unsafe random function in form-data for choosing boundary (Critical)

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7433

Upgrade form-data to patched version 4.0.4 which moves away from using Math.random() to crypto.randomBytes(): https://github.com/form-data/form-data/commit/3d1723080e6577a66f17f163ecd345a21d8d0fd0.


kianamcc:
@nickgros Updating form-data also updates its dependencies, one of them being MUI. Do updates in SWC mean that the Synapse Web Monorepo must be updated to, for example, use the same MUI version (7.1.1 -> 7.3.1)?

7.3.1 patch notes: https://github.com/mui/material-ui/releases/tag/v7.3.1

nickgros:
@kianamcc Yes, we need to keep the MUI version in SWC synchronized with synapse-react-client.

That said, MUI does not depend on `form-data`, so we should not have to update it to upgrade form-data

kianamcc:
Updated wait-on from 8.0.3 to 8.0.4 to get form-data 4.0.4 (via axios 1.11.0).

I didn't realize form-data was a nested dependency:

devDependencies:
wait-on 8.0.3
└─┬ axios 1.9.0
  └── form-data 4.0.2

nickgros (approved):

PR #5771, opened 2025-09-16
SWC-7467 - Basic RecordSet support

nickgros:
- Add basic support for RecordSets

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
     <webappDirectory>
       ${project.build.directory}/${project.build.finalName}
     </webappDirectory>
-    <synapse.version>521.0</synapse.version>
+    <synapse.version>560.0</synapse.version>
     <gwtVersion>2.11.0</gwtVersion>
     <org.springframework.version>5.3.37</org.springframework.version>
     <guiceVersion>6.0.0</guiceVersion>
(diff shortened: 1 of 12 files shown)


src/main/java/org/sagebionetworks/web/client/SynapseClient.java
```diff
@@ -225,8 +225,6 @@ Boolean isAliasAvailable(String alias, String aliasType)
   HashMap<String, WikiPageKey> getPageNameToWikiKeyMap()
     throws RestServiceException;
 
-  String deleteApiKey() throws RestServiceException;
```
nickgros, line 228: This called `SynapseJavaClient.deleteApiKey`, which was removed since the last time we updated `SynapseJavaClient`.

pom.xml
```diff
@@ -18,7 +18,7 @@
     <webappDirectory>
       ${project.build.directory}/${project.build.finalName}
     </webappDirectory>
-    <synapse.version>521.0</synapse.version>
+    <synapse.version>560.0</synapse.version>
```
jay-hodgson, line 21: Wow, it's been a while

jay-hodgson (approved):

PR #5777, opened 2025-09-23
SWC-7468 - Metadata Tab

nickgros:
- Add the MetadataTab under a feature flag.
image

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/EntityCitation.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/EntityCitation.java
@@ -1,0 +1,7 @@
+package org.sagebionetworks.web.client.widget;
+
+import com.google.gwt.user.client.ui.IsWidget;
+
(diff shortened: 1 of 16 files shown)


src/main/java/org/sagebionetworks/web/client/widget/EntityCitation.java
```diff
@@ -0,0 +1,7 @@
+package org.sagebionetworks.web.client.widget;
+
+import com.google.gwt.user.client.ui.IsWidget;
+
+public interface EntityCitation extends IsWidget {
```
nickgros, line 5: I couldn't figure out why, but something in my changes in `EntityPageTopTest` triggered loading the `EntityCitationImpl`  class instead of just using a mock, which caused issues since it uses JsInterop bindings. The issue went away when I introduced an interface to hide the injected impl.

jay-hodgson (approved):

PR #5809, opened 2025-11-03
SWC-7536: patch release 567 with fallback behavior when origin is not…

jay-hodgson:
… set

### DIFF
--- a/src/test/java/org/sagebionetworks/web/unitserver/filter/XFrameOptionsFilterTest.java
+++ b/src/test/java/org/sagebionetworks/web/unitserver/filter/XFrameOptionsFilterTest.java
@@ -7,6 +7,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.sagebionetworks.web.server.servlet.filter.CORSFilter.ORIGIN_HEADER;
 import static org.sagebionetworks.web.server.servlet.filter.XFrameOptionsFilter.DENY;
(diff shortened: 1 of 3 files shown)


src/test/java/org/sagebionetworks/web/unitserver/filter/XFrameOptionsFilterTest.java
```diff
@@ -92,4 +93,24 @@ public void testAllowedSynapseSubdomainOrigin()
...
+    throws ServletException, IOException {
+    StringBuffer sb = new StringBuffer();
+    sb.append("https://www.synapse.org/index.html");
+    when(mockRequest.getRequestURL()).thenReturn(sb);
+    when(mockRequest.getQueryString()).thenReturn(null);
+    // Origin header is not set
+    when(mockRequest.getHeader(ORIGIN_HEADER)).thenReturn(null);
+    // Referer header contains allowed synapse subdomain
+    when(mockRequest.getHeader(REFERER_HEADER))
+      .thenReturn("https://eliteportal.synapse.org");
```
nickgros: The `Referer` header may be an entire URL. Does this strategy work with a full URL?

nickgros (approved):

src/test/java/org/sagebionetworks/web/unitserver/filter/XFrameOptionsFilterTest.java
jay-hodgson: Good point, I'll update to handle this case!

nickgros (approved):

PR #5812, opened 2025-11-06
SWC-7553 - Use new GridPage component that handles search params

nickgros:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2349

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/place/GridPlace.java
+++ b/src/main/java/org/sagebionetworks/web/client/place/GridPlace.java
@@ -3,21 +3,17 @@ package org.sagebionetworks.web.client.place;
 import com.google.gwt.place.shared.Place;
 import com.google.gwt.place.shared.PlaceTokenizer;
 import com.google.gwt.place.shared.Prefix;
+import com.google.gwt.user.client.Window;
 
 public class GridPlace extends Place {
 
-  private String gridSessionId;
+  public static final String SESSION_ID = "sessionId";
 
   public GridPlace(String token) {
-    this.gridSessionId = token;
-  }
-
-  public String toToken() {
-    return gridSessionId;
-  }
-
-  public String getGridSessionId() {
-    return gridSessionId;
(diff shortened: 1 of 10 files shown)


src/main/resources/org/sagebionetworks/web/client/view/GridPageViewImpl.ui.xml
```diff
@@ -7,23 +7,6 @@
   xmlns:b="urn:import:org.gwtbootstrap3.client.ui"
 >
   <g:HTMLPanel>
-    <div>
-      <div class="pageHeader">
```
nickgros, line 11: The new component will render a page header (and put data in it)

src/main/java/org/sagebionetworks/web/client/place/GridPlace.java
```diff
@@ -3,29 +3,25 @@
...
-
-  public String toToken() {
-    return gridSessionId;
-  }
-
-  public String getGridSessionId() {
-    return gridSessionId;
+    if (token != null && !token.startsWith("default")) {
+      // Redirect /Grid:{sessionId} to /Grid:default?sessionId={sessionId}
+      Window.Location.assign("/Grid:default?" + SESSION_ID + "=" + token);
```
jay-hodgson, line 15: In the past, added search query params tend to follow the user around while browsing in the GWT SPA (placechanger) framework.  I thought we avoided these and just used the place tokens in the DataAccessManagement place for this reason.  Have you tried browsing around in SWC after adding the search params?

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/place/GridPlace.java
nickgros: Tested and there seems to be no issues. And we do use these params in the DataAccessManagement place, and it has not seemed to cause user issues so far.

PR #5824, opened 2025-11-17
SWC-7571: On first use, sharing settings triggers an unsaved change warning from browser

kianamcc:
project tools -> project sharing settings -> make change -> save -> “Any unsaved changes may be lost. Are you sure that you would like to navigate away from this editor?”

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/sharing/EntityAccessControlListModalWidgetImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/sharing/EntityAccessControlListModalWidgetImpl.java
@@ -4,6 +4,7 @@ import static org.sagebionetworks.web.client.jsinterop.SRC.SynapseComponents.Ent
 
 import com.google.gwt.user.client.ui.Widget;
 import com.google.inject.Inject;
+import org.sagebionetworks.web.client.GlobalApplicationState;
 import org.sagebionetworks.web.client.jsinterop.EntityAclEditorModalProps;
 import org.sagebionetworks.web.client.jsinterop.React;
 import org.sagebionetworks.web.client.jsinterop.ReactElement;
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/widget/sharing/EntityAccessControlListModalWidgetImpl.java
```diff
@@ -39,8 +45,16 @@ public void configure(
       EntityAclEditorModalProps.create(
         entityId,
         false,
-        onUpdateSuccess,
-        () -> setOpen(false),
+        () -> {
+          globalApplicationState.setIsEditing(false);
+          if (onUpdateSuccess != null) {
+            onUpdateSuccess.run();
```
kianamcc: global editing state off on successful save

src/main/java/org/sagebionetworks/web/client/widget/sharing/EntityAccessControlListModalWidgetImpl.java
```diff
@@ -39,8 +45,16 @@ public void configure(
...
+          }
+        },
+        () -> {
+          setOpen(false);
```
kianamcc, line 50: global editing state off on on close

kianamcc:
Can be reproduced on Edge too: 
Screenshot 2025-11-17 at 2 53 42 PM

src/main/java/org/sagebionetworks/web/client/widget/sharing/EntityAccessControlListModalWidgetImpl.java
jay-hodgson: Would it make more sense to add the to the onUpdateSuccess function being used?  Namely, 
add  `getGlobalApplicationState().setIsEditing(false);` to `EntityActionControllerImpl.fireEntityUpdatedEvent()`

kianamcc: I think it makes more sense for a controller to update this rather than in the modal. I assume it's ok to leave the close state as is for immediate cleanup as no entity update occurs to call fireEntityUpdatedEvent()?

jay-hodgson (approved):

PR #5834, opened 2025-12-10
SWC-7502: Load SWC stylesheets through Vite 

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=SWC-7502

Playwright visual comparison regression testing: https://playwright.dev/docs/test-snapshots

Axe for accessibility testing: https://playwright.dev/docs/accessibility-testing


pom.xml
```diff
@@ -315,7 +315,6 @@
                 ${project.basedir}/src/main/webapp/generated/vite
               </directory>
               <directory>${project.basedir}/src/main/webapp/css</directory>
```
nickgros, line 317: What CSS files do we load from here? You could also pull these in with vite if it doesn't cause any problems

e2e/visual-regression.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+ *    pnpm e2e:report
+ *
+ * NOTE: Accessibility snapshots track violations to ensure no NEW violations
+ * are introduced by your changes. Existing violations are preserved.
+ */
+
+/**
+ * Reusable helper to test a page for visual and accessibility regressions
+ */
+async function testPageVisualAndAccessibility(
```
nickgros: Is the idea that we run these manually when we make a broad style change? Or are these intended to be run on every build?

e2e/visual-regression.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+    await page.waitForSelector('[role="presentation"]', { state: 'visible' })
+    await page.waitForTimeout(500)
+
+    await expect(page).toHaveScreenshot('applications-dropdown.png', {
+      maxDiffPixelRatio: 0.01,
+    })
+  })
+})
+
+testAuth.describe(
```
nickgros: I assume that these results will vary based on the account used (and maybe also the current state of the data associated with that account)

e2e/visual-regression.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+  await page.waitForTimeout(waitTime)
+
+  // Visual regression test
+  await expect(page).toHaveScreenshot(`${name}.png`, {
+    fullPage,
+    maxDiffPixelRatio: 0.01,
+  })
+
+  // Accessibility snapshot
+  const accessibilityScanResults = await new AxeBuilder({ page }).analyze()
```
nickgros: I'm unfamiliar with this package--what's your opinion of it? Is it helpful? How many accessibility violations do we have at baseline?

pom.xml
kianamcc: We seem to be be loading fonts and some css files here: 
Screenshot 2025-12-10 at 9 35 25 AM

e2e/visual-regression.spec.ts
kianamcc: I had it in mind that we would run these manually as needed for migrations and such!

e2e/visual-regression.spec.ts
kianamcc: Yes it is my understanding that the screenshots that get uploaded into the output folder is specific to the user’s account. So the developer would probably want to avoid pushing their screenshots by adding the output folder in .gitignore when not in use.

e2e/visual-regression.spec.ts
kianamcc: The package itself is pretty popular and is still being maintained so I think it's safe to use! It scans the Playwright page for common accessibility issues and outputs a json with a list of accessibility violations. I intended to use this to make sure that a feature branch isn’t bringing in additional violations in the visual tests but I think it's scanning feature is pretty cool! Unfortunately, the only closest thing I can think of that I worked with similar to this is storybook's accessibility add on 😅

kianamcc:
Working on this ticket opened a bit of a rabbit hole into visual regression testing. 

This adds coverage for detecting unwanted visual changes such as CSS regressions, layout shifts, and other UI issues we currently don't.

These tests are intended to be run locally as a person would have to review the diffs to determine whether those changes were expected.

Example use case: For this ticket, I wanted to make sure that loading swc stylesheets through vite does not cause unintended visual changes.

The visual regression test file can be expanded to provide broader coverage but I think a separate ticket to iterate on and improve it makes sense. This is just the start.

Here I run the test file (authenticated-pages.spec.ts) with SWC-7502 and the develop branch (on firefox):
Screenshot 2025-12-30 at 1 49 48 PM


I noticed webkit is more likely to fail tests. Look into ways to improve tests for webkit.

pom.xml
kianamcc: Looks like pulling the CSS from webapp/css into Vite causes styling/layout issues. Maybe in another ticket we can investigate migrating these styles individually and see which ones won't cause any issues?

e2e/visual-regression/authenticated-pages.spec.ts
```diff
@@ -0,0 +1,1037 @@
...
+    await userPage.waitForTimeout(LONGER_WAIT_TIME)
+
+    await userPage.addStyleTag({
+      content: `
+        /* Hide the second h1 in the dashboard */
+      .MuiBox-root h1:nth-of-type(2) {
+        visibility: hidden !important;
+      }
+      `,
+    })
```
nickgros: These selectors are very brittle. Hiding them in this way is probably not a viable way to maintain snapshot tests

e2e/visual-regression/authenticated-pages.spec.ts
```diff
@@ -0,0 +1,1037 @@
...
+  testAuth('all projects page', async ({ userPage }) => {
+    const userId = await getUserIdFromLocalStorage(userPage)
+
+    // Navigate to the page and wait for it to load
+    await userPage.goto(`/Profile:${userId}/projects/all`)
+    await waitForInitialPageLoad(userPage)
+    await userPage.waitForTimeout(LONGER_WAIT_TIME)
+
+    // Wait for the specific element
+    const selector = '#rootPanel .rootPanel .ProfileViewImpl'
```
nickgros: If we are going to wait for something to appear, we should use something more stable than a class selector (like an accessible role or some specific text that we can guarantee will appear)

e2e/visual-regression/authenticated-pages.spec.ts
```diff
@@ -0,0 +1,1037 @@
...
+    await userPage.waitForSelector(selector, { timeout: 10000 })
+
+    // Hide dynamic project names to prevent false failures
+    await userPage.addStyleTag({
+      content: `
+        /* Hide project names that vary between test runs */
```
nickgros: I get the idea but I don't think we can maintain this. Is it maybe possible to just snapshot _part_ of the document which doesn't change between runs?

A reasonable outcome of this work is deciding that this type of testing doesn't make sense for us, which would be ok! This would still be valuable research

nickgros: We could also keep the unauthenticated snapshots which may still have some value

e2e/helpers/visualRegression.ts
```diff
@@ -0,0 +1,207 @@
...
+import { waitForInitialPageLoad } from './utils'
+
+// ======================================================
+// Constants
+// ======================================================
+export const LONGER_WAIT_TIME = 2000
+export const SHORTER_WAIT_TIME = 500
+export const PAGE_SELECTOR = '#rootPanel .rootPanel'
+export const PROJECT_HEADER_SELECTOR = '.pageHeader'
+export const PROJECT_TAB_PAGE_SELECTOR = '.tab-content'
```
nickgros: I would want to avoid using class name as a selector for any integration/e2e test as much as possible.

e2e/helpers/visualRegression.ts
```diff
@@ -0,0 +1,207 @@
...
+    })
+  }
+
+  // Accessibility snapshot
+  const accessibilityScanResults = selector
+    ? await new AxeBuilder({ page }).include(selector).analyze()
+    : await new AxeBuilder({ page }).analyze()
+
+  // Normalize dynamic CSS class names to prevent false failures between builds
+  const normalizedViolations = accessibilityScanResults.violations.map(
```
nickgros: Can you say more about this?

e2e/helpers/visualRegression.ts
```diff
@@ -0,0 +1,207 @@
...
+      })),
+    }),
+  )
+
+  const violationsJson = JSON.stringify(normalizedViolations, null, 2)
+
+  expect(violationsJson).toMatchSnapshot(`${name}-a11y.json`)
+}
+
+export const pageElementInteractions = async (
```
nickgros: It is not clear to me what the purpose of this function is

e2e/helpers/visualRegression.ts
kianamcc: This is for styles/ids that are generated at runtime and can vary. For example we could get something like css-abcd in the develop branch when we update screenshots then something like css-efgh in a feature branch for the same element so by normalizing this we can make sure that the test doesn't fail and focuses on the structure of the HTML rather than the randomly generated hashes.

e2e/helpers/visualRegression.ts
kianamcc: This was to be a helper function to handle clicking buttons for example and taking screenshots but looking at it now, I can see how the purpose of this function can be confusing with combining logic for interacting with elements on the page and screenshotting logic.

kianamcc:
4px padding difference in datetimepicker:
Screenshot 2026-01-26 at 4 29 34 PM

nickgros (approved):

PR #5836, opened 2025-12-15
PORTALS-3819

lakikowolfe:
Set up SWC to render react component search page (SearchPageResults) at SearchV2

Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2391
https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2431


src/main/java/org/sagebionetworks/web/client/presenter/SearchV2Presenter.java
```diff
@@ -0,0 +1,34 @@
...
+  implements Presenter<SearchV2Place> {
+
+  private final SearchV2View view;
+
+  @Inject
+  public SearchV2Presenter(SearchV2View view, SynapseJSNIUtils jsniUtils) {
+    this.view = view;
+
+    jsniUtils.setPageTitle(DisplayConstants.LABEL_SEARCH);
+  }
```
nickgros, line 20: We will need to pass the query term into this class and include it in the page title, or delegate this behavior to the SearchPage react component

src/main/java/org/sagebionetworks/web/client/place/SearchV2Place.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class SearchV2Place extends Place {
+
+  public static final String QUERY = "query";
+
+  public SearchV2Place(String token) {
+    if (token != null && !token.startsWith("default")) {
+      // Redirect /SearchV2:{token} to /SearchV2:default?query={token}
+      Window.Location.assign("/SearchV2:default?" + QUERY + "=" + token);
+    }
```
nickgros: Not sure if we need this redirect, unless I am missing something

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/place/SearchV2Place.java
lakikowolfe: I was just following the convention in `GridPlace.java`. I will remove it.

src/main/java/org/sagebionetworks/web/client/presenter/SearchV2Presenter.java
lakikowolfe: Moved this to synapse-web-monorepo and will be handled in the related PR: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2431

src/main/java/org/sagebionetworks/web/client/place/SearchV2Place.java
nickgros: Ah, yeah, that was added because the Grid URL changed after the page was first released, so it was an actual redirect users may have needed. Since this is unreleased, you're good to omit it.

nickgros (approved):

PR #5845, opened 2026-01-21
fix binding issues for docker services, remove host network-mode

nickgros:


docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     image: maven:3.9.9-amazoncorretto-17-al2023
     environment:
       - ORG_SAGEBIONETWORKS_REPOSITORYSERVICE_ENDPOINT=${ORG_SAGEBIONETWORKS_REPOSITORYSERVICE_ENDPOINT}
-    command: mvn -Dorg.sagebionetworks.repositoryservice.endpoint=${ORG_SAGEBIONETWORKS_REPOSITORYSERVICE_ENDPOINT} gwt:run-codeserver
+    command: mvn -Dorg.sagebionetworks.repositoryservice.endpoint=${ORG_SAGEBIONETWORKS_REPOSITORYSERVICE_ENDPOINT} -Dgwt.bindAddress=0.0.0.0 gwt:run-codeserver
```
nickgros, line 9: Bind the GWT codeserver to 0.0.0.0 so it is reachable outside of the container

docker-compose.yml
```diff
@@ -17,7 +17,6 @@ services:
       - maven-repo:/root/.m2
       - ${HOME}/.m2/settings.xml:/root/.m2/settings.xml:ro
       - generated_artifacts:/usr/src/mymaven/src/main/webapp/generated
-    network_mode: host
```
nickgros, line 20: remove `network-mode: host`, which only works on linux

vite.config.ts
```diff
@@ -8,6 +8,7 @@ const config = defineConfig({
   // Assets will be served from the `/Portal/cdn` servlet, which will redirect requests to the CDN when appropriate
   base: '/Portal/cdn/generated/vite/',
   server: {
+    host: '0.0.0.0', // Bind to all interfaces so Docker can access it
```
nickgros, line 11: Bind the vite codeserver to 0.0.0.0 so other applications can connect to it

jay-hodgson:
Regarding the removal of `network-mode: host`, claude tells me "Using bridge networking (default) with explicit port mappings is more portable and secure". 
With the explicit bindings that you've set, if you are able to confirm that SWC starts up properly with this configuration (codeserver is accessible and client works as expected), then this changeset looks good to me!

jay-hodgson (approved):

PR #5854, opened 2026-02-12
SWC-7660

jay-hodgson:


jay-hodgson:
Cytoscape viz works without the proxy parameter:
Screenshot 2026-02-11 at 4 44 22 PM

Even if param is supplied, servlet redirects and downloads the attachment:
Screenshot 2026-02-11 at 4 44 05 PM

nickgros (approved):

PR #5857, opened 2026-02-18
SWC-7671: move to SRC entity ACL editor and remove GWT entity ACL editor

jay-hodgson:
Builds on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2509 (since SRC entity ACL editor supports realm-specific entries)

Replaced in other contexts, including from the Download Cart:
Screenshot 2026-02-17 at 3 59 32 PM

Screenshot 2026-02-17 at 3 59 41 PM


src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
```diff
@@ -723,32 +722,15 @@ public QueryResultBundle getCurrentQueryResultBundle() {
...
-  private AccessControlListModalWidget getAccessControlListModalWidget() {
+  private EntityAccessControlListModalWidget getAccessControlListModalWidget() {
     if (aclModal == null) {
-      aclModal = ginInjector.getAccessControlListModalWidget();
+      aclModal = ginInjector.getEntityAccessControlListModalWidget();
     }
     return aclModal;
   }
 
   public void onViewSharingSettingsClicked(String benefactorEntityId) {
```
nickgros, line 732: Could we just update the react component to show the modal instead of using this historical indirection?

nickgros (approved):
I think there's a more straightforward approach but this is still an improvement and unblocks further realms work!

src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
jay-hodgson: YES!  I was going to move forward with this refactor, but thought of this more iterative solution.  Will open up a follow-up ticket

PR #5863, opened 2026-02-24
remove unnecessary padding/outline from file browser

nickgros:
Fixing a minor annoyance

| Before | After |
| --- | --- |
| image | image |


src/main/resources/org/sagebionetworks/web/client/widget/entity/browse/FilesBrowserViewImpl.ui.xml
```diff
@@ -19,10 +19,9 @@
       <bh:Div ui:field="actionMenuContainer" />
     </bh:Div>
     <bh:Div ui:field="addToDownloadListContainer" />
-    <bh:Div ui:field="reactFileBrowserContainer" />
```
nickgros, line 22: unused

jay-hodgson (approved):

PR #5864, opened 2026-02-27
SWC-7717

jay-hodgson:


jay-hodgson:
Replaces Down GWT place with a standalone Synapse unavailable page that periodically checks stack status.  If stack status check fails, surfaces that error.
Did not remove SynapseStatusDetector, and rely on it to set endpoint and redirect URL in the session storage.

Screenshot 2026-02-27 at 11 10 45 AM

xschildw (approved):
Sounds good to me, I can try on 579 while it's in RO mode.

nickgros (approved):

PR #5874, opened 2026-03-06
add pnpm link/unlink scripts for local SRC development

hallieswan:


package.json
```diff
@@ -85,6 +85,10 @@
     "start:production": "docker compose --env-file .env.production up",
     "start:staging": "docker compose --env-file .env.staging up",
     "start:dev": "docker compose --env-file .env.development up",
+    "start:dev:link": "pnpm run link:src && docker compose --env-file .env.development -f docker-compose.yml -f compose-link.yml up",
```
nickgros, line 88: May be more practical to create a node script that takes CLI arguments for the stack and whether to link, up to you. That could be a later iteration

nickgros (approved):

hallieswan:
Closing this for now, we can reopen if we decide it would be useful in the future.

PR #5888, opened 2026-03-18
SWC-7752: Release SearchV2 out of Experimental Mode

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7752?search_id=e44537d4-a8a1-4ae3-b2c5-5a24e60d4e9a

Screenshot 2026-03-18 at 10 20 15 AM
Screenshot 2026-03-18 at 10 21 19 AM


jay-hodgson:
Could you also remove the block `} else if (path.startsWith("/Search")) {` from `HtmlInjectionFilter`, and update `/Search:` links to valid `/SearchV2:` links in `CrawlFilter`?

jay-hodgson:
Are there any other references in the codebase to the old search place? `/Search:`

kianamcc:
> Are there any other references in the codebase to the old search place? `/Search:`

Therre was one in opensearch.xml and some old search related tests in HtmlInjectionFilterTests.java

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
```diff
@@ -596,7 +596,7 @@ public String getAllProjectsHtml(SearchQuery inputQuery)
     inputQuery.setStart(newStart);
     String newJson = EntityFactory.createJSONStringForEntity(inputQuery);
     html.append(
-      "<a href=\"https://www.synapse.org/Search:" +
+      "<a href=\"https://www.synapse.org/SearchV2:" +
```
jay-hodgson: I don't think this is right.  I would expect something like `<a href=\"https://www.synapse.org/SearchV2:default?query=`

src/main/webapp/opensearch.xml
```diff
@@ -7,7 +7,7 @@
   <Url
     method="GET"
     type="text/html"
-    template="https://www.synapse.org/Search:{searchTerms}"
+    template="https://www.synapse.org/SearchV2?query={searchTerms}"
```
jay-hodgson: Not sure if this will be handled properly either, since it's missing a GWT place token.  Expect `/SearchV2:default?query={searchTerms}`

src/main/java/org/sagebionetworks/web/server/servlet/filter/CrawlFilter.java
kianamcc: You're right, I missed the GWT place token and query parameter for all of these 😅

jay-hodgson (approved):

PR #5892, opened 2026-03-30
SWC-7760: fix challenges evaluation editor back navigation and unsaved changes modal display

hallieswan:
Fixes three navigation bugs in the challenges evaluation editor (see [SWC-7760](https://sagebionetworks.jira.com/browse/SWC-7760)):
1. the "Back to challenge" button was non-functional due to use of `ReactDOM.unmountComponentAtNode` (removed in React 18)
2. the "Unsaved changes" modal was incorrectly shown even when no edits had been made, because the editing flag was set unconditionally on editor open rather than on actual user edits and the modal continued to appear after saving changes

Updates in https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2620 will ensure React components accurately report when there are unsaved changes in the forms.

"Back to Challenge" button is functional: 

https://github.com/user-attachments/assets/af9374ca-a050-4253-88ba-55b25a4e09cf

"Unsaved changes" modal shown when there are changes in the form, but not after the form has been saved: 

https://github.com/user-attachments/assets/050d41e0-71ff-4010-a17c-2bdb3e24d588

"Unsaved changes" modal is not shown when form is clean and user clicks on another tab: 

https://github.com/user-attachments/assets/8d8ff294-cc78-46ae-8e4f-ecfc8465e1bc

[SWC-7760]: https://sagebionetworks.jira.com/browse/SWC-7760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
@@ -1,6 +1,5 @@
 package org.sagebionetworks.web.client.jsinterop;
 
-import com.google.gwt.dom.client.Element;
 import jsinterop.annotations.JsFunction;
 import jsinterop.annotations.JsPackage;
 import jsinterop.annotations.JsType;
@@ -13,8 +12,6 @@ public class ReactDOM {
   public interface Callback {
     void run();
   }
-
-  public static native boolean unmountComponentAtNode(Element container);
 
   /**
    * flushSync lets you force React to flush any updates inside the provided callback synchronously. This ensures that the DOM is updated immediately.
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTab.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTab.java
@@ -4,7 +4,9 @@ import com.google.inject.Inject;
 import java.util.function.Consumer;
 import org.sagebionetworks.repo.model.auth.UserEntityPermissions;
 import org.sagebionetworks.repo.model.entitybundle.v2.EntityBundle;
+import org.sagebionetworks.web.client.DisplayConstants;
 import org.sagebionetworks.web.client.GlobalApplicationState;
+import org.sagebionetworks.web.client.PopupUtilsView;
 import org.sagebionetworks.web.client.PortalGinInjector;
 import org.sagebionetworks.web.client.place.Synapse;
 import org.sagebionetworks.web.client.place.Synapse.EntityArea;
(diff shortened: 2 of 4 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTab.java
```diff
@@ -107,18 +112,30 @@ public void configure(
...
+            "",
+            DisplayConstants.NAVIGATE_AWAY_CONFIRMATION_MESSAGE,
+            () -> {
+              globalApplicationState.setIsEditing(false);
+              doNavigateBack.run();
+            }
+          );
+        } else {
+          doNavigateBack.run();
+        }
```
hallieswan, line 138: Show "Unsaved changes" modal when clicking on the "Back to Challenges" button, if there are unsaved edits. Otherwise, just navigate back

src/main/java/org/sagebionetworks/web/client/jsinterop/ReactDOM.java
```diff
@@ -14,8 +13,6 @@ public interface Callback {
     void run();
   }
 
-  public static native boolean unmountComponentAtNode(Element container);
```
hallieswan, line 17: Removed in React v18

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTab.java
```diff
@@ -107,18 +112,30 @@ public void configure(
   private void showEvaluationEditor(String entityId, String evaluationId) {
     EvaluationEditorReactComponentPage evaluationEditor =
       ginInjector.createEvaluationEditorReactComponentPage();
-    globalApplicationState.setIsEditing(true);
+    Runnable doNavigateBack = () -> {
+      evaluationEditor.removeFromParent();
```
nickgros, line 116: Can you verify if `removeFromParent` invokes `root.unmount` on the React tree rendered by this component? I think it should via `onUnload` but not 100% sure

nickgros (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/ChallengeTab.java
hallieswan: That's right! `removeFromParent` invokes `root.unmount()` on the React tree via an `onUnload` override. Here's the full chain:
1. `removeFromParent` triggers `onDetach` on the widget, which calls `onUnload` before anything else, see the [GWT Widget lifecycle docs](https://www.gwtproject.org/javadoc/latest/com/google/gwt/user/client/ui/Widget.html#onDetach()) and [Widget.java source](https://github.com/gwtproject/gwt/blob/main/user/src/com/google/gwt/user/client/ui/Widget.java#L386-L396)
2. [ReactComponent](https://github.com/Sage-Bionetworks/SynapseWebClient/blob/develop/src/main/java/org/sagebionetworks/web/client/widget/ReactComponent.java#L87-L91) overrides `onUnload` to call `destroyRoot`, which schedules `root.unmount`

PR #5904, opened 2026-04-08
SWC-7504: Replace CSV preview widget GWT implementation with React implementation

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/PORTALS-4145?atlOrigin=eyJpIjoiNTlhY2JlNWI3OGQ4NGY1MGExYzc0NmViMjY5YjM2OTEiLCJwIjoiaiJ9


src/main/java/org/sagebionetworks/web/client/jsinterop/ColumnModelJsObject.java
```diff
@@ -0,0 +1,17 @@
...
+
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+
+/**
+ * JsInterop bridge for the ColumnModel JavaScript object expected by the React CsvPreview component
+ * Only includes fields needed by GWT, add more as needed.
+ */
+@JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
+public class ColumnModelJsObject {
```
kianamcc: Saw this pattern in some other files (e.g. EvaluationJSObject). From my understanding, we would want to do this when we're passing around structured objects across gwt and js whether that's a prop or callback argument.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/upload/UploadCSVPreviewPageImpl.java
```diff
@@ -44,21 +37,18 @@ public class UploadCSVPreviewPageImpl
   @Inject
   public UploadCSVPreviewPageImpl(
     UploadCSVPreviewPageView view,
-    UploadPreviewWidget uploadPreviewWidget,
     CSVOptionsWidget csvOptionsWidget,
-    JobTrackingWidget jobTrackingWidget,
```
kianamcc, line 49: React component handles all the fetching, loading, and preview UI now so we can remove things related to job tracking and upload preview widget.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/upload/UploadCSVPreviewPageImpl.java
```diff
@@ -138,53 +127,63 @@ private UploadToTablePreviewRequest createDefaultPreviewRequest() {
...
 
-  /**
-   * Generate a new preview using the current options.
-   */
-  private void generatePreview() {
-    this.view.setPreviewVisible(false);
-    this.view.setTrackerVisible(true);
-    this.presenter.setPrimaryButtonText(NEXT);
-    this.presenter.setInstructionMessage(PREPARING_A_PREVIEW);
+  private void refreshPreview() {
```
kianamcc, line 129: Re-configure React component when csv options change.

src/main/java/org/sagebionetworks/web/client/widget/table/modal/upload/UploadCSVPreviewPageImpl.java
```diff
@@ -138,53 +127,63 @@ private UploadToTablePreviewRequest createDefaultPreviewRequest() {
...
+
+      if (jsColumn.maximumSize != null) {
+        column.setMaximumSize(jsColumn.maximumSize.longValue());
+      }
+
+      ColumnType columnType = ColumnType.STRING;
+      if (jsColumn.columnType != null) {
+        try {
+          columnType = ColumnType.valueOf(jsColumn.columnType);
+        } catch (IllegalArgumentException e) {
```
kianamcc: Not sure how likely it is we hit this case or how we should handle it?

nickgros: Could happen if we add a new column type and handle it in React without updating the model in SWC. Not very likely, so I think this is ok

src/main/java/org/sagebionetworks/web/client/jsinterop/ColumnModelJsObject.java
```diff
@@ -0,0 +1,17 @@
...
+
+  public String id;
+  public String name;
+  public String columnType; // e.g. "STRING", "BOOLEAN", etc.
+  public Double maximumSize;
```
nickgros: how do we know this is the complete set of ColumnModel fields we will use? what if the suggestion service (which is a backend service) adds support for another ColumnModel field?

Other suggested approaches don't have this drawback where this needs to enumerate every single used field.

src/main/java/org/sagebionetworks/web/client/jsinterop/UploadToTablePreviewResultJsObject.java
```diff
@@ -0,0 +1,10 @@
+package org.sagebionetworks.web.client.jsinterop;
+
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+
+@JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
+public class UploadToTablePreviewResultJsObject {
+
+  public ColumnModelJsObject[] suggestedColumns;
```
nickgros: Maybe to avoid capturing all of the ColumnModel properties in a new class, we do something like this?
```suggestion
  public Object[] suggestedColumns;
```

Then when you need the Java ColumnModel, you could do

```java
String columnModelAsJsonString = Global.JSON.stringify(column);
ColumnModel cm = new ColumnModel(columnModelAsJsonString);
```

There are similar utilities:
- JSONEntityUtils, which handles going from Java to JsInterop (the opposite of what you're doing, so it might be reasonable for that class to also have a method for this)
- SynapseJavascriptFactory#newInstance, which handles transforming a JSON string to a Java object (it basically does the code sample above)

Maybe you could use/extend/refactor one or both of these?

src/main/java/org/sagebionetworks/web/client/widget/table/modal/upload/UploadCSVPreviewPageImpl.java
```diff
@@ -138,53 +127,63 @@ private UploadToTablePreviewRequest createDefaultPreviewRequest() {
...
+
+    for (ColumnModelJsObject jsColumn : data.suggestedColumns) {
+      if (jsColumn == null) {
+        continue;
+      }
+
+      ColumnModel column = new ColumnModel();
+
+      column.setId(jsColumn.id);
+      column.setName(jsColumn.name);
```
nickgros: See comment on UploadToTablePreviewResultJsObject, basically I don't think this class should be responsible for object translation at this level of detail. The inside of this `for` loop should basically just be a call to a utility

src/main/java/org/sagebionetworks/web/client/jsinterop/UploadToTablePreviewResultJsObject.java
nickgros: Actually, I wonder if you could try `new UploadTableToPreviewResult()` instead and get rid of this object, too

src/main/java/org/sagebionetworks/web/client/jsinterop/ColumnModelJsObject.java
nickgros: It was reasonable for you to take this approach. I think in retrospect, this is a hacky way to do it. We created it when we were first trying to figure out how to interoperate code and I think we have better, more maintainable ways to do this now.

I suggested other approaches in another comment in this review.

src/main/java/org/sagebionetworks/web/client/jsinterop/CsvTableDescriptorJsObject.java
```diff
@@ -0,0 +1,29 @@
...
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+import org.sagebionetworks.repo.model.table.CsvTableDescriptor;
+
+/**
+ * JsInterop bridge for the CsvTableDescriptor plain JavaScript object expected by the React CsvPreview component
+ * Only includes fields needed by GWT, add more as needed.
+ */
+@JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
+public class CsvTableDescriptorJsObject {
```
nickgros: Since this is prop input I am ok with using a JsInterop type here. There might be a similar way as I suggested to handle this too since CsvTableDescriptor is also a type we have in Java, but I think we've only used this pattern for props so far.

src/main/java/org/sagebionetworks/web/client/jsinterop/CsvPreviewProps.java
```diff
@@ -0,0 +1,47 @@
...
+    OnCsvPreviewDataChangeFunction onCsvPreviewDataChange,
+    OnIsLoadingChangeFunction onIsLoadingChange
+  ) {
+    CsvPreviewProps props = new CsvPreviewProps();
+    props.fileHandleId = fileHandleId;
+    props.csvTableDescriptor = csvTableDescriptor;
+    props.onCsvPreviewDataChange = onCsvPreviewDataChange;
+    props.onIsLoadingChange = onIsLoadingChange;
+    return props;
+  }
```
nickgros: Related to comment on CsvTableDescriptorJsObject, would something like this be possible? 


```suggestion
  public Object csvTableDescriptor;

  @JsNullable
  public OnCsvPreviewDataChangeFunction onCsvPreviewDataChange;

  @JsNullable
  public OnIsLoadingChangeFunction onIsLoadingChange;

  @JsOverlay
  public static CsvPreviewProps create(
    String fileHandleId,
    CsvTableDescriptor csvTableDescriptor,
    OnCsvPreviewDataChangeFunction onCsvPreviewDataChange,
    OnIsLoadingChangeFunction onIsLoadingChange
  ) {
    CsvPreviewProps props = new CsvPreviewProps();
    props.fileHandleId = fileHandleId;
    props.csvTableDescriptor = JSONEntityUtils.toJsInteropCompatibleObject(csvTableDescriptor);
    props.onCsvPreviewDataChange = onCsvPreviewDataChange;
    props.onIsLoadingChange = onIsLoadingChange;
    return props;
  }
```

kianamcc:
Looks like we can just get rid of all the JsObject files! Added a new fromJsInteropCompatibleObject utility which does the opposite of toJsInteropCampatibleObject in JSONEntityUtils.

src/main/java/org/sagebionetworks/web/client/jsinterop/UploadToTablePreviewResultJsObject.java
kianamcc: Since we already know we're expecting a new UploadToTablePreviewResult and the SynapseJavascriptFactory newInstance seems to have to read concreteType from the json itself to figure out what to instantiate, does it make sense to just handle it all in the new JSONEntityUtils utility (fromJsInteropCompatibleObject)?

src/main/java/org/sagebionetworks/web/client/widget/table/modal/upload/UploadCSVPreviewPageImpl.java
```diff
@@ -2,33 +2,25 @@
...
 import org.sagebionetworks.repo.model.table.UploadToTableRequest;
-import org.sagebionetworks.web.client.widget.asynch.AsynchronousProgressHandler;
-import org.sagebionetworks.web.client.widget.asynch.JobTrackingWidget;
-import org.sagebionetworks.web.shared.asynch.AsynchType;
+import org.sagebionetworks.schema.adapter.JSONObjectAdapterException;
+import org.sagebionetworks.web.client.jsinterop.JSONEntityUtils;
+import org.sagebionetworks.web.client.widget.CsvPreview;
 
 public class UploadCSVPreviewPageImpl
   implements UploadCSVPreviewPage, UploadCSVPreviewPageView.Presenter {
```
nickgros, line 17: I would think about if it makes more sense to replace UploadCsvPreviewPage entirely with the CsvPreviewDialog component. Maybe there is sufficient logic here that it makes sense to keep UploadCsvPreviewPage, but if they really are the same then there might be even more code we can delete.

nickgros (approved):

PR #5909, opened 2026-04-10
SWC-7778

jay-hodgson:
Depends on https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2677


src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -232,20 +237,23 @@ public interface ProfileViewImplUiBinder
...
-    JSONObjectAdapter jsonObjectAdapter
+    JSONObjectAdapter jsonObjectAdapter,
+    GlobalApplicationState globalAppState
   ) {
     initWidget(binder.createAndBindUi(this));
     this.headerWidget = headerWidget;
     this.cookies = cookies;
     this.orientationBanner = orientationBanner;
     this.jsonObjectAdapter = jsonObjectAdapter;
+    this.globalAppState = globalAppState;
```
jay-hodgson, line 256: Inject GlobalApplicationState to avoid full page reload after Project creation

nickgros (approved):

PR #5912, opened 2026-04-14
PORTALS-4099 - Upgrade Vite to v8

nickgros:
Upgrade Vite to v8.0.8

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,13 +63,14 @@
     "npm-run-all": "^4.1.5",
     "prettier": "^3.7.4",
     "prettier-plugin-java": "2.5.0",
+    "qs": "^6.15.1",
     "semver": "^7.7.4",
     "tsx": "^4.19.4",
     "typescript": "5.8.3",
     "typescript-eslint": "^8.52.0",
     "uuid": "^9.0.1",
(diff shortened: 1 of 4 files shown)


src/main/webapp/sass/_core.scss
```diff
@@ -1648,7 +1648,6 @@ h6 {
 .previewtable {
   border: 1px solid #dddddd;
   border-collapse: separate;
-  *border-collapse: collapsed;
```
nickgros, line 1651: Vite 8's toolchain no longer supports whatever this is. I think it's an IE7 hack!

package.json
```diff
@@ -63,13 +63,14 @@
     "npm-run-all": "^4.1.5",
     "prettier": "^3.7.4",
     "prettier-plugin-java": "2.5.0",
+    "qs": "^6.15.1",
```
nickgros, line 66: Transitive dependency of `vite-plugin-node-polyfills` that did not resolve without installing it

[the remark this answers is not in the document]
nickgros: Needed for new version of vite-plugin-node-polyfills

jay-hodgson (approved):

PR #5913, opened 2026-04-15
SWC-7798: fix project search box on the user dashboard

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/view/ProfileViewImpl.java
```diff
@@ -321,9 +325,7 @@ private String getCurrentProjectSearchJSON() {
       projectsOnly.setValue("project");
       query.getBooleanQuery().add(projectsOnly);
 
-      query.setQueryTerm(
-        Arrays.asList(projectSearchTextBox.getValue().split("\\s+"))
-      );
+      query.setQueryTerm(Arrays.asList(projectSearchTextBox.getValue()));
```
kianamcc: Question, why was .split() removed here?

jay-hodgson: Nice catch.  I don't see why this change should be made (we should still split the search string to create individual terms).  Thank you!

xschildw (approved):

PR #5931, opened 2026-04-25
SWC-7793 - Curator Dashboard

nickgros:


src/main/webapp/css/bootstrap-custom-theme.min.css
```diff
@@ -396,7 +396,7 @@ h6 small {
 h1,
 h2,
 h3 {
-  margin-top: 21px;
+  margin-top: 10px;
```
nickgros, line 399: Updated so we effectively get the same styling accomplished by the `!important` in _core.scss (which pretty much always overrode this)

src/main/webapp/sass/_core.scss
```diff
@@ -2773,7 +2773,7 @@ h2,
 .h2,
 h3,
 .h3 {
-  margin-top: 10px !important;
```
nickgros, line 2776: this `!important` was wreaking havoc on headline elements styled by MUI. I didn't see any meaningful regression caused by this, but it's hard to check comprehensively.

The corresponding change in the bootstrap custom stylesheet mitigates this change

jay-hodgson (approved):

PR #5935, opened 2026-05-01
SRC 4.0.10

jay-hodgson:


pnpm-workspace.yaml
```diff
@@ -2,3 +2,4 @@ minimumReleaseAge: 4320 # 3 days
 minimumReleaseAgeExclude:
   - '@sage-bionetworks/*'
   - 'synapse-react-client'
+  - 'markdown-it-synapse'
```
jay-hodgson, line 5: markdown-it-synapse "2.0.0" published at 4/29/2026, adding to exclusion list

PR #5936, opened 2026-05-01
SWC-7818

jay-hodgson:


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -984,6 +984,14 @@ private FluentFuture<Void> configureContainerDownload() {
       return getDoneFuture(null);
     }
 
+    if (!isUserAuthenticated) {
+      actionMenu.setDownloadMenuEnabled(false);
+      actionMenu.setDownloadMenuTooltipText(
+        "You must be logged in to download folder contents."
```
jay-hodgson, line 990: Not necessary, since the menu item is not shown, but explicit tooltip text also doesn't hurt

kianamcc (approved):

PR #5941, opened 2026-05-08
SWC-7794: Replace GWT CSV upload wizard with React CsvPreviewDialog

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=SWC-7794


src/main/java/org/sagebionetworks/web/client/jsinterop/CreateTableFromCsvDialogProps.java
```diff
@@ -0,0 +1,46 @@
...
+    String parentId,
+    OnSuccessFunction onSuccess,
+    OnCloseFunction onClose
+  ) {
+    CreateTableFromCsvDialogProps props = new CreateTableFromCsvDialogProps();
+    props.open = open;
+    props.parentId = parentId;
+    props.onSuccess = onSuccess;
+    props.onClose = onClose;
+    return props;
```
nickgros, line 48: You might set `props.key = open.toString()` (or something like that) here, since we will likely always want to reset state after closing + reopening

nickgros (approved):

PR #5950, opened 2026-05-29
SWC-7839

jay-hodgson:
## Overview
Adds full support for rendering `SearchIndex` entities in the GWT portal. A
`SearchIndex` is a new Synapse entity type that defines an OpenSearch index via a
SQL query. The changes wire it into the Tables tab, entity action menus, and the
React-based `SearchQueryWrapperPlotNav` component.

---

### `SearchIndexWidget.java`
A thin GWT `ReactComponent` wrapper that instantiates the React
`SearchQueryWrapperPlotNav` component using the entity's ID and name as props.

### `SearchQueryWrapperPlotNavProps.java`
JsInterop props bridge for `SearchQueryWrapperPlotNav`. Configures the component
with `defaultShowPlots: false`, `defaultShowSearchBar: true`, and a
`tableConfiguration` with `showDownloadColumn: false`.

---

### `EntityTypeUtils.java`
- Added `SEARCH_INDEX_DISPLAY_NAME = "Search Index"` constant.
- Added `SearchIndex` import.
- `getEntityTypeForEntityClassName()`: returns `EntityType.searchindex` for
  `SearchIndex` class name (previously fell through to the `file` default, causing
  "File Tools" in the action menu and wrong icon resolution).
- `getFriendlyTableTypeName()`: returns `"Search Index"` for the type label in the
  table list.

### `SynapseJavascriptClient.java`
- `isTable()`: added `EntityType.searchindex` to the entity children API request
  so the Tables tab badge/indicator counts SearchIndex entities.

### `SRC.java`
- Registered `SearchQueryWrapperPlotNav` as a typed `ReactComponentType`.

### `EntityPageTop.java`
- `getAreaForEntity()`: routes `SearchIndex` to `EntityArea.TABLES` (before the
  generic `Table` check, which doesn't match `SearchIndex`).
- Entity-loaded callback: dispatches `SearchIndex` to `tableChanged()` so the
  Tables tab is activated on direct navigation.

### `AbstractTablesTab.java`
- `setTargetBundle()`: added an early-return branch for `SearchIndex` that renders
  the entity detail view (breadcrumb, title bar, action menu) and mounts a
  `SearchIndexWidget` into the tab content area.
- `createSearchIndexWidget()`: protected factory method (overridable in tests to
  return `null`, avoiding GWT class-loading issues in plain JUnit).

### `TablesTab.java`
- `getTypesShownInList()`: added `EntityType.searchindex` so SearchIndex entities
  appear in the project's Tables list.
- `isEntityShownInTab()`: extended to accept `SearchIndex` (which does not extend
  `Table`).

### `EntityActionControllerImpl.java`
- `configureTableCommands()` refactored into two independent checks:
  1. **`instanceof HasDefiningSql`** (covers `SearchIndex`, `MaterializedView`,
     `VirtualTable`): shows `EDIT_DEFINING_SQL` (if user can edit) or
     `VIEW_DEFINING_SQL` (if read-only). Previously this logic only ran inside
     the `Table` branch, so SearchIndex never exposed SQL editing.
  2. **`instanceof Table`**: configures table-specific actions
     (`UPLOAD_TABLE_DATA`, `EDIT_TABLE_DATA`, `SHOW_TABLE_SCHEMA`, etc.)
     unchanged.

### `DefaultEntityActionMenuLayoutUtil.java`
- Added `case searchindex:` to `getLayout(EntityType)` with:
  - **Button actions**: `EDIT_DEFINING_SQL`, `VIEW_DEFINING_SQL`,
    `SHOW_ANNOTATIONS`, `SHARE_THIS_PAGE`.
  - **Primary menu**: sharing settings, provenance, DOI/link, move/rename/delete,
    report violation, ACT actions.
  - Label: `"Search Index Tools"` (previously fell through to no case, which
    caused the menu to render as `"File Tools"` after `getEntityTypeForEntityClassName`
    defaulted to `EntityType.file`).

---

## Tests
`TablesTabTest`: added `testConfigureUsingSearchIndex` and updated the test
subclass to override `createSearchIndexWidget()` returning `null` to avoid GWT
class-loading errors in plain JUnit.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -1802,10 +1802,6 @@ public class EntityActionControllerImpl
     );
   }
 
-  public static boolean isDefinedBySql(Entity entity) {
-    return entity instanceof HasDefiningSql;
-  }
-
   public static boolean isEditCellValuesSupported(Entity entity) {
     // EntityRefCollectionView results are not editable, even if the user has permissions, because rows may reference immutable versions (SWC-5870, SWC-5903)
     return QueryResultEditorWidget.isTableTypeQueryResultEditable(
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -1365,6 +1365,22 @@ public class EntityActionControllerImpl
   }
 
   private void configureTableCommands() {
+    if (entityBundle.getEntity() instanceof HasDefiningSql) {
+      actionMenu.setActionVisible(
+        Action.EDIT_DEFINING_SQL,
+        permissions.getCanCertifiedUserEdit() && isCurrentVersion
+      );
+      actionMenu.setActionListener(Action.EDIT_DEFINING_SQL, this);
+      actionMenu.setActionVisible(
+        Action.VIEW_DEFINING_SQL,
+        !permissions.getCanCertifiedUserEdit() && isCurrentVersion
+      );
+      actionMenu.setActionListener(Action.VIEW_DEFINING_SQL, this);
+    } else {
+      actionMenu.setActionVisible(Action.VIEW_DEFINING_SQL, false);
+      actionMenu.setActionVisible(Action.EDIT_DEFINING_SQL, false);
+    }
+
     if (entityBundle.getEntity() instanceof Table) {
       boolean isEntityRefCollectionView = entityBundle.getEntity() instanceof
       EntityRefCollectionView;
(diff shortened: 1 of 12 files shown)


jay-hodgson:
Screenshot 2026-05-29 at 8 14 51 AM

Screenshot 2026-05-29 at 8 15 21 AM

Screenshot 2026-05-29 at 8 15 30 AM

Screenshot 2026-05-29 at 8 15 50 AM

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1389,21 +1405,6 @@ private void configureTableCommands() {
         isEntityRefCollectionView &&
         isCurrentVersion
       );
-      actionMenu.setActionVisible(
-        Action.EDIT_DEFINING_SQL,
-        permissions.getCanCertifiedUserEdit() &&
-        isDefinedBySql(entityBundle.getEntity()) &&
```
nickgros, line 1395: can we delete `isDefinedBySql`?

src/main/java/org/sagebionetworks/web/client/widget/entity/tabs/AbstractTablesTab.java
```diff
@@ -369,6 +379,37 @@ public void showProjectLevelUI() {
   public void setTargetBundle(EntityBundle bundle, Long versionNumber) {
     this.entityBundle = bundle;
     Entity entity = bundle.getEntity();
+
+    if (entity instanceof SearchIndex) {
```
nickgros: Seems odd to treat SearchIndex as a special case for all logic instead of just the parts that should differ (widget, anything else?), is there a reason to do this? Seems to be that you aren't using `isShownInTab` but why is that?

jay-hodgson: Refactored to share the logic (remove early return).  It's still a different way to abstract away the SearchIndexWidget (for the java unit test to run, without encountering js).  It works, I'm just not sure how I feel about it yet

nickgros (approved):

PR #5957, opened 2026-06-04
SWC-7758

jay-hodgson:


jay-hodgson:
From the favicon update:
Screenshot 2026-06-04 at 1 10 44 PM

From the manifest xml update:
Screenshot 2026-06-04 at 1 07 14 PM

src/main/webapp/opensearch.xml
```diff
@@ -7,10 +7,10 @@
   <Url
     method="GET"
     type="text/html"
-    template="https://www.synapse.org/SearchV2:default?query={searchTerms}"
+    template="https://www.synapse.org/Search:{searchTerms}"
```
jay-hodgson, line 10: Utilize the legacy search place to put the handle the search term properly (either embedding in a search query json object, or sending the user directly to the entity page)

src/main/java/org/sagebionetworks/web/client/presenter/LegacySearchPresenter.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Override
+  public void start(AcceptsOneWidget panel, EventBus eventBus) {
+    String token = place.toToken();
+    String normalizedToken = (token != null) ? token.toLowerCase().trim() : "";
+    Place redirect;
+    if (normalizedToken.matches(WebConstants.SYNAPSE_ENTITY_ID_REGEX)) {
+      redirect = new Synapse(normalizedToken);
+    } else {
+      String queryJson = "{\"queryTerm\":[\"" + token + "\"]}";
```
kianamcc: Would this break if we search term is something with double quotes like "cancer"? Should we be escaping? Maybe it's acceptable for legacy redirect?

kianamcc (approved):

kianamcc (approved):

PR #5966, opened 2026-06-10
SWC-7046 - Upgrade to Java 21

nickgros:


mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
+java = 'corretto-21'
```
nickgros, line 2: I use [mise](https://mise.jdx.dev/) to manage per-project tool versions, so it's handy to commit this in case someone else does

src/main/java/org/sagebionetworks/web/client/PortalGinModule.java
```diff
@@ -1254,6 +1256,7 @@ protected void configure() {
 
     bind(WizardProgressWidgetView.class).to(WizardProgressWidgetViewImpl.class);
 
+    bind(UploadDialogWidget.class).to(UploadDialogWidgetV2.class);
```
nickgros, line 1259: Fix missing binding, not related to Java upgrade

.mvn/jvm.config
```diff
@@ -0,0 +1,3 @@
+--add-opens java.base/java.lang=ALL-UNNAMED
+--add-opens java.base/java.util=ALL-UNNAMED
+--add-opens java.base/java.io=ALL-UNNAMED
```
nickgros, line 3: Allow reflection on java.* classes, for backwards compatibility

jay-hodgson (approved):

PR #5968, opened 2026-06-10
SWC-7046 - Tomcat 10, Replace GIN with Dagger-GWT for client dependency injection

nickgros:
The Jakarta/Tomcat 10 cutover bumps Guice to 7.0.0 (jakarta guice-servlet), but GIN 2.1.2 runs its GinjectorGenerator inside the GWT compile against the project classpath and requires Guice <=6 (javax.inject). In a single module these cannot coexist, which blocked the GWT compile (GinjectorGenerator: "No implementation for Provider<GinjectorBindings>"). Replacing GIN with dagger-gwt removes Guice from the GWT-compile classpath entirely, so the server can use Guice 7 / jakarta without a module split.

Changes:
- pom: add dagger-gwt + dagger-compiler (annotation processor), remove gin. Upgrade maven-compiler-plugin 3.6.0 -> 3.13.0 (3.6.0 silently swallowed Dagger component validation/generation).
- Portal.gwt.xml: inherit dagger.Dagger instead of com.google.gwt.inject.Inject.
- Client @Inject: com.google.inject.Inject -> javax.inject.Inject (686 files). Server keeps com.google.inject.* (Guice 7).
- PortalGinModule: rewritten as a Dagger @module (385 https://github.com/BINDS, 16 @singleton), plus 8 @provides for GWT-RPC async services (GWT.create) that GIN created implicitly.
- GwtBindingsModule (new): @provides for the types GIN auto-created via GWT deferred binding - 263 UiBinders, ClientBundles, AppPlaceHistoryMapper, and a few framework/dependency classes lacking an @Inject constructor.
- PortalGinInjector: @GinModules/Ginjector -> @singleton https://github.com/component.
- Portal.java: GWT.create(PortalGinInjector) -> DaggerPortalGinInjector.create().
- Add @Inject constructors to ~44 bound impls; annotate 16 concrete singletons with @singleton; make 7 UiBinder interfaces public; drop @Inject from 2 abstract-class constructors; widen 1 private @Inject constructor.
- Remove dead QueryInputWidget/View/ViewImpl (+ test, ui.xml) - a pre-existing broken UiBinder template (HelpWidget width) only surfaced once Dagger forced binder generation.


src/main/java/org/sagebionetworks/web/client/analytics/SearchAnalyticsClientImpl.java
```diff
@@ -1,12 +1,16 @@
...
+import javax.inject.Inject;
 import org.sagebionetworks.web.client.jsinterop.SRC;
 import org.sagebionetworks.web.client.jsinterop.analytics.SearchQueryEventData;
 import org.sagebionetworks.web.client.jsinterop.analytics.SearchResultEventData;
 import org.sagebionetworks.web.client.jsinterop.analytics.SearchResultPageReturnedEventData;
 
 public class SearchAnalyticsClientImpl implements SearchAnalyticsClient {
 
+  @Inject
+  public SearchAnalyticsClientImpl() {}
```
nickgros, line 12: Update all of our client widgets so Dagger can inject them. Dagger does compile-time injection, so we need to be a bit more verbose

src/main/java/org/sagebionetworks/web/client/cache/ClientCacheImpl.java
```diff
@@ -1,6 +1,6 @@
 package org.sagebionetworks.web.client.cache;
 
-import com.google.inject.Inject;
+import javax.inject.Inject;
```
nickgros, line 3: Since our client code is using Dagger instead of GIN, use javax.inject

pom.xml
```diff
@@ -351,7 +351,16 @@
...
+        <version>3.13.0</version>
+        <configuration>
+          <annotationProcessorPaths>
+            <path>
+              <groupId>com.google.dagger</groupId>
+              <artifactId>dagger-compiler</artifactId>
+              <version>${daggerVersion}</version>
+            </path>
+          </annotationProcessorPaths>
+        </configuration>
```
nickgros, line 363: Upgrade the maven-compiler-plugin; v3.6.0 suppressed dagger compiler errors.

Add the dagger compiler to the workflow

src/main/java/org/sagebionetworks/web/client/widget/table/v2/QueryInputView.java
nickgros, line 1: Dagger's compile-time injection surfaced a compile-time error in this code. It was unused, so with Gin it never caused any problems. Since it's dead code, we can just remove it.

src/main/java/org/sagebionetworks/web/client/GwtBindingsModule.java
```diff
@@ -0,0 +1,66 @@
...
+import org.sagebionetworks.web.client.mvp.AppPlaceHistoryMapper;
+
+/**
+ * Dagger module for the handful of injected types that have no @Inject
+ * constructor and are not created locally by a single owner: framework
+ * classes, dependency classes, and GWT generator interfaces (ClientBundles,
+ * the place-history mapper).
+ */
+@Module
+public abstract class GwtBindingsModule {
```
nickgros, line 21: GIN used to bind these implicitly/at runtime

src/main/java/org/sagebionetworks/web/client/PortalGinModule.java
```diff
@@ -749,836 +741,1780 @@
 import org.sagebionetworks.web.client.widget.user.UserBadgeView;
 import org.sagebionetworks.web.client.widget.user.UserBadgeViewImpl;
 
-public class PortalGinModule extends AbstractGinModule {
+/**
+ * Dagger module replacing the former GIN PortalGinModule.
+ * Generated from the original bind(I).to(Impl) declarations.
+ */
+@Module
```
nickgros, line 748: Replace every binding with Dagger syntax

jay-hodgson (approved):

PR #5971, opened 2026-06-11
Upgrade dependencies

nickgros:


src/main/webapp/Portal.html
```diff
@@ -204,60 +203,155 @@ <h2 class="center" style="color: white; padding-top: 180px">
       <noscript>
         <style>
           .sg-noscript {
-            font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
```
nickgros, line 207: just reformatted, no changes

src/main/webapp/Portal.html
```diff
@@ -161,7 +161,6 @@
     <script src="${cdnEndpoint}/generated/prop-types.min.js"></script>
     <script src="${cdnEndpoint}/generated/plotly.min.js"></script>
     <script src="${cdnEndpoint}/generated/create-plotly-component.min.js"></script>
-    <script src="${cdnEndpoint}/generated/universalCookie.min.js"></script>
```
nickgros, line 164: Remove script to load `universal-cookie` globally. While I propose dropping universal-cookie in https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/2894, I don't think this global actually does anything -- Vite already handles bundling its usage in SRC and it's not otherwise used by SWC.

MIGRATION.md
nickgros, line 1: Migration of Java/Tomcat is done, so we can delete the temporary guide

jay-hodgson (approved):

PR #6008, opened 2026-07-27
PORTALS-4329: Entity Page Redesign: Update Download Options dropdown menu

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=PORTALS-4329

SRC: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/3048

Screenshot 2026-07-24 at 5 22 32 PM
Screenshot 2026-07-24 at 4 56 40 PM

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/jsinterop/entity/actionmenu/EntityActionMenuDropdownConfiguration.java
+++ b/src/main/java/org/sagebionetworks/web/client/jsinterop/entity/actionmenu/EntityActionMenuDropdownConfiguration.java
@@ -11,7 +11,7 @@ public class EntityActionMenuDropdownConfiguration {
   /* Only strings that map to synapse-react-client Icons */
   private String tooltipText;
   private boolean disabled;
-  private boolean showConditionalNote;
+  private boolean showFileAccessRequestNotice;
 
   @JsOverlay
   public static EntityActionMenuDropdownConfiguration create(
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/web/client/jsinterop/entity/actionmenu/EntityActionMenuDropdownConfiguration.java
```diff
@@ -11,6 +11,7 @@ public class EntityActionMenuDropdownConfiguration {
   /* Only strings that map to synapse-react-client Icons */
   private String tooltipText;
   private boolean disabled;
+  private boolean showConditionalNote;
```
jay-hodgson: Same comment as https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/3048#discussion_r3669158288

src/main/java/org/sagebionetworks/web/client/widget/entity/menu/v3/DefaultEntityActionMenuLayoutUtil.java
```diff
@@ -36,6 +36,14 @@ private DefaultEntityActionMenuLayoutUtil() {
     .setColor("error.main")
     .setFontWeight(700L);
 
+  static SxProps downloadMenuTextStyle = SxProps
+    .create()
+    .setColor("#3B4046")
+    .setFontSize("14px")
+    .setFontWeight(540L)
```
jay-hodgson: I'm surprised that fontWeight can be a long type because in GWT, I didn't think long could be natively representable in JavaScript. GWT emulates long internally as a pair of 32-bit integers, which I thought was incompatible with native JS interop.
I see that SxProps has the @JsType(isNative = true) class, but I guess it's not forbidden by the GWT compiler??
@nickgros ?

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/client/widget/entity/menu/v3/DefaultEntityActionMenuLayoutUtil.java
kianamcc: Hm maybe this should be int instead?

PR #6011, opened 2026-07-27
SWC-7908: Bulk Edit button flashes briefly then disappears in Synapse Tables

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7908?atlOrigin=eyJpIjoiNzlmNmNiMTRiOTUwNDRmNWJhOTAxYTc5YjNkOWY1NWYiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
@@ -344,6 +344,7 @@ public class TableEntityWidgetV2
     }
     this.actionMenu.setActionVisible(
         Action.EDIT_TABLE_DATA,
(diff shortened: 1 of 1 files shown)


src/main/java/org/sagebionetworks/web/client/widget/table/explore/TableEntityWidgetV2.java
```diff
@@ -344,6 +344,7 @@ private void configureActions() {
     }
     this.actionMenu.setActionVisible(
         Action.EDIT_TABLE_DATA,
+        canEditResults &&
```
kianamcc, line 347: Other setActionVisible calls were based off of user access but this one was not. We only had a type check here so it can be visible even without user access.

afwillia (approved):

PR #6016, opened 2026-07-28
Merge release-600 into develop

kianamcc:


kianamcc:
@jay-hodgson playwright test failing due to missing AccessApproval module in synapse-types 0.0.7

jay-hodgson:
Merging, but we should sort out the synapse-types issue in a separate PR

PR #6018, opened 2026-07-30
SWC-7915: put project visibility chip behind a feature flag

jay-hodgson:


kianamcc (approved):
I think we still call configure even if we hide the chip right? Should those be gated behind a feature flag too? Or maybe that's a non-issue

jay-hodgson:
> I think we still call configure even if we hide the chip right? Should those be gated behind a feature flag too? Or maybe that's a non-issue

You are absolutely right, this is an extra call.  But since reflecting the Experimental Mode state seems to require a page refresh, I think I'd like to ensure the visibility level is accurate and always reconfigure (even if we sometimes do not show it today).

PR #6028, opened 2026-08-13
SWC-7865 - Preload error mitigation

nickgros:
Mitigation for SWC-7865 with some additional diagnostics

### DIFF
--- a/js/moduleLoadDiagnostics.js
+++ b/js/moduleLoadDiagnostics.js
@@ -24,6 +24,16 @@ const MAX_REPORTS = 25
 const seen = new Set()
 let reportCount = 0
 
+// Chunk load/link failures are usually transient (bad cache entry, edge
+// glitch) - a page reload reliably picks up a clean copy (this mirrors
+// Vite's own documented handling of vite:preloadError, and the existing
+// GWT runAsync-failure reload in Portal.java). Cap attempts per tab so a
+// genuinely broken deploy doesn't reload every affected session forever;
+// past the cap, surface a visible fallback instead of looping silently.
+const MAX_RELOAD_ATTEMPTS = 3
+const RELOAD_COUNT_KEY = 'swcModuleReloadCount'
+let hasAttemptedRecoveryThisLoad = false
+
 window.__swcModuleLoadErrors = window.__swcModuleLoadErrors || []
 
 /** Regexes that identify a lazy-module load/link failure (across browsers). */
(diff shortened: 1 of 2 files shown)


js/moduleLoadDiagnostics.js
```diff
@@ -102,10 +115,8 @@ async function inspectModule(url) {
     }
   }
 
-  const [cached, network] = await Promise.all([
-    probe('force-cache'),
-    probe('reload'),
-  ])
+  const cached = await probe('force-cache')
+  const network = await probe('reload')
```
nickgros, line 119: In my few local reproductions of the issue, the `cached` and `network` module contents are the same, which seems to defy the nature of the bug itself. It might be possible that the `network` fetch is invalidating the `cache` fetch before we get there, so running these in series instead of parallel would fix that bug (though I honestly think it is unlikely).

js/moduleLoadDiagnostics.js
```diff
@@ -172,6 +183,65 @@ function report(kind, message, stack) {
...
 
+function getReloadCount() {
+  try {
+    return parseInt(sessionStorage.getItem(RELOAD_COUNT_KEY), 10) || 0
+  } catch {
+    return 0
+  }
+}
+
+function showReloadBanner() {
```
nickgros, line 194: Show a banner if this session hit Vite module load errors triggering >3 refresh attempts. This would probably only happen if we deployed a bad artifact.

js/moduleLoadDiagnostics.js
```diff
@@ -172,6 +183,65 @@ function report(kind, message, stack) {
...
+    /* diagnostics must never break the app */
+  }
+}
+
+/**
+ * Attempt to self-heal from a preload failure by reloading, bounded by
+ * MAX_RELOAD_ATTEMPTS per tab session. At most one attempt is spent per
+ * page load, even if several chunks fail together in the same load.
+ */
+function attemptRecovery() {
```
nickgros, line 228: The last time I encountered this bug, a plain reload seemed to fix it, so I am hopeful this will help mitigate the issue a bit.

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -22,6 +22,14 @@ public class CORSFilter extends OncePerRequestFilter {
...
   public static final String ACCESS_CONTROL_ALLOW_CREDENTIALS_HEADER =
     "Access-Control-Allow-Credentials";
+  public static final String ACCESS_CONTROL_EXPOSE_HEADERS_HEADER =
+    "Access-Control-Expose-Headers";
+  // Cache/validation headers that client-side forensics (e.g. CDN chunk-load
+  // diagnostics) need to read via fetch() Response.headers - without this,
+  // the browser silently withholds them from cross-origin JS even though
+  // they're present on the wire.
+  public static final String EXPOSED_HEADERS =
+    "Age, X-Cache, ETag, Content-Encoding, Content-Length";
```
nickgros, line 32: Need to use this header for us to access these properties from the CDN network call and log them in our diagnostics, so we can continue investigating this issue with a bit more information.

jay-hodgson (approved):

PR #6031, opened 2026-08-14
SWC-7923: Entity Page Redesign: Enable the Download dropdown if use has unmet ARs, but disable the direct Download File menu item

kianamcc:
Jira: https://sagebionetworks.jira.com/jira/software/c/projects/PLFM/boards/238?assignee=712020%3Aa7fbbeb0-d7a6-443a-9dd4-246363ff09dd&selectedIssue=SWC-7927

Screenshot 2026-08-14 at 12 14 19 PM

### DIFF
--- a/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
+++ b/src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
@@ -939,9 +939,26 @@ public class EntityActionControllerImpl
           public void onSuccess(
             @Nullable RestrictionInformationResponse restrictionInformation
           ) {
+            boolean hasUnmetAccessRequirement =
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -805,53 +805,6 @@ private FluentFuture configureFileDownload() {
           );
         } else {
           actionMenu.setDownloadMenuTooltipText(NO_PERMISSION_TO_DOWNLOAD);
-          // Queue up request to identify reasons why the file cannot be downloaded, and update the tooltip when the request finishes.
```
kianamcc, line 808: Moved into helper. Called only when the menu is disabled.

jay-hodgson (approved):

PR #6037, opened 2026-08-26
SWC-7865

nickgros:
- Simplify Vite config to reduce redirects
- More diagnostics & CORS header tweaks to collect more info about static paths
- Send diagnostics to CloudWatch via /log service instead of GTM

### DIFF
--- a/js/moduleLoadDiagnostics.js
+++ b/js/moduleLoadDiagnostics.js
@@ -9,23 +9,34 @@
  * This module hooks the browser-level signals that fire
  * regardless of the boundary (Vite's `vite:preloadError`, global `error`, and
  * `unhandledrejection`) to answer two questions:
- *   1. How often and when is this happening? (an event is pushed to GTM `dataLayer` + logged)
+ *   1. How often and when is this happening? (logged to CloudWatch via
+ *      `POST /repo/v1/log`, see `sendLog`, and to the console)
  *   2. Why? When a failing chunk URL is known, we re-fetch BOTH the cached copy and a fresh
- *      network copy and record status / size / encoding / whether the trailing `export {...}`
- *      is present. That distinguishes the competing hypotheses at the point of failure:
- *        - truncated / corrupted body (parses but has no exports)
- *        - stale-or-wrong cached body (cached copy differs from network copy)
- *        - a missing export name (build/skew)
+ *      network copy and record status / size / head+tail bytes / encoding / how many times the
+ *      URL was actually requested on this page (via Resource Timing) / whether the *specific*
+ *      named export Vite reported missing is present (`hasNamedExport` - not just "some export
+ *      {...} exists somewhere in the body", which is true of every valid module and false of
+ *      every stylesheet, and so cannot distinguish any of the following). That distinguishes the
+ *      competing hypotheses at the point of failure:
+ *        - a short read (body ends cleanly, but before the trailing export statement)
+ *        - wrong bytes (a different module's body under this URL)
+ *        - a duplicate module record (the URL was fetched more than once on this page)
  *
  * This module must never throw or interfere with the app: every path is guarded.
  */
-
 const MAX_REPORTS = 25
 const seen = new Set()
+// Separate from `seen`: which (url, message) pairs have already been counted
+// against MAX_REPORTS and had the forensic re-fetch run. `seen` dedupes per
+// *boundary* (kind is part of its key); this dedupes per *failure*, so a
+// failure caught by two boundaries only spends one report-budget slot and
+// only triggers one forensic re-fetch. See the comment at its use in
+// `report()`.
+const firstSeenFailures = new Set()
 let reportCount = 0
 
 // Chunk load/link failures are usually transient (bad cache entry, edge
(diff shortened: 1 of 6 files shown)


vite.config.ts
```diff
@@ -6,8 +6,7 @@ import path from 'path'
  * Vite config to generate the ESM & CJS bundles for Synapse React Client.
  */
 const config = defineConfig({
-  // Assets will be served from the `/Portal/cdn` servlet, which will redirect requests to the CDN when appropriate
-  base: '/Portal/cdn/generated/vite/',
```
nickgros, line 10: I added this redirect service because I thought it was the only way to get Vite to consistently point to the CDN. Apparently, we can just use a relative URL and it should just work.

I have the faintest hope that this fixes the module load errors, but I don't really have evidence for that. Still, this becomes more simple and should eliminate a ton of unnecessary round trips and reduce load on the portal machines when fetching assets.

If this works (hard to know 100% without testing against a CDN), then we can remove the CdnRedirectorServlet, which I will create a ticket for after this gets merged.

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -30,6 +30,14 @@ public class CORSFilter extends OncePerRequestFilter {
...
   public static final String EXPOSED_HEADERS =
     "Age, X-Cache, ETag, Content-Encoding, Content-Length";
+  // Enables moduleLoadDiagnostics.js to capture size fields (decodedBodySize,
+  // encodedBodySize, transferSize) on cross-origin requests.
+  //
+  // Deliberately NOT sent on every response. `Timing-Allow-Origin: *` lets any
+  // origin read a resource's transferred/encoded size even when the CORS check
+  // denied it the body. This is public info for static assets.
+  public static final String TIMING_ALLOW_ORIGIN_HEADER = "Timing-Allow-Origin";
+  public static final String STATIC_ASSET_PATH_SEGMENT = "/generated/";
```
nickgros, line 40: Since asset fetches are cross-origin (`www.synapse.org` to `cdn-www.synapse.org`), we need to set this header to allow JavaScript to access timing/performance data. This header will only be set on asset requests (checked by path).

js/moduleLoadDiagnostics.js
```diff
@@ -9,23 +9,34 @@
  * This module hooks the browser-level signals that fire
```
nickgros, line 9: Main changes in this file are

1. Collect a lot of new data, including
  - A label for the message shape to distinguish between the different types of module load failures
  - Stack version and JS build ID
  - Reload attempts
  - `hasNamedExport` now specifically checks for the correct export name
  - The `head` of the network/cached module file
  - `PerformanceResourceTiming` data
  - Lots of network data that was logged but not pushed to GTM since we hit our custom dimensions limit there (once I delete the custom dimensions I added for this investigation, we'll still have some headroom)
2. Now use the `POST /repo/v1/log` service to log the data to CloudWatch. In [PLFM-9565](https://sagebionetworks.jira.com/browse/PLFM-9565), only the CloudWatch Metrics were disabled (it was very expensive because of the high cardinality of `User-Agent`). The data is still logged and queryable.

src/main/java/org/sagebionetworks/web/server/servlet/filter/CORSFilter.java
```diff
@@ -130,6 +148,9 @@ protected void doFilterInternal(
 
     response.addHeader(ACCESS_CONTROL_ALLOW_ORIGIN_HEADER, allowOrigin);
     response.addHeader(ACCESS_CONTROL_EXPOSE_HEADERS_HEADER, EXPOSED_HEADERS);
+    if (isStaticAssetRequest(request)) {
+      response.addHeader(TIMING_ALLOW_ORIGIN_HEADER, DEFAULT_ALLOW_ORIGIN);
```
jay-hodgson, line 152: Unlike `Access-Control-Allow-Origin`, I verified `*` is valid for `Timing-Allow-Origin` even if `Access-Control-Allow-Credentials` is set to `true`.

src/main/java/org/sagebionetworks/web/server/servlet/ViteHTMLGeneratorImpl.java
```diff
@@ -12,8 +12,7 @@
 public class ViteHTMLGeneratorImpl implements ViteHTMLGenerator {
 
   private static final String VITE_DEV_SERVER_ORIGIN = "http://localhost:5173";
-  // The base path should match the value in the Vite config.
-  private static final String VITE_BASE_PATH = "/Portal/cdn/generated/vite";
+  private static final String VITE_BASE_PATH = "";
```
jay-hodgson, line 15: If this path is now an empty string, can it be removed?

jay-hodgson (approved):

src/main/java/org/sagebionetworks/web/server/servlet/ViteHTMLGeneratorImpl.java
nickgros: It could be removed, which I can do in the same pass to remove the redirector service

PR #6041, opened 2026-08-28
SWC-7960

jay-hodgson:
### Production code (`FileHandleAssociationServlet.java`)
- **New behavior :** `DataAccessRequestAttachment` is streamed same-origin with:
  - `Content-Type: application/pdf`
  - `Content-Disposition: inline`
  - `Cache-Control: private, no-store` (overrides the default 30s cache header)
  - Errors continue to flow through the existing `SynapseException` → error-place redirect.
- **Refactor:** Converted the pre-existing `UserProfileAttachment` / `TeamAttachment` streaming branch from `IOUtils.closeQuietly` in a `finally` block to `try-with-resources`, matching the style of the new branch.
  - Rationale: `IOUtils.closeQuietly(Closeable)` has been deprecated since Commons IO 2.6; try-with-resources properly surfaces close-time failures (as suppressed exceptions) instead of silently swallowing them.

### Tests (`FileHandleAssociationServletTest.java`)
Added a helper and three new tests that avoid mocking the final `URL` class by pointing the stubbed `getFileURL` at a real temp file (`file://…`) and capturing bytes written to the mocked `ServletOutputStream`.

- `stubStreamingUrlAndCaptureResponse(byte[])` — creates a temp file, wires `getFileURL` to return its `file://` URL, and returns a `ByteArrayOutputStream` capturing writes to the response.
- `@After teardown()` — deletes the temp file.

New tests:
- **`testDoGetDataAccessRequestAttachmentStreamsAsPdf`** — verifies `Content-Type: application/pdf`, `Content-Disposition: inline`, `Cache-Control: private, no-store`, no redirect, and that the file bytes reach the response output stream.
- **`testDoGetDataAccessRequestAttachmentErrorRedirects`** — verifies that a `SynapseException` from `getFileURL` still redirects to the error place and does not set the PDF content type.
- **`testDoGetUserProfileAttachmentStreamsAndCaches`** — locks in the long-cache header and byte-copy behavior for the pre-existing streaming branch (previously untested), covering the try-with-resources refactor.

### DIFF
--- a/src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URL;
+import java.net.URLConnection;
 import java.net.URLEncoder;
 import org.apache.commons.io.IOUtils;
 import org.sagebionetworks.client.SynapseClient;
@@ -126,13 +127,19 @@ public class FileHandleAssociationServlet extends HttpServlet {
             fha.getAssociateObjectType()
           )
         ) {
-          // SWC-7960: Stream DUC PDFs same-origin so an iframe preview isn't blocked by CORS on the S3 hop.
-          response.setContentType("application/pdf");
+          // SWC-7960: Stream Data Access Request attachments (eDUCs, traditional DUCs, IRB
+          // approvals, etc.) same-origin so an iframe preview isn't blocked by CORS on the S3 hop.
+          URLConnection connection = resolvedUrl.openConnection();
+          // Forward the upstream Content-Type so DOCX / images / PDF are all served correctly.
+          String upstreamContentType = connection.getContentType();
+          if (upstreamContentType != null) {
+            response.setContentType(upstreamContentType);
+          }
           response.setHeader("Content-Disposition", "inline");
-          // PDF contains identifying information, so don't cache it
-          response.setHeader("Cache-Control", "private, no-store");
+          // Attachments may contain identifying information; never store.
+          response.setHeader("Cache-Control", "no-store");
           try (
-            InputStream in = resolvedUrl.openStream();
+            InputStream in = connection.getInputStream();
             OutputStream out = response.getOutputStream()
           ) {
             IOUtils.copy(in, out);
--- a/src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
+++ b/src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
@@ -130,12 +130,18 @@ public class FileHandleAssociationServlet extends HttpServlet {
           // SWC-7960: Stream Data Access Request attachments (eDUCs, traditional DUCs, IRB
           // approvals, etc.) same-origin so an iframe preview isn't blocked by CORS on the S3 hop.
           URLConnection connection = resolvedUrl.openConnection();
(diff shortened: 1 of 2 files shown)


src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
```diff
@@ -111,19 +111,31 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+          ) {
+            IOUtils.copy(in, out);
+          }
+        } else if (
+          FileHandleAssociateType.DataAccessRequestAttachment.equals(
+            fha.getAssociateObjectType()
+          )
+        ) {
+          // SWC-7960: Stream DUC PDFs same-origin so an iframe preview isn't blocked by CORS on the S3 hop.
+          response.setContentType("application/pdf");
```
nickgros: I do wonder if DataAccessRequestAttachments could be other types, like a DOCX traditional DUC. Can you clarify: will this service only be used for eDUC preview?

src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
```diff
@@ -111,19 +111,31 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
+          response.setHeader("Content-Disposition", "inline");
+          // PDF contains identifying information, so don't cache it
+          response.setHeader("Cache-Control", "private, no-store");
```
nickgros: These are conflicting directives, is this supposed to act as a fallback? Seems like it should either be `private, no-cache` or just `no-store`?

nickgros (approved):
Had a few questions

src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
jay-hodgson: Good call, changed to just `no-store`

src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
jay-hodgson: Only used for pdf today, but maybe we will reuse for other file types later.  Let's implement in a generic way (sniff the original content type instead)

src/main/java/org/sagebionetworks/web/server/servlet/FileHandleAssociationServlet.java
```diff
@@ -126,13 +127,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response)
...
-          response.setContentType("application/pdf");
+          // SWC-7960: Stream Data Access Request attachments (eDUCs, traditional DUCs, IRB
+          // approvals, etc.) same-origin so an iframe preview isn't blocked by CORS on the S3 hop.
+          URLConnection connection = resolvedUrl.openConnection();
+          // Forward the upstream Content-Type so DOCX / images / PDF are all served correctly.
+          String upstreamContentType = connection.getContentType();
+          if (upstreamContentType != null) {
+            response.setContentType(upstreamContentType);
+          }
           response.setHeader("Content-Disposition", "inline");
```
nickgros: Should we do `inline` only if it's a PDF? Images might be fine too

jay-hodgson: Good point.  There's a security issue here, since this may eventually be user uploaded content (which could cause the reviewer to run code if we reuse this servlet for that purpose).
Ok, let's be strict here, and only allow `inline` if pdf or a subset of known raster image types (svg is also potentially dangerous, for example).

nickgros (approved):

PR #6051, opened 2026-09-18
SWC-7994: includes SRC version update, repo artifact version update

jay-hodgson:
Bumping `synapse.version` to `606.0` in a JDK 21 project surfaced three separate blockers:

1. An intermittent `javac` NPE inside Dagger's annotation processor during `pnpm dev`
2. GWT compile failures at browser load time (`RequestInterface` / `Submission` "has no instantiable subtypes")
3. Two smaller code changes for SWC-7994 (unrelated to 606.0)

## 1. Feature flag wiring for `QueryWrapperPlotNav`

Added a new `showQueryBuilderControl` prop to the `QueryWrapperPlotNav` React component and piped it through from Java. Value comes from a new feature flag.

- [FeatureFlagKey.java](src/main/java/org/sagebionetworks/web/client/FeatureFlagKey.java) — added `QUERY_BUILDER_CONTROL`.
- [QueryWrapperPlotNavProps.java](src/main/java/org/sagebionetworks/web/client/jsinterop/QueryWrapperPlotNavProps.java) — added `showQueryBuilderControl` field + `create()` parameter.
- [QueryWrapperPlotNav.java](src/main/java/org/sagebionetworks/web/client/widget/table/explore/QueryWrapperPlotNav.java) — threaded new param.
- [TableEntityWidgetViewImpl.java](src/main/java/org/sagebionetworks/web/client/widget/table/v2/TableEntityWidgetViewImpl.java) — reads `FeatureFlagKey.QUERY_BUILDER_CONTROL` from `ginInjector.getFeatureFlagConfig()` and passes it.
- [DataCatalogPageViewImpl.java](src/main/java/org/sagebionetworks/web/client/view/DataCatalogPageViewImpl.java) — passes `false` at existing call site to preserve behavior.

## 2. `pnpm dev` compile reliability (Dagger + javac NPE)

Under `pnpm dev`, `mvn clean gwt:run-codeserver` intermittently died in Dagger's XProcessing round with `Cannot invoke JavaFileObject.isNameCompatible(...) because "toplevel.sourcefile" is null` inside `javac`'s `Modules.enterModule`. Applied a layered fix in [pom.xml](pom.xml):

- Bumped `maven-compiler-plugin` from `3.13.0` → `3.14.0`.
- Added `<fork>true</fork>` — runs `javac` in a separate JVM.
- Added `<proc>full</proc>` — explicit annotation-processing mode (JDK 21+); this was the change that made things reliable in isolated testing.

Failure still reproduces occasionally under `pnpm dev`, so as a belt-and-suspenders in package.json:

- Wrapped `dev:codeserver` in a shell retry loop (up to 10 attempts). On first success the codeserver keeps running and the loop exits.

## 3. GWT compile: `has no instantiable subtypes`

After the codeserver came up, loading the app produced:

```
'org.sagebionetworks.repo.model.dataaccess.RequestInterface' has no instantiable subtypes
[ERROR] Line 2497: Failed to resolve 'org.sagebionetworks.web.client.DataAccessClient' via deferred binding
```

**Root cause:** In synapse `lib-auto-generated` 606.0, `Request`, `Renewal`, and `Submission` gained a `private Object schemaData;` field. GWT-RPC can't serialize `java.lang.Object` (subtypes are unknowable at compile time), so those classes were excluded from serialization and `RequestInterface` ended up with no instantiable subtypes.

**Fix — refactored `DataAccessClient` to stop exposing these types as RPC parameters/returns:**

- [DataAccessClient.java](src/main/java/org/sagebionetworks/web/client/DataAccessClient.java) — deleted dead methods `getDataAccessRequest` / `updateDataAccessRequest` (never called anywhere); changed `updateDataAccessSubmissionState` return type from `Submission` → `void`.
- [DataAccessClientAsync.java](src/main/java/org/sagebionetworks/web/client/DataAccessClientAsync.java) — mirroring changes; callback is now `AsyncCallback<Void>`.
- [DataAccessClientImpl.java](src/main/java/org/sagebionetworks/web/server/servlet/DataAccessClientImpl.java) — removed the dead server impls; `updateDataAccessSubmissionState` no longer returns the `Submission` from the SynapseClient call.
- [ACTDataAccessSubmissionWidget.java](src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/submission/ACTDataAccessSubmissionWidget.java) — on success, updates the local `submission` (`setState` + `setRejectedReason` when rejecting) and re-calls `configure(submission)` to refresh the view.
- [ACTDataAccessSubmissionWidgetTest.java](src/test/java/org/sagebionetworks/web/unitclient/widget/accessrequirements/submission/ACTDataAccessSubmissionWidgetTest.java) — updated `AsyncMockStubber.callSuccessWith(...)` to pass `(Void) null`.

After this, `mvn clean compile gwt:compile` returns `BUILD SUCCESS` with `Compile of permutations succeeded`.

## Verification

- `mvn clean compile gwt:compile` → BUILD SUCCESS.
- `pnpm dev` → codeserver ready in ~35–65 sec (occasionally after 1–4 automatic retries).
- The pre-existing `Ignored 4 units with compilation errors in first pass` warning is unchanged (from `DockerNameUtil`, `ListWrapper`, `IdExtractorUtil`, `ExceptionUtil` referencing non-GWT-translatable Java reflection APIs) and unrelated to this work.


src/main/java/org/sagebionetworks/web/client/widget/accessrequirements/submission/ACTDataAccessSubmissionWidget.java
```diff
@@ -203,15 +203,19 @@ public void updateDataAccessSubmissionState(
...
         @Override
         public void onFailure(Throwable caught) {
           synAlert.handleException(caught);
         }
 
         @Override
-        public void onSuccess(Submission result) {
-          configure(result);
+        public void onSuccess(Void result) {
+          submission.setState(state);
```
jay-hodgson, line 214: This would have updated etag in submission, but we never actually pass the entire local Submission object to the server (just the ID), so this should work

package.json
```diff
@@ -90,7 +90,7 @@
     "start:staging": "docker compose --env-file .env.staging up",
     "start:dev": "docker compose --env-file .env.development up",
     "dev": "pnpm concurrently -k -n \"CODESERVER,WATCHER,TOMCAT,VITE\" -c \"auto,auto,auto,auto\" \"pnpm dev:codeserver\" \"pnpm dev:watcher\" \"pnpm dev:tomcat\" \"pnpm dev:vite\"",
-    "dev:codeserver": "mvn clean gwt:run-codeserver",
+    "dev:codeserver": "n=1; while [ $n -le 10 ]; do echo \"[dev:codeserver] attempt $n/10\"; mvn clean gwt:run-codeserver && exit 0; echo \"[dev:codeserver] attempt $n/10 failed (intermittent Dagger+javac21 NPE), retrying...\"; n=$((n+1)); done; exit 1",
```
jay-hodgson, line 93: For robustness only

src/main/java/org/sagebionetworks/web/client/DataAccessClient.java
```diff
@@ -28,11 +26,6 @@ public interface DataAccessClient extends RemoteService {
   ResearchProject updateResearchProject(ResearchProject researchProject)
     throws RestServiceException;
 
-  RequestInterface getDataAccessRequest(Long id) throws RestServiceException;
```
jay-hodgson, line 31: Dead code, delete it

src/main/java/org/sagebionetworks/web/client/DataAccessClient.java
```diff
@@ -28,11 +26,6 @@ public interface DataAccessClient extends RemoteService {
...
-
-  RequestInterface updateDataAccessRequest(RequestInterface dataAccessRequest)
```
jay-hodgson, line 33: Dead code, delete it

package.json
nickgros: Do you get the same issue when using `start:<mode>` commands? I mainly use those currently

nickgros (approved):

package.json
jay-hodgson: I'll switch to using `start:<mode>` and let you know!

PR #6052, opened 2026-09-18
SWC-7867: Add "Add Conditions for Use" to the entity Tools menu

kianamcc:
Jira: https://sagebionetworks.jira.com/browse/SWC-7867
SRC: https://github.com/Sage-Bionetworks/synapse-web-monorepo/pull/3220

- Port SRC's ImposeRestrictionDialog
- new ADD_CONDITIONS_FOR_USE action
- visible when canChangePermissions + restriction level is OPEN + not a view/collection + not a project area
- cleanup old button click references

Screenshot 2026-09-18 at 12 30 55 PM
Screenshot 2026-09-18 at 12 36 34 PM
Screenshot 2026-09-18 at 5 23 00 PM
Screenshot 2026-09-18 at 5 23 41 PM


kianamcc:
Old button showed for canChangePermissions || isActMember. Dropping || isActMember - ACT member w/o canChangePermissions will not see this item (previously they would just be redirected to the Access Requirements page, which the existing "Manage Access Requirements (ACT) item already does.

src/main/java/org/sagebionetworks/web/client/widget/entity/controller/EntityActionControllerImpl.java
```diff
@@ -1342,6 +1354,11 @@ private void onCreateOrUpdateDoi() {
       .configure(entity, getVersionIfNotLatest().orElse(null), true);
   }
 
+  private void onAddConditionsForUse() {
+    getImposeRestrictionDialog()
```
jay-hodgson, line 1359: I think we are losing some functionality here, but please tell me if I'm wrong.  If an ACT member clicks Add Conditions For Use, we should not pop up the ImposeRestrictionsDialog (since they do not need to go through the imposed LockAccessRequirement workflow), and instead we should send the ACT member directly to the Access Requirements management page (see deleted code in `BasicTitleBar.addActClickhandler()` in this PR).

kianamcc: @jay-hodgson Updated so that ACT members get redirected to the Access Requirements management page.

Question: Should this item even be visible to ACT members at all? They already have "Manage Access Requirements (ACT)" in the same menu which seems to go to the exact same page.

jay-hodgson: Good point!  Makes sense if this action was hidden for ACT (with a code comment about the Manage AR action)

kianamcc: @jay-hodgson 
"Add conditions for use" not in tools menu for ACT members, behavior already reflected in "Manage Access Requirements" action:
Screenshot 2026-09-21 at 2 43 40 PM
Screenshot 2026-09-21 at 2 43 46 PM

non-act with relevant conditions met open dialog:
Screenshot 2026-09-21 at 2 55 48 PM
Image

jay-hodgson (approved):

